PR #1, opened 2022-05-17
improve join conditions and better hash function

mkarmona:


d0choa (commented):
modification on the on clause. Including joining by the partiioned field in left and right

PR #2, opened 2022-08-10
Do black action

d0choa:
Trying to see if the github actions are considered


d0choa:
This one fails because of some extra white lines in the overlaps.py file

d0choa:
this one works after running black on the offending file

PR #4, opened 2022-08-16
Managing build and dependencies using poetry + makefile

d0choa:
The point of these changes is to be able to submit the job to dataproc as a bundle (zip in this case).

This will allow having a system to manage dependencies, update them and submit the code to dataproc with minimum intervention. Including dependencies that are not in dataproc should be trivial now, the same as adding more components to the codebase that go beyond coloc. 

The Makefile handles this sort of high-level devops of creating the build, the cluster and submitting the job. `make help` shows how to do all these things, so it's relatively well self-documented.

Just a first pass.


d0choa:
I broke the pylint check in the process apparently

d0choa:
Pylint checks now consider the project dependencies specified in poetry

d0choa:
Now all code is reformated and code quality is assessed when performing a commit using `pre-commit`. 

Next, a list of all the implemented checks

Screenshot 2022-08-17 at 21 37 10

DSuveges (approved):

PR #11, opened 2022-08-26
ba_cheers

bruno-ariano:
add cheers to the ETL. The config files are examples.


d0choa:
I fixed the Makefile. The idea is that you can do the next operations:

1. `make build`: To create the build of the package that will be stored with all the dependencies
2. `make prepare_cheers`: it will create the machine for you in GCP
3. `make run_cheers`: to submit the job to Dataproc

At the moment, this is crashing because you are importing Utils.py which is not in the project

```
Waiting for job output...
  File "/tmp/4a5848ff238c4c1a80eea32fb3fd5df0/run_cheers.py", line 22, in <module>
    from Utils import *
ModuleNotFoundError: No module named 'Utils'
```

If you need to import the module, you can create a directory `src/cheers` similar to the ones we have created for intervals and coloc.

On a separate note. After you have done `make setupdev` once in your project, try to run the next line:

```
poetry run pre-commit run --all-files 
```

It will automatically reformat your code when it can and complain if you are not following the rules. Try to clean it up until your code passes all tests. After the pain, you will hopefully appreciate the beauty ;)

DSuveges:
For the time being, the branch is still under development. For the record, the branch is being kept to make sure no implemented logic is lost.

PR #13, opened 2022-09-01
Variant annot

DSuveges:
Updates:
- Variant annotation is added as new feature.
- Unittesting as a pre-commit hook is added, however no test yet, as I was not testing the above code.
- Adding a folder with documentation
- Adding documentation for intervals,
- Adding documentation for variant annotation.

**Caveat**: Variant annotation is not added to the pipeline yet. Because there's a problem: this script needs hail, however hail depends on pyspark. SO, we we have to package pyspark as dependency when deploying. We need to discuss this.


d0choa:
To make sure the new `build` and `submit` system make it to production I will merge this PR.

However, the variant annotation step crashes. The error didn't ring any bells for me @DSuveges

d0choa (approved):

PR #16, opened 2022-09-09
Refactoring the data model of the allele frequencies in variant annotation

DSuveges:
For details see ticket: [#2717](https://github.com/opentargets/issues/issues/2717)


ireneisdoomed (commented):
The schema looks ok, however with this way we will have for all variants an array with the frequencies of all possible populations, even though these are 0. Is this what we want? I'd rather have them set to null.

For example, see [2_22419067_C_T](https://gnomad.broadinstitute.org/variant/2-22419067-C-T?dataset=gnomad_r3)

PR #20, opened 2022-09-23
Intervals from target

d0choa:
Intervals calculated from the platform target

The intention now is to use Platform target for gene index as-is.
The changes in this PR aim to maintain all functionality from intervals.
Coverage should increase, due to mappings to obsoleted gene symbols.
Counts not yet compared to the main branch.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
           - pep8-naming==0.13.1
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.37.3
+    rev: v2.38.0
     hooks:
       - id: pyupgrade
         args:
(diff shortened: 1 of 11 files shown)


DSuveges:
The new code runs, due to changes in the target index generation the number of intervals grew significantly. This increase can be explained by:
- Now other biotypes are also included when joining by region.
- An incerased number of genes can be matched with a wider set of gene symbols.

The runtime is not affected significantly. All column names are now camel cased. Schema of the dataset:
```
root
 |-- chromosome: string (nullable = true)
 |-- start: string (nullable = true)
 |-- end: string (nullable = true)
 |-- geneId: string (nullable = true)
 |-- score: double (nullable = true)
 |-- datasetName: string (nullable = true)
 |-- dataType: string (nullable = true)
 |-- experimentType: string (nullable = true)
 |-- pmid: string (nullable = true)
 |-- bioFeature: string (nullable = true)
```

PR #33, opened 2022-10-19
Add variant index generation module

ireneisdoomed:
This PR includes:
- `variant_index.py` with the functions to generate an index based on the variant annotation. The main logic produces 2 outputs:
    - Variant index - variants contained in the credible set annotated with data from gnomad (~6M variants)
    - Invalid variants from the credible set. List of variants in the credible set we have in production that fall outside of the variant annotation set (~40k).
- Docs about both variant datasets.
- `spark_helpers.py` with functions that might be helpful in other sides of the pipelines. I wrote these to process the target index to extract the nearest gene of a variant (which doesn't happen anymore in this pipeline). However, I think the windowing functions are useful and can be used in places like [this](https://github.com/opentargets/genetics_etl_python/blob/4919545cf6fde1b52b8c57d3b28e3debcb2616dc/src/etl/gwas_ingest/process_associations.py#L71).


d0choa:
`get_gene_tss` is duplicated (`spark_helpers.py` and `utils.py`)

src/etl/variants/variant_index.py
```diff
@@ -0,0 +1,98 @@
...
+    credset = etl.spark.read.parquet(credible_sets_path).select(
+        "leadVariantId",
+        "tagVariantId",
+        f.split(f.col("leadVariantId"), "_")[0].alias("chromosome"),
+    )
+    return (
+        credset.selectExpr("leadVariantId as id", "chromosome")
+        .union(credset.selectExpr("tagVariantId as id", "chromosome"))
+        .dropDuplicates(["id"])
+    )
```
DSuveges, line 65: I have a very strong impression that getting the union of the lead and tag variants is not needed as the lead variant is always in the tag variant list. Although this is just an assumption, it worth double check with someone from genetics.

DSuveges (changes requested):
Just a few comments really. Also it would make sense to remove .DStore file, and add this pattern to gitignore.

src/etl/variants/variant_index.py
ireneisdoomed: This is actually not the case. 6% of lead variants are not within the tagging set. I've asked this to @bruno-ariano and he also does not expect this happens necessarily.

src/etl/common/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+from typing import TYPE_CHECKING
+
+import pyspark.sql.functions as f
+
+if TYPE_CHECKING:
+    from pyspark.sql import Column
+
+
+def get_gene_tss(strand_col: Column, start_col: Column, end_col: Column) -> Column:
```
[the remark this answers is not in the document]
ireneisdoomed: Duplicated, yes. Thanks for flagging.

src/etl/variants/variant_index.py
DSuveges: I don't get this. This should not be happening. We'll look into this later, but in the meantime, we can keep this more stringent logic.

DSuveges (approved):

PR #34, opened 2022-10-20
Fixing bug in pvalue -> z-score calculation

DSuveges:


DSuveges:
The above method is not usable to convert p-values lower than 1e-15 to z-score.

PR #37, opened 2022-10-20
Do_removepandas

d0choa:
Skip first row of dataset by providing spark schema. We remove pandas
as a hard dependency (although hail will install it anyway)


d0choa:
I branched this from another branch :(

DSuveges (approved):
All makes sense. Great to see pandas go.

PR #38, opened 2022-10-21
Adding variant pipeline docs

DSuveges:
the __init__ script was missing!


d0choa (approved):
One more section tested in docs. Great!

You can always check how docs will look like using:
`poetry run mkdocs serve`

PR #39, opened 2022-10-31
codecov coverage target

d0choa:
Aiming to report a coverage target and a tolerance threshold

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -1,5 +1,5 @@
 name: tests
-on: [push, pull_request]
+on: [pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


d0choa:
I think it works now, but we will find out once someone tries to push a PR without enough coverage

.github/workflows/pytest.yaml
```diff
@@ -1,5 +1,5 @@
 name: tests
-on: [push, pull_request]
+on: [pull_request]
```
ireneisdoomed, line 2: Just as a note, I was in doubt whether the action would only be triggered when opening a PR and I learnt that PRs are triggered on three subactivities by default: opened, synchronize, and reopened. So all subsequent commits to the PR will be checked ✅

ireneisdoomed (approved):
Great, would be cool to organise some sort of testathon in the future and bring that threshold higher 🚀

PR #41, opened 2022-11-04
Extract V2G evidence from functional predictions

ireneisdoomed:
This PR includes:
- the module that parses functional predictions from VEP as described in [#2789](https://github.com/opentargets/issues/issues/2789)
- adjustments to the interval parsers to be generated as part of the V2G dataset and not independently, also described in [#2790](https://github.com/opentargets/issues/issues/2790)
- Necessary documentation
- Good testing coverage.

Each source of evidence is illustrated in the relevant ticket.
 
The testing module is long because there is a lot of logic to create the mock and expected data, however, I am quite happy with the way tests are defined. Instead of having many for each situation, a single function is parametrised with the data covering the different scenarios. I'll keep learning how to improve this, for example, I've read that it is a good practice to define fixtures inside `conftest.py`. I need to think about this because in our case we have a lot of these, I've seen examples where `conftest.py` consists of a list of imports to other submodules to keep things tidy.

The generated dataset is exported **partitioned per chromosome and unsorted by position**. In this case, I don't want to sort because adding this functionality added a lot of computation time + V2G is not a dataset that we intend to use to join data primarily if I understand it correctly.
Here is the location: `gs://genetics_etl_python_playground/XX.XX/output/python_etl/parquet/v2g`

The [job](https://console.cloud.google.com/dataproc/jobs/389d770e5d084ccb947f85a10edff142/monitoring?region=europe-west1&project=open-targets-genetics-dev) takes **2h40** to run on a high-memory machine. Computation-wise, I've seen that CPU utilisation is 100% all the time, so the more resources you throw at it, the better. I got a 1h40 minutes run the other day, but after doing some changes I got worse. I'd keep looking at how to optimise it, but I think it is good enough for the time being.


src/etl/v2g/functional_predictions/vep.py
```diff
@@ -0,0 +1,201 @@
...
+        variant_consequence_lut_path (str): The path to the LUT between the functional consequences and their assigned V2G score
+
+    Returns:
+        DataFrame: variant to gene assignments from VEP
+    """
+    # function that takes the index, gets the vep data and parses it
+    etl.logger.info("Loading variant index")
+    variant_consequence_lut = read_consequence_lut(etl, variant_consequence_lut_path)
+    va_with_vep = (
+        etl.spark.read.parquet(variant_annotation_path)
```
d0choa, line 37: better to use `read_parquet(variant_annotation_path, "variant_annotation.json")` as we already have it

src/etl/v2g/functional_predictions/vep.py
```diff
@@ -0,0 +1,201 @@
...
+        # exploding the array already removes record without VEP annotation
+        .select(
+            "id", f.explode("vep.transcriptConsequences").alias("transcriptConsequence")
+        )
+    )
+    vi = etl.spark.read.parquet(variant_index_path).select("id")
+    annotated_variants = (
+        va_with_vep.join(vi, on="id", how="inner").coalesce(20).persist()
```
d0choa, line 45: `coalesce(20)`?

src/run_v2g.py
```diff
@@ -0,0 +1,89 @@
...
+import hydra
+import pyspark.sql.functions as f
+
+from etl.json import validate_df_schema
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
+
+from etl.common.ETLSession import ETLSession
+from etl.v2g.functional_predictions.vep import main as extract_v2g_from_vep
```
d0choa, line 16: does this sound like a good idea?

src/run_v2g.py
```diff
@@ -0,0 +1,89 @@
...
+        cfg.etl.v2g.inputs.vep_consequences,
+    )
+    v2g_datasets = [interval_df, *func_pred_datasets]
+    v2g = reduce(lambda x, y: x.unionByName(y, allowMissingColumns=True), v2g_datasets)
+
+    validate_df_schema(v2g, "v2g.json")
+    (
+        v2g.select("*", f.split("variantId", "_")[0].alias("chromosome"))
+        .coalesce(cfg.etl.v2g.parameters.partition_count)
+        .write.partitionBy("chromosome")
```
d0choa, line 80: I would do:
- `sort("position")` and `repartition("chromosome")` 
and after
- `partitionBy("sourceID")` (a new column derived from `datasourceId`)

src/run_v2g.py
```diff
@@ -0,0 +1,89 @@
...
+    ).transform(lambda df: get_variants_in_interval(df, vi))
+    func_pred_datasets = extract_v2g_from_vep(
+        # Parsing VEP functional consequences
+        etl,
+        cfg.etl.v2g.inputs.variant_index,
+        cfg.etl.v2g.inputs.variant_annotation,
+        cfg.etl.v2g.inputs.vep_consequences,
+    )
+    v2g_datasets = [interval_df, *func_pred_datasets]
+    v2g = reduce(lambda x, y: x.unionByName(y, allowMissingColumns=True), v2g_datasets)
```
d0choa, line 74: keep an eye on this step as a potential liability when the v2g datasets are too big. We don't really need it as we could be writing directly each datasource, but I understand it's convenient on subsequent steps (e.g. `partitionBy`)

src/run_v2g.py
ireneisdoomed: renaming the main function you mean?
Why I did it like this: since each module features the functions needed to build each dataset I thought it was a good idea to have a `main` function to make clear what is the main function that hosts all the logic. And when I import that in the ETL process, I want to give it a specific name so that `run_v2g` is easy to follow.

I don't know if this is the best practice, so I am open to changing it.

src/run_v2g.py
ireneisdoomed: I tried sorting by position and the result took >6h. It is true that the current job removed the sorting step and included other improvements, so I cannot tell what is the absolute contribution of the sorting step. The step would be beneficial in L2G, where we join V2G features with other tables like the variant index, but I am not sure if it is worth the price. Should I give it a go? 

For the same reason, I am not sure about partitioning by type of evidence. Since we usually join this dataset on the variant, I think we want to keep partitioning by chromosome. The V2G source is not something that you expect to have on the right dataframe, right? And filtering on the source ID is usually a cheap step.

src/run_v2g.py
ireneisdoomed: Exactly, that is the main reason why I think it makes sense to work with a single object. I'll try to test what is the cost of the performance of this operation.

src/etl/v2g/functional_predictions/vep.py
ireneisdoomed: `va_with_vep` has  440M rows in 400 partitions. `vi` has 6M rows in 200 partitions.
The intersection of both is a much smaller dataset with 4M rows that I don't need to have distributed in a lot of partitions, that is why I am downgrading to 20. I use 20 because I read that the number of partitions should be a multiple of the default number of nodes (200).

src/etl/v2g/functional_predictions/vep.py
ireneisdoomed: I forgot to add this, thanks :) Changed in https://github.com/opentargets/genetics_etl_python/commit/ebe6717b1207d1629685a71421b4cbce3c4b590b

ireneisdoomed:
I've done many runs of this to see the best approach since the review. In the end I got good results by concatenating all datasets and partitioning them **only once** (at the end) by chromosome and ordered by position.

<details>
<summary>A summary of the successful job runs</summary>
<br>

Job where outputs are unioned and filtered per biotype, repartitioned by chromosome, ordered by position and partitioned:
- 1h50
- [95d2efd1f54f41b99f1022e0d8e85a31](https://console.cloud.google.com/dataproc/jobs/95d2efd1f54f41b99f1022e0d8e85a31?region=europe-west1&project=open-targets-genetics-dev)
- gs://genetics_etl_python_playground/XX.XX/output/python_etl/parquet/v2g

Job where outputs are split on type of evidence, repartitioned by chromosome, ordered by position and partitioned:
- 1h50
- [438b05d49f7a4694aea286320facc36a](https://console.cloud.google.com/dataproc/jobs/438b05d49f7a4694aea286320facc36a?region=europe-west1&project=open-targets-genetics-dev)
- gs://genetics_etl_python_playground/XX.XX/output/python_etl/parquet/v2g_split

Job where outputs are unioned, and just partitioned by chromosome
   - 2h40
- [47da3d257709436f8c64954915174058](https://console.cloud.google.com/dataproc/jobs/47da3d257709436f8c64954915174058?region=europe-west1&project=open-targets-genetics-dev)
- gs://genetics_etl_python_playground/XX.XX/output/python_etl/parquet/v2g_frozen

Job where outputs are unioned, and just partitioned by chromosome but the intermediary interval data is sorted theoretically to optimise the joins
- 6h50
- [9dd81e80a766462b8e2360bc0fde8183](https://console.cloud.google.com/dataproc/jobs/9dd81e80a766462b8e2360bc0fde8183?region=europe-west1&project=open-targets-genetics-dev)
- gs://genetics_etl_python_playground/XX.XX/output/python_etl/parquet/v2g
<br><br>
</details>

Other than that I also made other changes:
- Taking the input/output logic out of the individual modules, so that shared dependencies like the variant index are only read once.
- Filter the dataset by biotype. One comment here: **when v2g is joined with the variant index, I only keep those assignments that are resolvable to ENSGID in the genes set (inner join)**. This means that we will be dropping all assignments that fall outside. Given the recent conversation we had, @DSuveges @d0choa, do you think it is a better idea having a left join instead and adding this condition as a flag in another column? Otherwise we can come back to this in the future when we have a better grasp of where are all the steps where we might be dropping data.

d0choa (approved):
Pending work will be actioned in subsequent PRs:
- Remove `label` from schema. In the case of the func. consequence derived dataset this will only be captured by `variantFunctionalConsequenceId` In the case of the polyphen and sift datasets, we will interpret these scores also in the `variantFunctionalConsequenceId`.
- Remove pmid from schema
- Remove isHighQualityPlof from schema, this info is redundant and can be derived from the scores already
- Modularise the `run_v2g.py` step into a function in the module

PR #43, opened 2022-11-16
feat: etl session to allow for optional config

d0choa:
not required, but it allows for providing a `sparkConf` if provided, otherwise uses defaults.

Simplifies development, when more complex sparkConf might be required (e.g. for hail)


d0choa:
I will include this as part of other branch

PR #45, opened 2022-11-21
Minor improvements to the variant datasets generation

ireneisdoomed:
These are optimisation improvements to both the variant index and annotation.

**Variant index**:
- Improved plan. We avoid one shuffling step in the credible set by partitioning the dataset on the chromosome and then providing the partitioning key also in step where I drop the duplicates.
- [Job](https://console.cloud.google.com/dataproc/jobs/e8c01368a0444dc3a068447b3887cf26?region=europe-west1&project=open-targets-genetics-dev): 6'. Very small improvement in comp time (~40sec) 

**Variant annotation**
- Improved output. Dataset is now partitioned by chromosome and ordered by position without a big detriment in the comp time.
- [Job](https://console.cloud.google.com/dataproc/jobs/12bd4aa3c9f64c43850211e58971d7e6/configuration?region=europe-west1&project=open-targets-genetics-dev): 2h40'

### DIFF
--- a/configs/etl/reference.yaml
+++ b/configs/etl/reference.yaml
@@ -53,7 +53,6 @@ variant_annotation:
     partition_count: 400
     partition_columns:
       - chromosome
-      - position
   machine: ${machine.variant_annotation}
 
 variant_index:
@@ -63,6 +62,10 @@ variant_index:
   outputs:
     variant_index: ${etl.outputs}/variant_index
     variant_invalid: ${etl.outputs}/invalid/credible_sets
+  parameters:
+    partition_count: 400
+    partition_columns:
+      - chromosome
   machine: ${machine.default}
 
 gwas_ingest:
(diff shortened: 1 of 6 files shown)


DSuveges (approved):
Things looks good and makes sense. However, what is not clear to me, if we need to change how the variant annotation dataset is used eg. when mapped to the GWAS Catalog associations.

ireneisdoomed:
@DSuveges I do not think so. The schemas has remained the same. The only thing which this could impact is ensuring that you include `chromosome` as a joining key. From what I have seen that is already how you join the dataset with the associations (and even though you broadcast it, so everything should remain the same). Thanks for the review!

d0choa:
I don't think it's a good idea to externalise the columns to partition as a configuration parameter. This should be part of the business logic as nobody should have a reason to alter it through configuration.

PR #47, opened 2022-11-22
Amend and separate the script for dependency installation

tskir:
This separates the dependency installation into a separate script and adds the following functionality:
* If necessary, install and configure Pyenv
* If necessary, install Poetry
* Autodetect and, if necessary, install the Python version required for the project
* Activate the appropriate Python version for the project


DSuveges (approved):
All looks reasonable. It makes sense to delegate environment building to a separate script if we consider the makefile temporary.

PR #59, opened 2023-01-16
[Draft] FinnGen ingestion

tskir:
Warning: this is untested, and EFO lookup is not yet implemented.

Once completed, will close https://github.com/opentargets/issues/issues/2771.


tskir:
PR's mostly ready, the only things which remain are to:
* Validate dataframe after generation
* Address the failing tests
* Test run of the entire updated pipeline

configs/etl/reference.yaml
```diff
@@ -137,3 +137,14 @@ gwas_ingest:
     min_r2: 0.5
     k: 6.4 # Empiric constant that can be adjusted to fit the curve, 6.4 recommended.
   machine: ${machine.default}
+
+finngen_ingest:
+  inputs:
+    phenotype_table_url: https://r7.finngen.fi/api/phenos
```
ireneisdoomed, line 143: Any reason why we are not using the R8 release?

configs/etl/reference.yaml
```diff
@@ -137,3 +137,14 @@ gwas_ingest:
...
+  outputs:
+    finngen_studies: ${etl.outputs}/finngen_studies
+  parameters:
+    finngen_release_prefix: FINNGEN_R7_
```
ireneisdoomed, line 147: Same as above

configs/etl/reference.yaml
```diff
@@ -137,3 +137,14 @@ gwas_ingest:
...
+    sumstat_url_prefix: https://storage.cloud.google.com/finngen-public-data-r7/summary_stats/finngen_R7_
```
ireneisdoomed, line 148: Same as above

src/etl/finngen_ingest/finngen_ingestion.py
```diff
@@ -0,0 +1,80 @@
...
+    json_data = urlopen(phenotype_table_url).read().decode("utf-8")
+    rdd = etl.spark.sparkContext.parallelize([json_data])
+    df = etl.spark.read.json(rdd)
+
+    # Select the desired columns.
+    df = df.select("phenocode", "phenostring", "num_cases", "num_controls")
+
+    # Rename the columns.
+    df = df.withColumnRenamed("phenocode", "studyId")
+    df = df.withColumnRenamed("phenostring", "diseaseTrait")
```
ireneisdoomed, line 55: According to the studies schema

```suggestion
    df = df.withColumnRenamed("phenostring", "traitFromSource")
```

ireneisdoomed (changes requested):
The logic to process the study table looks good to me. As I said, we are following this convention of concatenating as many functions as possible instead of splitting the steps, even though Pyspark would generally handle them similarly.

I am not approving it because of the Ontoma implementation, which I have questions about. We did a lot of manual curation on Finngen up to their R6 release, the plan would be to do it similarly for the R8 traits and store these mappings in our central repo. I think it would be a more efficient approach.

DSuveges (changes requested):
Besides my biggest concern (ontoma) things mostly fine. We can help with putting the disease mapping table together.

src/etl/finngen_ingest/finngen_ingestion.py
```diff
@@ -0,0 +1,80 @@
...
+    # Set constant value columns.
+    # Then f.when(f.lit(True)) trick makes sure that the column is created as nullable, to ensure that it is not flagged as incorrect by validate_df_schema. See: https://stackoverflow.com/a/68578278.
+    df = df.withColumn(
+        "initialSampleSize",
+        f.when(f.lit(True), f.lit("309,154 (173,746 females and 135,408 males)")),
+    )
+    df = df.withColumn("hasSumstats", f.when(f.lit(True), f.lit(True)))
+
+    # Compute the EFO mapping iformation.
+    df = df.withColumn("efos", ontoma_udf(df["diseaseTrait"]))
```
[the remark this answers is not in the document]
d0choa: 🙏

configs/etl/reference.yaml
tskir: Updated to use FinnGen release 8

configs/etl/reference.yaml
tskir: Updated to use FinnGen release 8

configs/etl/reference.yaml
tskir: Updated to use FinnGen release 8

src/etl/finngen_ingest/finngen_ingestion.py
```diff
@@ -0,0 +1,80 @@
...
+        sumstat_url_prefix (str): URL prefix for constructing the sumstat file download path.
+        sumstat_url_suffix (str): URL suffix for constructing the sumstat file download path.
+
+    Returns:
+        DataFrame: Parsed and annotated FinnGen study table.
+    """
+    # Read the JSON data from the URL.
+    json_data = urlopen(phenotype_table_url).read().decode("utf-8")
+    rdd = etl.spark.sparkContext.parallelize([json_data])
+    df = etl.spark.read.json(rdd)
```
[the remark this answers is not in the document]
tskir: Thank you Irene, this is a very good suggestion. Now applied

src/etl/finngen_ingest/finngen_ingestion.py
tskir: You're right, now updated

src/etl/finngen_ingest/finngen_ingestion.py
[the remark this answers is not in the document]
tskir: Agreed. The latest version of this PR does not use OnToma anymore

tskir:
@DSuveges @ireneisdoomed I've updated the PR. As discussed, it does not use OnToma anymore. I will create a separate issue for updating the format of the curations table and doing the manual curation

src/otg/finngen.py
```diff
@@ -0,0 +1,69 @@
...
+    - Trait name (for example, Amoebiasis)
+    - Number of cases and controls
+    - Link to the summary statistics location
+
+    Some fields are also populated as constants, such as study type and the initial sample size.
+    """
+
+    session: Session = Session()
+
+    def run(self: FinnGenStep) -> None:
```
DSuveges, line 29: I think it is not necessarily a blocker for merging the PR, but the organization of the logic in PR doesn't follow the existing implementation. Currently we have datasets, eg. [StudyIndex](https://github.com/opentargets/genetics_etl_python/blob/bff5c0d0c2124e7f89ad132d13f375fb891c5bea/src/otg/dataset/study_index.py#L28) whit methods to generate the dataset (and brings the logic to validate data) or there are other [classes](https://github.com/opentargets/genetics_etl_python/blob/bff5c0d0c2124e7f89ad132d13f375fb891c5bea/src/otg/dataset/study_index.py#L60) that inherit from the dataset. While the "run" scripts are just wrapper around these methods. Up to you @d0choa, @ireneisdoomed to request for change or not.

DSuveges (changes requested):
I understand that the curation of the diseases and phenotypes are not yet there, but I would still add the logic that would deal with even a mock mapping data (maybe just using the raw ontoma output). So that once the mapping is in place, we could just plug in. Otherwise the branch would need to be touched again. I assume the config also needs to be extend with path pointing to the mappings.

But let's hear what others say. If they are fine with the branch as it is, it can go.

src/otg/finngen.py
tskir: Thank you for pointing this out Daniel! This has now all been addressed

tskir:
@d0choa @DSuveges @ireneisdoomed 

Changes compared to the previously reviewed version of this PR:

* Implement the new StudyIndexFinnGen class
* Rewrite FinnGenStep to follow the new logic
* Implement FinnGen mock study index fixture
* Test schema compliance for StudyIndexFinnGen
* Prepare sample FinnGen study data
* Implement sample FinnGen study data fixture
* Test ingestion from source for StudyIndexFinnGen
* Add StudyIndexFinnGen documentation

Ready for another and, hopefully, final iteration of review ;-)

tskir:
@DSuveges Regarding ontology mappings, I think the best solution would be to merge this branch without them, and then do the ontology mappings as a separate issue (& separate PR).

The reason is that, because we want to display both primary and background traits, as those aren't currently supported anywhere, introducing those would require a series of coordinated PRs in OnToma, the mappings repository, and this repository.

It shouldn't take long, but I think if we organise those as separate RPs it will make tracking and reviews easier.

tskir:
P.S. I can see the tests are failing :-/ I'll address those tomorrow first thing in the morning...

ireneisdoomed (changes requested):
Thank you! I think it is almost there.
The only thing keeping me from approving is that the workflow to generate the dataset fails. There seems to be some configuration missing. You can see the full log [here](https://console.cloud.google.com/dataproc/jobs/my_finngen-iuxi34ervdscw/monitoring?region=europe-west1&project=open-targets-genetics-dev)

Other than that, I think that the structure and the logic makes sense. Tests are practically in place, documentation is in place, a minor comment aside.

src/otg/finngen.py
DSuveges: The circular dependency doesn't exist for other sources but the GWAS Catalog. It's due to the curation practices.

DSuveges (approved):
Looks good to me. Thanks.

d0choa (approved):

README.md
```diff
@@ -22,30 +22,26 @@
...
-```bash
-pyenv versions
-```
+Then run VS Code (`code`) and select the interpreter.
 
-Otherwise, install
+In some cases, Pyenv and Poetry may cause various exotic errors which are hard to diagnose and get rid of. In this case, it helps to remove them from the system completely before running the `make setup-dev` command. See instructions in [utils/remove_pyenv_poetry.md](utils/remove_pyenv_poetry.md).
 
-```bash
-pyenv install 3.8.15
```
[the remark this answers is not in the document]
tskir: I think, given that we do all our environment configuration seamlessly through Poetry and Pyenv, it shouldn't really matter to the end user which version of Python is it running under the hood. So same as for the other dependencies, I'm happy to just have it handled under the hood with no indication in the README

src/otg/finngen.py
```diff
@@ -0,0 +1,69 @@
...
+        """Run FinnGen study table ingestion step."""
+        # Read the JSON data from the URL.
+        json_data = urlopen(self.finngen_phenotype_table_url).read().decode("utf-8")
+        rdd = self.session.spark.sparkContext.parallelize([json_data])
+        df = (
+            self.session.spark.read.json(rdd)
+            # Select the desired columns.
+            .select("phenocode", "phenostring", "num_cases", "num_controls")
```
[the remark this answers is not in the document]
tskir: Thank you, that's really useful to know! Addressed in the latest version of the PR. Now just two `withColumn` statements remain out of the initial ~10

src/otg/dataset/study_index.py
```diff
@@ -412,3 +412,74 @@ def _annotate_discovery_sample_sizes(
...
+                .withColumn(
+                    "summarystatsLocation",
+                    f.concat(
+                        f.lit(finngen_sumstat_url_prefix),
+                        f.col("studyId"),
+                        f.lit(finngen_sumstat_url_suffix),
+                    ),
+                )
+                # Set constant value columns.
+                # Then f.when(f.lit(True)) trick makes sure that the column is created as nullable, to ensure that it is not flagged as incorrect by validate_df_schema. See: https://stackoverflow.com/a/68578278.
```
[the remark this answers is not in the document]
tskir: Thank you for letting me know, now addressed

src/otg/dataset/study_index.py
```diff
@@ -412,3 +412,74 @@ def _annotate_discovery_sample_sizes(
...
+                .withColumn(
+                    "initialSampleSize",
+                    f.when(
+                        f.lit(True),
+                        f.lit("342,499 (190,879 females and 151,620 males)"),
+                    ),
+                )
+                .withColumn("hasSumstats", f.when(f.lit(True), f.lit(True)))
```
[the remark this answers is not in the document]
tskir: Now also addressed :)

PR #62, opened 2023-02-02
Add tissue enrichment by CHEERS to ETL

xyg123:
I have clumped all the logic into tissue_enrichment/tissue_enrichment.py  the computation of tissue enrichment is in the cheers function. 

The new run_tissue_enrichment rule in the Makefile submits run_tissue_enrichment.py.


src/run_tissue_enrichment.py
```diff
@@ -0,0 +1,58 @@
...
+
+    summary_stats_enrichment = cheers(
+        tissue_annotations,
+        summary_stats_credible_set,
+    )
+
+    pics_enrichment = cheers(
+        tissue_annotations,
+        pics_credible_set,
+    )
```
DSuveges, line 43: Is there a reason to keep the sumstats and pics data streams separated? Unless there is a fundamental difference, I would suggest to pool the pics credset and sumstats credset datasets into one and run cheers once. Not sure if this move would pose any risk though.

configs/etl/reference.yaml
```diff
@@ -93,3 +93,15 @@ gwas_ingest:
...
   machine: ${machine.default}
+
+tissue_enrichment:
+  inputs:
+    tissue_annotations: gs://genetics-portal-dev-analysis/xg1/EPIMAP/CHEERS_annotations/Full_epimap.parquet
+    summary_stats_credible_set: gs://genetics-portal-dev-analysis/xg1/CHEERS_credible_set/FM_220401_hg19.parquet
+    pics_credible_set: gs://genetics-portal-dev-analysis/xg1/CHEERS_credible_set/Non_sumstat_hg19.parquet
+  outputs:
+    summary_stats_enrichment: gs://genetics-portal-dev-analysis/xg1/CHEERS_enrichments/sumstats_enrichments/results.parquet
+    pics_enrichment: gs://genetics-portal-dev-analysis/xg1/pics_enrichments/results.parquet
```
DSuveges, line 104: For the outputs, I would recommend to use the `${etl.outputs}/` root for placing files.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+from pyspark.sql.types import FloatType
+from pyspark.sql.window import Window
+from scipy import special as sps
+
+if TYPE_CHECKING:
+    from pyspark.ml.linalg import VectorUDT
+    from pyspark.sql import DataFrame
+
+
+def _melt(
```
DSuveges, line 20: I would recommend to add an example doctest here. Every steps seems to be reasonable, however it is hard to follow what is exactly happening.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+    # Add to the DataFrame and explode
+    _tmp = df.withColumn("_vars_and_vals", f.explode(_vars_and_vals))
+
+    cols = id_vars + [
+        f.col("_vars_and_vals")[x].alias(x) for x in [var_name, value_name]
+    ]
+    return _tmp.select(*cols)
+
+
+def _ndtr(x: float, mean: float, std: float) -> float:
```
DSuveges, line 56: I'm sure it is all fine, and there's a reason why this is the case, but when this function is called in line 93: `_ndtr(x, z, y)`, the type of these variables are `x: VectorUDT, y: VectorUDT, z: VectorUDT`. Is it fine to change the assumed type?

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+    Args:
+        x (VectorUDT): expected variance of peak rankings.
+
+    Returns:
+        float: Standard deviation of the mean of the n peak ranks overlapping GWAS SNPs.
+    """
+    return math.sqrt(x)
+
+
+def v_cdf(x: VectorUDT, y: VectorUDT, z: VectorUDT) -> float:
```
DSuveges, line 82: As this function always calls `_ndtr`, I would rename x,z,y to x, mean and std for a better understandability.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+        mean (float): Expected rank under the null hypothesis.
+        std (float): Standard deviation of the null.
+
+    Returns:
+        float: Area under the pdf.
+    """
+    return float(0.5 + 0.5 * sps.erf((x - mean) / (std * 2**0.5)))
+
+
+def m_sd(x: VectorUDT) -> float:
```
DSuveges, line 70: As this function is always called as udf, I would suggest to use with the `@udf(t.DoubleType())` decorator.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+            ),
+            how="inner",
+        )
+        .drop("chrom")
+        .select("study_id", "chr", "start", "end")
+        .distinct()
+        .persist()
+    )
+
+    mean_sd = f.udf(m_sd, t.DoubleType())
```
DSuveges, line 125: This line can be saved if m_sd is decorated.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+        .join(
+            snps,
+            (
+                (f.col("chrom") == f.col("chr"))
+                & (f.col("pos") >= f.col("start"))
+                & (f.col("pos") <= f.col("end"))
+            ),
+            how="inner",
+        )
+        .drop("chrom")
```
DSuveges, line 119: You don't need to drop a column if it is not in the select.

src/etl/tissue_enrichment/tissue_enrichment.py
```diff
@@ -0,0 +1,195 @@
...
+        .count()
+        .withColumnRenamed("count", "count_peaks")
+    )
+    n_total_peaks = float(peaks_wide.count())
+    n_unique_peaks = n_unique_peaks.withColumn(
+        "mean_variance", ((float(n_total_peaks) ** 2) - 1) / (12 * f.col("count_peaks"))
+    )
+    n_unique_peaks = n_unique_peaks.withColumn(
+        "mean_sd", mean_sd("mean_variance")
+    ).drop("mean_variance")
```
DSuveges, line 140: I'm not 100% sure what is happening here, but I have a feeling much of this process can be sped up by using windows.

DSuveges (changes requested):
Mostly things look absolutely fine. There are a few bits here and there which can make things a bit nicer. I'm not 100% sure about the cheers logic though. It would be great to have a chat about that.

src/etl/tissue_enrichment/tissue_enrichment.py
xyg123: Seems to have worked fine when I tried it, but will do more testing to make sure, and change back as required.

src/etl/tissue_enrichment/tissue_enrichment.py
xyg123: This is estimating the expected rank mean and variance for each "peak" in the dataset as described[ in the paper:](https://www.nature.com/articles/s41588-019-0493-9#Sec9:~:text=where%20max%20and,overlap%20at%20random.)

tskir:
@xyg123 @DSuveges @d0choa This PR is now almost a year old and has accumulated a lot of conflicts with the base branch. In particular, the architecture has had major changes several times over. Are there plans to complete it in the new year, or can we shelve it & close for now?

xyg123:
There are no plans to complete this in the new year at the moment, I can stash a local copy of the branch, happy to shelf for now.

tskir:
Thank you @xyg123! I'll go ahead and close this PR then. However, I won't delete the branch — you can do it if/when you feel necessary :upside_down_face:

PR #64, opened 2023-02-10
Clumping refactor

DSuveges:
This PR explored the possibilities to resolve LD based clumping using a graph based approach relying on the GraphFrames library. It didn't work out. So a more rudimentary solution was implemented: all variants that are explained by other, more significant associations in high LD got the same general message, without any details, which variant is that.

BTW the graph approach worked, but it was not scaling well: the applied pregel algorithm was not converging, and took way more time than it was sensible (on 96 core machine it took 1h). For future reference, the 31cab8de2d0aa206211e578aa0fb701dd5e064b2 commit has the code. Also it has code to install custom jar on the dataproc cluster and install custom python packages which are not available through pip (this problem also raises issues against using GraphFrames in production). 

For the record the result of the graph based clumping is here: "gs://ot-team/dsuveges/deconvoluted" (PICS not resolved though)

**!! Warning: the code seems alright, but I haven't tried a start to end run.**


ireneisdoomed (approved):
Looks good to me, it is now significantly slimmer. It is a shame that the graph approach didn't work out.
I have also run it on the whole data without problems, taking 2h30.

PR #66, opened 2023-03-10
Do hydra

d0choa:
Several major changes:
- All business logic is now migrated to the new structure based on Datasets, Methods and Steps.
- CLI is now available with the name `otg`
- First working implementation of Google Cloud Dataproc Workflow to run sequence of steps.
- Hydra-based configuration now distinguishes default behaviours (CLI) from YAML-based configuration.
- Pytest is operational in the new structure. Some tests can still be rescued from the `deprecated_tests` directory
- Ability to check Dataset methods based on mock spark dataframes 
- Compatible with Hail-dependent steps when providing `session.hail_home` parameter
- Directory structure and names modified
- Extensive changes in documentation, including schemas for all datasets, contributors, etc.


src/otg/dataset/variant_index.py
```diff
@@ -0,0 +1,78 @@
...
+            "alternateAllele",
+            "chromosomeB37",
+            "positionB37",
+            "alleleType",
+            "alleleFrequencies",
+            "cadd",
+        ]
+        vi = cls(
+            path=path,
+            df=variant_annotation.df.select(
```
DSuveges, line 65: Shouldn't it be:
```suggestion
            _df=variant_annotation.df.select(
```

d0choa: yes. Not surprisingly tests did not pass here. I have a pending commit that will fix this a many more minor issues

src/otg/dataset/variant_index.py
```diff
@@ -0,0 +1,78 @@
...
+            "referenceAllele",
+            "alternateAllele",
+            "chromosomeB37",
+            "positionB37",
+            "alleleType",
+            "alleleFrequencies",
+            "cadd",
+        ]
+        vi = cls(
+            path=path,
```
DSuveges, line 64: In general what does it mean having `path` as a data feature? When do we need to access it? Once the data is instantiated, when is it accessed?

src/otg/dataset/variant_index.py
```diff
@@ -0,0 +1,78 @@
...
+            df=variant_annotation.df.select(
+                *unchanged_cols,
+                f.col("vep.mostSevereConsequence").alias("mostSevereConsequence"),
+                # filters/rsid are arrays that can be empty, in this case we convert them to null
+                nullify_empty_array(f.col("filters")).alias("filters"),
+                nullify_empty_array(f.col("rsIds")).alias("rsIds"),
+                f.lit(True).alias("variantInGnomad"),
+            ),
+        )
+        vi.validate_schema()
```
DSuveges, line 74: I believe, this validation might be superfluous shouldn't the dataset class's df setter takes care of the validation already?

tests/dataset/test_study_index.py
```diff
@@ -0,0 +1,69 @@
...
+    """Mock v2g dataset."""
+    schema = parse_spark_schema("studies.json")
+
+    data_spec = (
+        dg.DataGenerator(
+            spark,
+            rows=400,
+            partitions=4,
+            randomSeedMethod="hash_fieldname",
+            name="v2g",
```
DSuveges, line 27: Why is the name `v2g`? I guess this is a typo, but in general, what is the importance of this field? Taking a quick look at the [documentation](https://databrickslabs.github.io/dbldatagen/public_docs/APIDOCS.html), this field is optional and doesn't do much.

d0choa: Defaults to None. I'll get rid off all of them

ireneisdoomed (approved):
This is awesome. I have tested that the gene-index, variant-annotation, and LD indices run by firing up the workflow script like so `poetry run python workflow/workflow_template.py` after building the project.
3 notes:
- Steps run in parallel by default, which is great.
- Right now to run the steps that we want at will we have to modify the pipeline YAML (`dag.yaml`)
- My suggestion when testing the steps would be to run things one step at a time. This is because if one step fails, the other steps will also fail, even if they are not dependent on the step that failed.

**This branch has breaking changes with other steps. We will address them in independent PRs as we progress.**

PR #67, opened 2023-03-13
L2G progress based on Hydra branch

ireneisdoomed:
This PR includes:
- L2G feature factory with working distance features
- Parsing of L2G gold standard processing is working
- L2G step with Hydra's cli is working. There are 2 main configurations (run_mode):
    - Train ([job](https://console.cloud.google.com/dataproc/jobs/locus_to_gene-fnyzufrmcvm2i/monitoring?region=europe-west1&project=open-targets-genetics-dev)) - 5min currently based on only the distance to a gene's TSS 
    - Predict ([job](https://console.cloud.google.com/dataproc/jobs/locus_to_gene-umk4aci3zhm5g/monitoring?region=europe-west1&project=open-targets-genetics-dev)) - <2 mins
- Training after cross validation is working - we use the L2G yaml to test the desired parameters to be fine tuned
- Set up of the spark session now occurs at the step level (more here https://github.com/opentargets/genetics_etl_python/pull/210)
- Logic to generate colocalisation based features worked given simulated colocalisation results
    - Commented out at the moment because this data has not been generated and is likely to change 

## Step workflow
image

### Training: we build associations from gold standards
- Gold standards are parsed and annotated with the fixed features.
- Class imbalance
    - Positive: 500
    - Negative: 7461
- L2GFeatureMatrix is a table result of the concatenation in a wide format of all L2GFeatures 
image

### Inferencing: associations are annotated and a score per gene is calculated
image

---
(obsolete) POV of this PR :P 
IMG_4736


d0choa:
@DSuveges and I concluded the quality of the snow is not good enough ;)

d0choa (approved):
The time has reached... 😢 You might want to consider a squash. Up to you

ireneisdoomed:
Thank you for the review and all the suggestions @d0choa 
I am not squashing the branch because that would mean that I have authored the contributions from the past months. This makes sense when the features are isolated and we don't rebase the feature branch with the main branch.

I have been reading the reason for this behaviour: when a "merge" is performed, Git doesn't just look at the final state of the code, but also how that state was reached through the commits. If the "squash" combines all commits into a single one, Git sees that commit as a unique set of changes. Even though those changes may be the same than what's already in `main`, from Git's perspective, they come from a different origin (my "squashed" commit). This is not what I want, to I'll merge it as is and bear in mind I shouldn't have such a long lived branch in the future.

Thanks again!

PR #70, opened 2023-04-11
Validate schemas overlooking the nullability fields

ireneisdoomed:
This PR includes:
- [x] Changes to the validate_schema Dataset method to compare schemas between dataframes by looking at the names.
    - The functionality is essentially the same, only that we iterate on the column names, instead of on the properties of each schema field.
    - There are 2 checks in place: a) checking that the observed schema does not have any extra field; b) checking that the mandatory fields dictated by the expected schema are present in the observed_schema
- [x] Added 2 unit tests for each of the functionalities I describe above

Notes:
- I also had to tweak the `pytest_generate_tests` function. TIL that when we use pytest's metafuncs, we are defining a fixture that will be used in every test, unless specified otherwise.
- Mypy doesn't check the script that tests the schema. This is because mypy is crashing when checking this script, even with the newer version, hence I was not able to commit. It's not that the script is raising a typing error, but mypy is failing to deserialize something. I've found an [issue](https://github.com/python/mypy/issues/12257) in their repo that might be related. Full trace:
```
  File "/Users/irenelopez/.cache/pre-commit/repomwnp37wt/py_env-python3.8/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/Users/irenelopez/.cache/pre-commit/repomwnp37wt/py_env-python3.8/lib/python3.8/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 95, in main
  File "mypy/main.py", line 174, in run_build
  File "mypy/build.py", line 193, in build
  File "mypy/build.py", line 276, in _build
  File "mypy/build.py", line 2903, in dispatch
  File "mypy/build.py", line 3284, in process_graph
  File "mypy/build.py", line 3362, in process_fresh_modules
  File "mypy/build.py", line 2101, in load_tree
  File "mypy/nodes.py", line 397, in deserialize
  File "mypy/nodes.py", line 3689, in deserialize
  File "mypy/nodes.py", line 3630, in deserialize
  File "mypy/nodes.py", line 274, in deserialize
  File "mypy/nodes.py", line 854, in deserialize
  File "mypy/types.py", line 189, in deserialize_type

  File "mypy/types.py", line 2038, in deserialize
KeyError: 'unpack_kwargs'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ outputs/
 coverage.xml
 docs/assets/schemas/
 .cache/
+mock_data/
+notebooks/wandb
+src/wandb
--- a/config/datasets/gcp.yaml
+++ b/config/datasets/gcp.yaml
@@ -34,4 +34,4 @@ ld_index: ${datasets.outputs}/ld_index
 catalog_study_index: ${datasets.outputs}/catalog_study_index
 catalog_study_locus: ${datasets.outputs}/catalog_study_locus
 #templates
-ld_index_template: ${datasets.outputs}/gnomad_r2.1.1.{POP}.common.ld.variant_indices.parquet
+ld_index_template: ${datasets.outputs}/ld_indices/gnomad_r2.1.1.{POP}.common.ld.variant_indices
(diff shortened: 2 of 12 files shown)


ireneisdoomed:
Closing because other commits that are not relevant leaked into this branch.

PR #71, opened 2023-04-11
Validate schemas overlooking the nullability fields

ireneisdoomed:
This PR includes:
- [x] Changes to the validate_schema Dataset method to compare schemas between dataframes by looking at the names.
    - The functionality is essentially the same, only that we iterate on the column names, instead of on the properties of each schema field.
    - There are 2 checks in place: a) checking that the observed schema does not have any extra field; b) checking that the mandatory fields dictated by the expected schema are present in the observed_schema
- [x] Added 2 unit tests for each of the functionalities I describe above

Notes:
- I also had to tweak the `pytest_generate_tests` function. TIL that when we use pytest's metafuncs, we are defining a fixture that will be used in every test, unless specified otherwise.
- Mypy doesn't check the script that tests the schema. This is because mypy is crashing when checking this script, even with the newer version, hence I was not able to commit. It's not that the script is raising a typing error, but mypy is failing to deserialize something. I've found an [issue](https://github.com/python/mypy/issues/12257) in their repo that might be related. Full trace:
```
  File "/Users/irenelopez/.cache/pre-commit/repomwnp37wt/py_env-python3.8/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/Users/irenelopez/.cache/pre-commit/repomwnp37wt/py_env-python3.8/lib/python3.8/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 95, in main
  File "mypy/main.py", line 174, in run_build
  File "mypy/build.py", line 193, in build
  File "mypy/build.py", line 276, in _build
  File "mypy/build.py", line 2903, in dispatch
  File "mypy/build.py", line 3284, in process_graph
  File "mypy/build.py", line 3362, in process_fresh_modules
  File "mypy/build.py", line 2101, in load_tree
  File "mypy/nodes.py", line 397, in deserialize
  File "mypy/nodes.py", line 3689, in deserialize
  File "mypy/nodes.py", line 3630, in deserialize
  File "mypy/nodes.py", line 274, in deserialize
  File "mypy/nodes.py", line 854, in deserialize
  File "mypy/types.py", line 189, in deserialize_type

  File "mypy/types.py", line 2038, in deserialize
KeyError: 'unpack_kwargs'
```

### DIFF
--- a/src/otg/dataset/study_index.py
+++ b/src/otg/dataset/study_index.py
@@ -331,9 +331,7 @@ class StudyIndexGWASCatalog(StudyIndex):
             europeans_deconvoluted, on="projectId", how="outer"
         )
 
-        self.df = self.df.drop("discoverySamples", "replicationSamples").join(
-            parsed_ancestry_lut, on="projectId", how="left"
-        )
+        self.df = self.df.join(parsed_ancestry_lut, on="projectId", how="left")
         return self
 
     def _annotate_sumstats_info(
@@ -366,7 +364,7 @@ class StudyIndexGWASCatalog(StudyIndex):
         )
 
         self.df = (
-            self.df.drop("hasSumstats", "summarystatsLocation")
+            self.df.drop("hasSumstats")
             .join(parsed_sumstats_lut, on="projectId", how="left")
             .withColumn("hasSumstats", f.coalesce(f.col("hasSumstats"), f.lit(False)))
         )
@@ -412,7 +410,5 @@ class StudyIndexGWASCatalog(StudyIndex):
                 f.sum("sampleSize").alias("nSamples"),
             )
         )
-        self.df = self.df.drop("nCases", "nControls", "nSamples").join(
-            sample_size_lut, on="projectId", how="left"
-        )
+        self.df = self.df.join(sample_size_lut, on="projectId", how="left")
         return self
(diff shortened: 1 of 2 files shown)


d0choa (approved):
Do you think is worth considering the next 2 cases?

- use `StructField` `.dataType` method to compare if the expected and observed types are the same. I believe they are ignored at the moment. (e.g. `df.schema[0].dataType`)
- While we were comparing the whole `StructField` before, this PR only compares the names. This makes all nested information ignored. If I understand this correctly you would be ignoring all differences in nested columns. We might need some recursivity similar to this [chunk of code from Chispa](https://github.com/MrPowers/chispa/blob/e1ea9b714c1385baa522f34a050421eb7bc766b5/chispa/schema_comparer.py#L67).

Since you have some tests already, perhaps it's useful to add some of these cases and try to do test-driven development.

We can also merge and work separately if this unblocks other work. I'm approving just in case.

DSuveges:
I agree with @d0choa, we should only drop the nullability check not the type check.

ireneisdoomed:
@d0choa @d0choa Thank you so much for your comments. They were all on point. The new changes include:
- types of each field are now validated, not only the names (leaving apart only the nullability property)
- compatibility for nested validation. this is achieved by flattening the schema of both dataframes to extract their fields. the function is heavily inspired in the one we have to calculate the metrics.
- unit tests for: missing field, extra field, field of different type. run both in nested and not nested mock data.

.gitignore
d0choa, line 1: I don't think this belongs here

ireneisdoomed:
Last change: added an ad hoc check for duplicated fields in https://github.com/opentargets/genetics_etl_python/pull/71/commits/d0b34891e16bfa48aaf2c85170d562d8a70fe47c

PR #72, opened 2023-04-12
test: is in credset function

d0choa:
Can we double-check this test? I think it's doing the right thing but I would appreciate another pair of eyes.

### DIFF
--- a/src/otg/method/pics.py
+++ b/src/otg/method/pics.py
@@ -69,16 +69,8 @@ class PICS:
 
         Examples:
             >>> d = [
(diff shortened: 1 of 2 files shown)


ireneisdoomed:
I've simply added the range of probabilities into a separate function so that the doctest is lighter and is useful for documentation purposes.
Along the way I had to modify the `_is_in_credset` function to account for the cases where there is a single probability per `(chromosome, study_id, variant_id)` window. The fact that pics_postprob_cumsum has only one element means that skipping it with `f.lag(pics_postprob_cumsum, 1)` will set the probability to null, hence falling in the `otherwise` clause and always returning True.

ireneisdoomed (approved):
LGTM

PR #75, opened 2023-04-19
Test session streamline

d0choa:
Testing sessions are now configured to run as single-threaded operations and other optimisations to minimise the cost of initialising the spark session. This change also ensures that data frames are not shuffled making it easier to test the expected outputs. (This was otherwise failing when changing python/spark functions)


ireneisdoomed (approved):
Perfect, thanks!

PR #76, opened 2023-04-24
Improved validation of nested structures

ireneisdoomed:
This PR includes changes to avoid errors when dealing with nested schemas.

Taking a schema like this one as an example:
```
root
 |-- studyLocusId: string (nullable = false)
 |-- arrayOfStruct: array (nullable = false)
 |    |-- element: struct (containsNull = true)
 |    |    |-- tagVariantId: string (nullable = false)
 |-- arraySimple: array (nullable = true)
 |    |-- element: string (containsNull = true)
```

The way that the flattening function was working before, was that the higher level column was defined with everything that was nested inside, leading to potential conflicts when, for example, the order of the columns in the observed data was not the same:
```
[('studyLocusId', StringType()),
 ('arrayOfStruct', ArrayType(StructType(List(StructField(tagVariantId,StringType,false))),true)), True)),
 ('arrayOfStruct.tagVariantId', StringType()),
 ('arraySimple', ArrayType(StringType(), True))]
```

Now the high level array of structs only describes the fact that it is an array of structs (by hardcoding it), and the structure inside is dynamically flattened.
```
[('studyLocusId', StringType()),
 ('arrayOfStruct', ArrayType(StructType([]), True)),
 ('arrayOfStruct.tagVariantId', StringType()),
 ('arraySimple', ArrayType(StringType(), True))]
```

I have also changed the flattening function to return elements of type "Field", with the 2 properties that matter to us: `name` and `dataType`. This is to emulate the attributes that are available in a `pyspark.schema` by default. This also improves readability, since we are no longer working with tuples of 2 elements where x[0] = name and x[1] = data type.


src/otg/dataset/dataset.py
```diff
@@ -80,11 +80,11 @@ def validate_schema(self: Dataset) -> None:
...
         # Required fields not in dataset
-        required_fields = [
-            (x.name, x.dataType) for x in expected_schema if not x.nullable
-        ]
+        required_fields = [x.name for x in expected_schema if not x.nullable]
         if missing_required_fields := [
-            x for x in required_fields if x not in observed_fields
+            req
+            for req in required_fields
+            if not any(field.name == req for field in observed_fields)
```
d0choa, line 87: Wouldn't something like this work and clean the sourcery issue?
`if any(field.name != req ...)`

d0choa (approved):
Minor comments

src/otg/dataset/dataset.py
ireneisdoomed: No. Saying that `any(field.name != req ...)` will always* evaluate to True, because in this second loop you are iterating through the observed_fields.
Say that `required_fields = ["A", "B"]` and `observed_fields = ["B", "C"]`:

- When req = "A", the second loop will be `[True, True]`, so the any() condition will be True, which is correct because "A" is indeed missing. 
- When req = "B", the second loop will be `[False, True]`, so `any()` will be True and therefore "B" will count as missing, when it is not true.

By writing `not any()`, we are really evaluating that `all` in required are observed_fields. (But I can't use `all` in the second loop because I'm looping through the observed_fields). 

It is a bit of a tongue twister, but it is the best I got. The evaluation could be simple (`all(any(field.name == req for field in observed_fields) for req in required_fields)`), but the fact that we want to list which fields are missing makes the logic more complex.

.gitignore
```diff
@@ -5,5 +5,8 @@ outputs/
 **/.DS_Store
 .coverage
 coverage.xml
+mock_data
+notebooks/wandb
+src/wandb
```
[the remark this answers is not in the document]
ireneisdoomed: L2G's gitignore keeps appearing as guest star in my PRs 😅

PR #77, opened 2023-04-24
Redefinition of credible set annotation

ireneisdoomed:
This PR is coauthored with @d0choa, and addresses the main piece of logic that defines which is the credible set in a region.

It is a significant change in terms of logic and style. To define a credible set, we sort the variants in a locus by their posterior probability, informing about the most likely variants to be explaining that signal. For the definition of which variants are part of the 95% credible set, the cumulative sum of the probabilities (already sorted) is calculated, so that we pick which all variants that explain 95% of the association.

We realised that the current logic was not doing this, as for the calculation of the cumulative sum, we were iterating on each tagging variant within `credibleSet`, and aggregating over the `posteriorProbability`, i.e. the cumulative sum was always equal to the probability of that variant.

The implementation is complicated, as it involves making aggregations and iterating at the same time, to avoid having to flatten the schema (and then aggregating again). I've done my best at documenting the function, but please make comments if this is still not clear.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,8 +4,7 @@
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
       "source.organizeImports": true
-    },
-    "editor.defaultFormatter": "ms-python.black-formatter"
+    }
   },
   "python.terminal.launchArgs": [
     "-m",
@@ -33,7 +32,7 @@
   "python.linting.pydocstyleArgs": [
     "--convention=google"
   ],
-  "python.formatting.provider": "none",
+  "python.formatting.provider": "black",
   "isort.args": [
     "--profile",
     "black"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,8 @@
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
       "source.organizeImports": true
-    }
+    },
+    "editor.defaultFormatter": "ms-python.black-formatter"
   },
   "python.terminal.launchArgs": [
     "-m",
@@ -32,7 +33,7 @@
   "python.linting.pydocstyleArgs": [
     "--convention=google"
   ],
-  "python.formatting.provider": "black",
+  "python.formatting.provider": "none",
   "isort.args": [
     "--profile",
     "black"
--- a/src/otg/common/schemas.py
+++ b/src/otg/common/schemas.py
@@ -45,7 +45,7 @@ def flatten_schema(schema: StructType, prefix: str = "") -> list:
         ... )
         >>> df = spark.createDataFrame([("A", [{"tagVariantId": "varA"}]), ("B", [{"tagVariantId": "varB"}])], schema)
         >>> flatten_schema(df.schema)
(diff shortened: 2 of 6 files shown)


d0choa:
Any thoughts about the failing test?

ireneisdoomed:
@d0choa Yes. Although there are many, all except one are due to validation and it is now fixed with the changes that got merged this morning.
The only real one is `test_pics`, and I'm trying to solve it. Itt is due to the same reason that got me to start debugging the `annotate_credible_set` one: the `withField` logic has been changed between pyspark versions and now it its failing at defining the attributes for calculating [the posterior probability as a new field](https://github.com/opentargets/genetics_etl_python/blob/a4bb77fbfdf89a45f650404a43ff5987f08ff3ed/src/otg/method/pics.py#L160). It is exactly the same error that we see in `do_gcp_image_2_1`.

I was concerned the windowing function was the problem, but in a toy example I was able to add the field and apply the window. I think it is about how nested fields are referenced inside the transformation, it complains how we call `x.r2Overall`.  WIP.

ireneisdoomed:
Also important to note, this is a **breaking change**. This implementation'd break on Spark 3.1.5 because of the `slice` function having changed between versions.

ireneisdoomed:
https://github.com/opentargets/genetics_etl_python/pull/77/commits/cd4dee540f6207f2aa919c69dc4b3ae9cb1c7fb1
addresses important changes in the PICS implementation that were raised thanks to the failing tests:
- Relevant change in the implementation of the PICS method based on my understanding:
    - mu and standard error inform about the association signal between a neutral SNP and the trait (useful because it is an estimation of the p value of that association). This is now populating the `tagPValue` and `tagStandardError` nested fields. **Question: PICS doesn't perform conditional analysis, right? So this P value is not after any conditioning.**
    - The estimation of this association signal based on the correlation factor is _only valid for neutral SNPs_: SNPs that have a minimum r2 of 0.5 with the lead. When r2 is below that, the linear correlation is gone, therefore we should not calculate any posteriors.
- In a more technical aspect, the way we perform the calculation of the posterior probabilities for every SNP in a locus has changed:
    - The source of all problems was trying to perform aggregation functions and UDFs in the context of nested data. This is not the proper way to work with nested data.
        - As the main argument to having an aggregated `credibleSet` column is to have the whole context of a studyLocus in a single row, aggregation or partitioning is not necessary.
        - At the same time, UDFs are expected to be applied on rows, so having a nested UDF always led to evaluation problems.
    - The solution: to encapsulate all the logic to fine map a locus within the UDF. This is what is covered in `PICS._finemap`, the function is well documented so it should be easy to follow.

ireneisdoomed:
https://github.com/opentargets/genetics_etl_python/pull/77/commits/fed6ea9b440b12f4d1b7de826381280a458b0a22 involves a minor bug converting `r`to `r2` values;  gnomad provides LD scores as `r` however, `LDAnnotatorGnomad.weighted_r_overall` expects r^2 values.

We now calculate r^2 early on for consistency (the PICS functions already take this into consideration).

.vscode/settings.json
d0choa, line 1: is this needed?

d0choa (approved):

.vscode/settings.json
ireneisdoomed: my bad. changes got reverted

PR #80, opened 2023-05-09
docs: missing docs for finngen step added

d0choa:


tskir:
Oh my, I knew I would have missed something... Thank you! I'll be aware for the future PRs as well

PR #82, opened 2023-05-18
Implement version aware code deployment

tskir:
Closes https://github.com/opentargets/issues/issues/2946.

* Updated make build to upload code to a version specific path
* Updated workflow_template.py to run the code from that version specific path
* Replaced hardcoded parameters in workflow_template.py with customisable ones
* Amended developer documentation to make it clear how versioning works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 PROJECT_ID ?= open-targets-genetics-dev
 REGION ?= europe-west1
-BUCKET_NAME=gs://genetics_etl_python_playground/initialisation/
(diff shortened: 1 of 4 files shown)


tskir:
Reorganised `workflow_template.py` to make argparse and pytest work together and amended some small things. The tests are now passing, and the manual test run of the code was successful. Ready for review.

d0choa (approved):
All changes look very positive. I ran the code and it went flawlessly.

2 things to keep in mind:
- we can still override other's people code if we don't change the code version. In the future, if the build logic is together with the "workflow" logic we could upload the code to a directory with the prefix of the individual and avoid clashes. 
-  we should start moving part of the readme info to a contributing section in the docs.

Let's merge this!

PR #94, opened 2023-06-16
feat: add create-dev-cluster rule to makefile

ireneisdoomed:
New makefile rule to help remote development:
- Builds the project
- Creates a n1-standard-16 Dataproc cluster with the project installed. The name is based on the version

No maximum idle time, instead of sending jobs (which is also possible), my intention is to connect VSCode to the cluster and work remotely.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ PROJECT_ID ?= open-targets-genetics-dev
 REGION ?= europe-west1
 APP_NAME ?= $$(cat pyproject.toml| grep name | cut -d" " -f3 | sed  's/"//g')
 VERSION_NO ?= $$(poetry version --short)
+CLEAN_VERSION_NO := $(shell echo "$(VERSION_NO)" | tr -cd '[:alnum:]')
 BUCKET_NAME=gs://genetics_etl_python_playground/initialisation/${VERSION_NO}/
 
 .PHONY: $(shell sed -n -e '/^$$/ { n ; /^[^ .\#][^ ]*:/ { s/:.*$$// ; p ; } ; }' $(MAKEFILE_LIST))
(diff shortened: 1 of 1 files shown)


d0choa (approved):

Makefile
```diff
@@ -18,6 +19,19 @@ setup-dev: SHELL:=/bin/bash
 setup-dev: ## Setup development environment
 	@. utils/install_dependencies.sh
 
+create-dev-cluster: ## Spin up a simple dataproc cluster with all dependencies for development purposes
+	@${MAKE} build
+	@echo "Creating Dataproc Cluster"
+	@gcloud config set project ${PROJECT_ID}
+	@gcloud dataproc clusters create "ot-genetics-dev-${CLEAN_VERSION_NO}" \
```
DSuveges, line 26: I think this solution might lead to clashes: if users are not incrementing the project file version, multiple dev instances might be started with the same `CLEAN_VERSION_NO`. Also if the cluster has no automatic turn-off time set, the cluter might hang there without anyone knowing who is working on it. 

My suggestion was to use the ${USER} built in variable to append (or prepend) to cluster name, hence would avoid any clashes and the cluster at all times would be identifiable. 

What's your view on it?

DSuveges (commented):
I have one question on the naming of the cluster, which is not necessarily a deal breaker.

Makefile
ireneisdoomed: That's a good suggestion, didn't thought about predefined variables.
At the same time since the cluster is bounded to a specific version of the project (regardless if the version was properly named), I think it still makes sense to include in the cluster name.
The way I understand this, I'm considering these clusters as use-and-delete machines, they're not intended to live for a long time. I'm happy to set up a max idle time of 8h, it would last for a working day. What do you think?

DSuveges: The main concern is that how multiple people can test different components at the same time given the version is not yet incremented. Or I might not understand the point of the dev machine. The deprecated makefile had this line:

```makefile
CLUSTER_NAME ?= ${USER}-genetics-etl
```

This ensured that the cluster is always built with the same environment the user is actually working with. (with even the packages the user is actually developing, which might not even committed). But as I said, I might missing the point.

DSuveges: OK. Fine with me.

PR #106, opened 2023-07-11
refactor: some minor issues sorted out around summary statistics

DSuveges:
I have moved some of the methods initially shipped with summary statistics, but thinking about them more closely, they can be re-used in the other processes eg. when ingesting curated associations.

Now GWAS Catalog harmonized sumstats can be ingested as follows:

```python
study_id = 'GCST90016941'
file = f'{sumstats_path}/33462485-GCST90016941-EFO_0007874.h.tsv.gz'
pvalue_threshold = 5e-6
distance_threshold = 2.5e5

df = spark.read.csv(file, sep='\t', header=True)

leads = (
    SummaryStatistics
    .from_gwas_harmonized_summary_stats(df, study_id)
    .pvalue_filter(pvalue_threshold)
    .window_based_clumping(distance_threshold)
)

print(type(leads))
leads.df.printSchema()
leads.df.show()
```

### DIFF
--- a/src/otg/dataset/summary_statistics.py
+++ b/src/otg/dataset/summary_statistics.py
@@ -1,7 +1,6 @@
 """Summary satistics dataset."""
 from __future__ import annotations
 
-import sys
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


d0choa (approved):
Looks good. If the helper functions become too messy or we identify any structure we can always create some classes with utils e.g `GWASUtils`

src/otg/dataset/summary_statistics.py
```diff
@@ -175,7 +102,9 @@ def from_gwas_harmonized_summary_stats(
             _df=processed_sumstats_df,
         )
 
-    def calculate_confidence_interval(self: SummaryStatistics) -> SummaryStatistics:
+    def calculate_confidence_interval_for_summary_statistics(
```
ireneisdoomed, line 105: This function is not used at the moment.
If possible, I think it is a good idea if the SummaryStatistics dataset reports the diff confidence interval bounds, and drops `standardError` from the schema (unless it is used downstream).

ireneisdoomed (approved):
Great restructure, thanks!
I have a suggestion that affects the main logic of `gwas_catalog_sumstat_preprocess.py` that we might want to consider for a diff PR. 
I also added a test for `calculate_confidence_interval`, I hope that's alright!
Approving.

src/otg/dataset/summary_statistics.py
DSuveges: I dropped this function. The logic to generate the confidence intervals for the betas is quite source specific, so I think there's no need to have a shared function.

ireneisdoomed: Thanks!

PR #108, opened 2023-07-11
Fixes to the `annotate_ld` function

ireneisdoomed:
This PR contains:
- fixes to close https://github.com/opentargets/issues/issues/3019
- changes to adjust the r2 threshold for the LD matrix to annotate all variants with a correlation of .2. The threshold of .5 is maintained once the r2Overall is calculated. __Note: allowing for a more flexible threshold didnt impact the step runtime.__

**As noted in https://github.com/opentargets/genetics_etl_python/pull/105, the function to annotate the LD set will change with the new LDIndex. These changes were mostly fixes on how to operate with the LDIndex. When the new dataset is in place, I expect that most of this logic will be affected (for the better).**

### DIFF
--- a/src/otg/dataset/study_locus.py
+++ b/src/otg/dataset/study_locus.py
@@ -1549,6 +1549,13 @@ class StudyLocusGWASCatalog(StudyLocus):
             min_r2=0.2,
         ).coalesce(400)
 
+        credible_set_schema = [
+            (subfield.name, subfield.dataType)
+            for field in self._schema.fields
+            if field.name == "credibleSet"
+            for subfield in field.dataType.elemenType  # type: ignore
+        ]
+
         ld_set = (
             self.unique_study_locus_ancestries(studies)
             .join(ld_r, on=["chromosome", "variantId", "gnomadPopulation"], how="left")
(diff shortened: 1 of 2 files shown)


src/otg/dataset/study_locus.py
```diff
@@ -1564,12 +1571,20 @@ def annotate_ld(
...
+            .filter(f.col("r2Overall") >= min_r2)
             .groupBy("chromosome", "studyId", "variantId")
             .agg(
                 f.collect_set(
                     f.when(
                         f.col("tagVariantId").isNotNull(),
-                        f.struct("tagVariantId", "r2Overall"),
+                        f.struct(
+                            [
+                                f.lit(None).cast(datatype).alias(field)
```
d0choa, line 1582: I don't think this is a good approach, but it's better not to block this PR considering the upcoming changes

src/otg/dataset/study_locus.py
```diff
@@ -1546,9 +1546,16 @@ def annotate_ld(
             ld_populations,
             ld_index_template,
             ld_matrix_template,
-            min_r2,
+            min_r2=0.2,
```
d0choa, line 1549: If you add a default it should be documented in the docstring

d0choa (approved):
Good to go

PR #110, opened 2023-07-14
Precompute LD sets for all possible variants

ireneisdoomed:
This PR includes the logic to close https://github.com/opentargets/issues/issues/3020 and https://github.com/opentargets/issues/issues/3037

- Logic to ingest gnomad's LD Matrices from 8 populations to dump all correlations that surpass a **r^2 threshold of 0.5**. This is then combined with the variant indices table to resolve the variantIds.
- Aggregation at the variant level of the LD Set as proposed [here](https://github.com/opentargets/issues/issues/3020#issuecomment-1636053704). In the end `ldSet` contains at least with all linked variants. Inside each `tagVariantId`, we have another list of what are the `r` values per population.
- Also quite a few of optimisations in the way we set up the cluster + the Spark configuration:
    - This module has worked on a **n1-highmem-96**  machine with a **disk boot size of 5TB** and **no SSDs**. I've changed `workflow_template` to be able to parametrise this from the CLI.
    - I encountered disk memory errors, so made a few changes, all more explained [here](https://github.com/opentargets/issues/issues/3037):
        - Set `openCostInBytes ` and `maxPartitionBytes ` to 50GB as recommended by the Hail docs. 
        -  to account for the local-dirs usable space error, I've changed this memory threshold to 98% (https://github.com/opentargets/genetics_etl_python/commit/34694342dd40630a589620b8bea84fbc968e22ac)
        - I increased driver and executor memory, not to use the default parameters and allocate them based on the resources of the machine ([9c8bc12](https://github.com/opentargets/genetics_etl_python/pull/110/commits/9c8bc1280da69f4c3814a6051a869ee04c46e8a9))
        - Even though there was a local SSD assigned, I saw that the disk type was still pd-standard. I explicitly told the cluster to use the SSD. (https://github.com/opentargets/genetics_etl_python/commit/df70a0d295846b6da3158d3cb70e4ed19b81d050)
        - YARN assigns 2 executors per node by default. This seemed like a waste of resources, so I set parameters to assign a dynamic number of executors based on the available resources. I went from 2 to 11 executors + driver. The script was much much quicker.(https://github.com/opentargets/genetics_etl_python/commit/7c21ee530bfe715290e5fe816d378c89c170e30f)

This is the [job](https://console.cloud.google.com/dataproc/jobs/my_ld_index-tvs42edrdg2k6?region=europe-west1&project=open-targets-genetics-dev) to generate the final dataset. It took **5h**.

## Data QC
Location of the data: `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/ld_index`

Final schema:
```
root
 |-- variantId: string (nullable = false)
 |-- chromosome: string (nullable = false)
 |-- ldSet: array (nullable = false)
 |    |-- element: struct (containsNull = false)
 |    |    |-- tagVariantId: string (nullable = false)
 |    |    |-- rValues: array (nullable = false)
 |    |    |    |-- element: struct (containsNull = false)
 |    |    |    |    |-- population: string (nullable = false)
 |    |    |    |    |-- r: double (nullable = false)
```

- The index consists of **33_997_554** variants.
- Variants have an average ldSet of size 68 (minimum is 1, maximum is 5425).
- We have variants from alternate contigs. I haven't filtered them out because they only represent 20_000 variants. **Question: do we want to filter these out?**
Any other metrics of interest?


d0choa (approved):
Important consequences of this PR to keep in mind:
- `StudyLocus` contains an `ldSet` column in addition to the `credibleSet`
- There is a bug on the calculation of PICS that causes most of our `credibleSet` to only contain the lead variant
- Spark session configuration was slightly adjusted to run this step. Changes have not been included in this PR

PR #112, opened 2023-07-26
New LD annotator for GWASCatalog curated studies

ireneisdoomed:
This PR contains:
- Logic for the new LDAnnotator, based on the work described here https://github.com/opentargets/issues/issues/3020#issuecomment-1636053704
- Logic changes to assign an LD set to every GWAS Catalog curate study/locus pairs. We now have 2 fields to differentiate the result of expanding the locus (`ldSet`) from the variants that belong to the credible set (`credibleSet`). The latter one is a subset of `ldSet`. 
- Adaptation of PICS.finemap 
- Minor bugs

The new GWASCatalog step runs in 12 minutes (as opposed to > 2h previously) and it is much simpler to follow. 
Some stats from the latest data (available at `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/catalog_study_locus`)
- 434,351 associations
- 47,326 associations with a null ldSet because they involve variants outside the LDIndex. This is now appropriately flagged.
- Among the associations with LD information, some stats about the sizes of the ldSet array (`sizeLd`) vs the credibleSet array (`credibleSet`) say that the average number of tagging variants is ~60, whereas after finemapping, if we stick to only the ones from the 95% credible set, the average size is below 1*.
```
+-------+-----------------+------------------+
|summary|           sizeLd|            sizeCs|
+-------+-----------------+------------------+
|  count|           387025|            387025|
|   mean|62.03514760028422|0.9171939797170725|
| stddev|161.1866631616384|2.5366012217228615|
|    min|                0|                 0|
|    max|             4366|              1187|
+-------+-----------------+------------------+
```
- This basically means that most of the credible sets are formed by a single variant (the lead SNP). This is true for 85% of the associations, since the calculated posterior probability is heavily biased by the r2 value. After discussing with Yakov, we've agreed that this is an undesired limitation from PICS, that we can't do much about it, and that it would be a very nice to compare between fine mapping methods once we implement them.

*It is below 1 because there are associations with an empty ldSet when the association has been clumped.

Questions:
- Is there anything we can do to avoid the singular credible sets?
- Is it useful to keep the `ldSet` field? This is basically a list of all tagging variants
- Do we want to empty the `ldSet`s from the clumped associations?

### DIFF
--- a/config/step/my_gwas_catalog.yaml
+++ b/config/step/my_gwas_catalog.yaml
@@ -7,7 +7,6 @@ catalog_ancestry_file: ${datasets.catalog_ancestries}
 catalog_associations_file: ${datasets.catalog_associations}
 catalog_sumstats_lut: ${datasets.catalog_sumstats_lut}
 variant_annotation_path: ${datasets.variant_annotation}
+ld_index_path: ${datasets.ld_index_path}
 catalog_studies_out: ${datasets.catalog_study_index}
 catalog_associations_out: ${datasets.catalog_study_locus}
-ld_matrix_template: ${datasets.ld_matrix_template}
-ld_index_template: ${datasets.ld_index_template} # TODO: remove
--- a/config/step/my_gwas_catalog.yaml
+++ b/config/step/my_gwas_catalog.yaml
@@ -7,6 +7,6 @@ catalog_ancestry_file: ${datasets.catalog_ancestries}
 catalog_associations_file: ${datasets.catalog_associations}
 catalog_sumstats_lut: ${datasets.catalog_sumstats_lut}
 variant_annotation_path: ${datasets.variant_annotation}
-ld_index_path: ${datasets.ld_index_path}
+ld_index_path: ${datasets.ld_index}
 catalog_studies_out: ${datasets.catalog_study_index}
 catalog_associations_out: ${datasets.catalog_study_locus}
--- a/config/step/my_gwas_catalog.yaml
+++ b/config/step/my_gwas_catalog.yaml
@@ -10,4 +10,4 @@ variant_annotation_path: ${datasets.variant_annotation}
 catalog_studies_out: ${datasets.catalog_study_index}
 catalog_associations_out: ${datasets.catalog_study_locus}
 ld_matrix_template: ${datasets.ld_matrix_template}
-ld_index_template: ${datasets.ld_index_template}
+ld_index_template: ${datasets.ld_index_template} # TODO: remove
--- a/src/otg/assets/schemas/ld_index.json
+++ b/src/otg/assets/schemas/ld_index.json
@@ -8,6 +8,12 @@
     },
     {
       "metadata": {},
+      "name": "chromosome",
+      "nullable": false,
+      "type": "string"
+    },
+    {
+      "metadata": {},
       "name": "ldSet",
       "nullable": false,
       "type": {
--- a/src/otg/assets/schemas/study_locus.json
+++ b/src/otg/assets/schemas/study_locus.json
@@ -114,6 +114,32 @@
     },
     {
       "metadata": {},
+      "name": "ldSet",
+      "nullable": true,
+      "type": {
+        "containsNull": true,
+        "elementType": {
+          "fields": [
+            {
+              "metadata": {},
+              "name": "tagVariantId",
+              "nullable": true,
+              "type": "string"
+            },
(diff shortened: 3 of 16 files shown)


xyg123:
**- Is there anything we can do to avoid the singular credible sets?**
85% singular credible sets seems very high, if this is due to the PICS method, are there parameters that can be tweaked?  Are we only running PICS using SNPs past a LD threshold? 

**- Is it useful to keep the ldSet field? This is basically a list of all tagging variants**
I can think of some uses for it in our new features, for example calculating the SNP-based heritability using LDSC, it adjusts association statistics using the number of SNPs in LD, but it would require the r2 of each SNP too.

**- Do we want to empty the ldSets from the clumped associations?**
Let x and y be two associations that are clumped together, will y appear in the ldSet of x?

ireneisdoomed:
> **- Is there anything we can do to avoid the singular credible sets?** 85% singular credible sets seems very high, if this is due to the PICS method, are there parameters that can be tweaked? Are we only running PICS using SNPs past a LD threshold?

PICS, as I understand it, is fundamentally based at inferencing the signal of the tagging SNPs based on LD. Fine mapping is the Bayes Theorem application. We could try tweaking the value for `k`, an empirical constant to calculate the posterior probability. Other than that, I don't know.

I have been comparing our new results, vs the result in the [PICS webtool](https://pics2.ucsf.edu/pics2-LD.html) for [11_68433827_C_T](https://genetics.opentargets.org/Variant/11_68433827_C_T/associations) and [bone mineral density](https://genetics.opentargets.org/Study/GCST000182/associations), a study with europeans.

#### Our PICS implementation
Credible set is only conformed by the lead variant because the PP is **1**.
```
 variantId       | 11_68433827_C_T
 ldSetSize       | 94
 tagVariantId    | 11_68433827_C_T
 neglogpvalue    | 11.221848749616356
 pics_mu         | 11.221848749616367
 pics_std        | -4.82433760706544E-4
 credibleSetSize | 1
 tagging         | {true, true, null, 1.0, 11_68433827_C_T, 5.999999999999856E-12, null, null, 0.9977000638225533, null, null, 1.0000000000000009}
```

#### PICS
If I go to https://pics2.ucsf.edu/pics2-LD.html and query `rs3736228` with a -logPValue of 11.22 using the European reference, I get a PICS posterior probability of **0.4656**.

#### Our PICS implementation after a bugfix
I've identified a bug in the formula used to get the PICS standard deviation. After correcting, I get bigger credible sets, however our posteriors still differ very much from what they should look like (PP = **0.22**).
```
 variantId       | 11_68433827_C_T
 ldSetSize       | 94
 tagVariantId    | 11_68433827_C_T
 neglogpvalue    | 11.221848749616356
 pics_mu         | 11.221848749616367
 pics_std        | -4.82433760706544E-4
 credibleSetSize | 37
 tagging         | {true, true, null, 0.22402647471149376, 11_68433827_C_T, 5.999999999999856E-12, null, null, 0.9999997927905098, null, null, 1.0000000000000009}
```
This is still not good. I'm opening a PR with the fix. I'd appreciate if you could take a look at the application of the Bayes theorem with the survival function. The formulae for the mean and st. deviation of the distribution follow the definitions from here https://pubs.broadinstitute.org/pubs/finemapping/assets/files/Supplementary%20Information.pdf (page 13)

> **- Is it useful to keep the ldSet field? This is basically a list of all tagging variants** I can think of some uses for it in our new features, for example calculating the SNP-based heritability using LDSC, it adjusts association statistics using the number of SNPs in LD, but it would require the r2 of each SNP too.

Thanks

> **- Do we want to empty the ldSets from the clumped associations?** Let x and y be two associations that are clumped together, will y appear in the ldSet of x?

Yes. If x is more significant than y, association with x will be flagged and considered redundant.

ireneisdoomed:
I've rerun the GWASCat ingestion with the full LDIndex. The process went only up to 15min, so very happy ([job](https://console.cloud.google.com/dataproc/jobs/my_gwas_catalog-zdn5pebpjjqpk?region=europe-west1&project=open-targets-genetics-dev)).
Data's been updated.

d0choa (approved):
Important consequences of this PR to keep in mind:
- `StudyLocus` contains an `ldSet` column in addition to the `credibleSet`
- There is a bug on the calculation of PICS that causes most of our `credibleSet` to only contain the lead variant

PR #113, opened 2023-07-27
Correct formula for PICS standard deviation

ireneisdoomed:
The formula used to calculate the PICS standard deviation was wrong due to the numerous parenthesis.
We were calculating this:
image

Instead of what [PICS defines](https://github.com/opentargets/genetics_etl_python/pull/112#issuecomment-1654101846):
image

I am extracting the absolute value of the st. deviation, because the survival function cannot be defined with negative values.
More details and context in this comment https://github.com/opentargets/genetics_etl_python/pull/112#issuecomment-1654101846

### DIFF
--- a/src/otg/method/pics.py
+++ b/src/otg/method/pics.py
@@ -64,7 +64,7 @@ class PICS:
             None
         """
         return (
-            ((1 - (abs(r2) ** 0.5) ** k) ** 0.5) * (neglog_p**0.5) / 2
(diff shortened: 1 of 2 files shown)


d0choa:
@addramir can you confirm if this is the right formula?

addramir:
Just to confirm: old formula used: r2^(0.5^k) and the new one has: r2^(k/2)?

d0choa:
@ireneisdoomed you are unblocked here

PR #115, opened 2023-08-07
Set `Dataset` as an abstract class

ireneisdoomed:
This PR includes:
    - Changes to the `Dataset` class to set it as an abstract base class. This means that Dataset is not intended to be instantiated directly, only used as a base class to define children classes (which is what we do). This way `from_parquet` (and other subsequent methods) can be abstracted out from having a specific `schema` attribute.
    - Changes to all children classes to define their own `_get_schema` method (which is the abstract method of Dataset), and therefore inherit `Dataset.from_parquet`
    - Changes needed to the testing fixtures that are an effect of the changes above.


d0choa (approved):
Now that conflicts are resolved, it's all looking good

Quite a boost in testing coverage by removing all the read_parquet functions

PR #122, opened 2023-08-21
feat: Airflow managed workflow

d0choa:
First implementation of Airflow DAG (managed by Cloud Composer). This could eventually replace Google Dataproc Workflow.


ireneisdoomed:
just a little more devops and you could change team @d0choa

PR #124, opened 2023-08-22
feat: add `StudyLocusOverlaps` step to the project

ireneisdoomed:
This PR contains the code necessary to output a dataset of overlapping signals that share common variants in their credible sets.

## Context
The initial purpose was to help @xyg123 use this as part of his pattern based colocalisation. Based on the discussion, I've learnt today that this is not the case for 2 main reasons:
- At this stage of the process we don't have credible sets, as this is a preliminary step to identify common signals prior to finemapping
- The proposed method, SMR-Theta, calculates the overlaps by looking at all variants in the region - so it is distance-based, rather than LD-based.
So this definition of an overlap dataset will fit in a different branch of work, and will have its own specification in a separate ticket.

The scope of this specific dataset is then to generate data to follow what we currently do in production (f[or example](https://genetics.opentargets.org/study-locus/GCST004365_195/1_154445939_T_C), last widget). Regardless of whether we decide to display it or not, it has use as input in the COLOC and eCAVIAR application.

## Features
- Updated the schema to include all statistics related to the tagging variant. This now lives in a `statistics` struct that we are carriying over to be used later on in coloc.
    - This field is technically redundant, since we have here metadata from summary statistics. Given the effort in describing what a locus consists along its metadata ([#3045](https://github.com/opentargets/issues/issues/3045)), we could drop this field. Later on, if we need to know what is the beta of a specific tagging variant, we go back to the studyLocus dataset.
- No breaking changes in downstream processes (coloc and ecaviar)

## Comments
- As suggested, this step assumes that all studyLocus and study indices can be loaded from a single path respectively. That is not the case as of today.
For the studies we have:
```
- /XX.XX/ukbiobank_study_index
- XX.XX/finngen_study_index
- XX.XX/catalog_study_index
```
I'd rather have a single bucket pointing to a datatype such as:
```
- /XX.XX/study_index
    - XX.XX/study_index/finngen_study_index
    - XX.XX/study_index/catalog_study_index
```
(same goes for studyloci)

## QC
 So far, I've prototyped the logic by looking at the overlaps in the GWASCatalog top hits. This is what is available here: `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/study_locus_overlap`. Some stats:
- 2,353,100 rows that represent the 2 overlapping studyLocusIds + the common tagging variant
- 2,121,633 distinct pairs of studyLocusIds

### DIFF
--- a/config/datasets/gcp.yaml
+++ b/config/datasets/gcp.yaml
@@ -31,6 +31,7 @@ gene_index: ${datasets.outputs}/gene_index/gene_index
 variant_annotation: ${datasets.outputs}/variant_annotation
 variant_index: ${datasets.outputs}/variant_index
 study_locus: ${datasets.outputs}/study_locus
+study_locus_overlap: ${datasets.outputs}/study_locus_overlap
 colocalisation: ${datasets.outputs}/colocalisation
 v2g: ${datasets.outputs}/v2g
 ld_index: ${datasets.outputs}/ld_index
--- a/config/step/my_study_locus_overlap.yaml
+++ b/config/step/my_study_locus_overlap.yaml
@@ -2,6 +2,6 @@
 defaults:
   - study_locus_overlap
 # Additional config
-study_locus_path: ${datasets.outputs}/study_locus
+study_locus_path: ${datasets.outputs}/catalog_study_locus
 study_index_path: ${datasets.outputs}/catalog_study_index
-study_locus_overlap_path: ${datasets.outputs}/study_locus_overlap
+overlaps_index_out: ${datasets.outputs}/study_locus_overlap
--- a/config/step/my_study_locus_overlap.yaml
+++ b/config/step/my_study_locus_overlap.yaml
@@ -1,0 +1,7 @@
+# Default config
+defaults:
+  - study_locus_overlap
+# Additional config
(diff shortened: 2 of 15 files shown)


hlnicholls (commented):
Everything looks great to me and makes sense with how I understand the use of credible sets going into coloc - including checking how the data looks in `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/study_locus_overlap `it all looks like it can go straight into coloc for `coloc.abf()`.

As a small note for us in the future, this PR has a very similar schema to what goes in coloc’s tutorial inputs, but I’m seeing [here](https://chr1swallace.github.io/coloc/articles/a02_data.html) that if we want to do `coloc.susie()` we will also need LD

ireneisdoomed:
Thank you for the review @hlnicholls! Good to know we are pretty consistent to COLOC's approach. Just as I think we will eventually remove the  statistics struct, I would not add information about LD for now, as this can come from the StudyLocus dataset.

ireneisdoomed:
Just realised @hlnicholls is on holiday. I'm merging without her explicit approval since there are no tasks or comments pending, I hope it's ok!!

PR #125, opened 2023-08-31
Il distance overlaps

ireneisdoomed:
This PR includes logic to calculate a dataset of SL overlaps based on a distance threshold + changes to the former overlap logic that looked at the SL credible sets:
- TBC


ireneisdoomed:
This branch is old. I prefer to close it and commit the changes to a newer branch of the project.

PR #128, opened 2023-09-14
Window based clustering with carrying over locus around the semi indices

DSuveges:
- [x] Upon identifying semi indices based on distance, all snps (reaching a defined significance) are collected in the `locus` object.
- [x] Tests are written to ensure the correctness of the logic for both clumping and locus collection.
- [x] Groupby logic was removed from clumping to avoid crashing due to memory error.
- [x] summary stats are clustered to further reduce the size of the dataset.
- [x] Exclude region functionality was added to summary stats dataset
- [x] Joining identified semi indices to summary statistics and then collect locus
- [x] Tests are passing

Potential extra optimization:

- [ ] <s>locus collection should only be done once semi indices were identified, the windowing method needs to be generalized for unfiltered data. </s> - I did this and things got even worse.


src/otg/common/utils.py
d0choa, line 1: Niceeee

src/otg/dataset/summary_statistics.py
```diff
@@ -122,14 +123,71 @@ def pvalue_filter(self: SummaryStatistics, pvalue: float) -> SummaryStatistics:
...
-        return WindowBasedClumping.clump(self, distance)
+        return WindowBasedClumping.clump_with_locus(
+            self,
+            window_length=distance,
+            p_value_significance=gwas_signiciance,
+            p_value_baseline=baseline_significance,
+        )
+
+    def window_based_clumping(
+        self: SummaryStatistics, distance: int, gwas_signiciance: float = 5e-8
```
d0choa, line 152: At least at this level, we could make the `include_locus` a parameter of the function `window_based_clumping` function.

DSuveges: You right, I'll change that.

src/otg/method/window_based_clumping.py
```diff
@@ -181,158 +184,183 @@ def _find_peak(position: ndarray, window_size: int) -> DenseVector:
...
+        sumstats_baseline = filtered_summary_stats.pvalue_filter(p_value_baseline).df
+
+        sumstats_baseline_renamed = (
+            # Renaming columns
+            reduce(
+                lambda df, column: df.withColumnRenamed(column, f"tag_{column}"),
+                columns,
+                sumstats_baseline,
+            )
+            # Drop associatinos above the baseline:
```
ireneisdoomed, line 323:
```suggestion
            # Drop associations above the baseline:
```

src/otg/method/window_based_clumping.py
```diff
@@ -181,158 +184,183 @@ def _find_peak(position: ndarray, window_size: int) -> DenseVector:
...
+            p_value_significance=p_value_significance,
+        ).df.alias("clumped")
+
+        # Extract column names:
+        columns = filtered_summary_stats.df.columns
+
+        # Dropping variants not meeting the baseline criteria:
+        sumstats_baseline = filtered_summary_stats.pvalue_filter(p_value_baseline).df
+
+        sumstats_baseline_renamed = (
```
ireneisdoomed, line 316: If I understand correctly, you are applying reduce to perform the renaming operation as many times as columns we have. 
As a suggestion, i think you can achieve the same with a single select statement. I'm more used to this solution, so for me it is more readable, but please go for whatever you like best

```suggestion
        sumstats_baseline_renamed = sumstats_baseline.selectExpr(*[f"{col} as tag_{col}" for col in sumstats_baseline.columns])

```

ireneisdoomed (commented):
Thank you so much! I haven't looked at everything yet, but I have some comments you might find useful

src/otg/common/utils.py
```diff
@@ -17,13 +17,57 @@
     from pyspark.sql import Column
 
 
+def parse_region(region: str) -> Tuple[str, int, int]:
```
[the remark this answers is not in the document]
DSuveges: I would assume every dataset that deals with genomic position should allow for region based filtering and exclusion. It's not yet generalized enough, but that is a intuitive expectation.

src/otg/dataset/summary_statistics.py
```diff
@@ -122,14 +123,71 @@ def pvalue_filter(self: SummaryStatistics, pvalue: float) -> SummaryStatistics:
...
+
+        Returns:
+            StudyLocus: _description_
+        """
+        # Calculate distance-based clumping:
+        return WindowBasedClumping.clump(
+            self, window_length=distance, p_value_significance=gwas_signiciance
+        )
+
+    def exclude_region(self: SummaryStatistics, region: str) -> SummaryStatistics:
```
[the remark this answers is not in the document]
DSuveges: I was also thinking about that, however to be precise with that, we should have something like a region object as hail has. I'm not 100% sure about that.

src/otg/method/window_based_clumping.py
DSuveges: This is great.

src/otg/dataset/summary_statistics.py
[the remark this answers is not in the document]
d0choa: There are different ways of improving this. One option to comes to my mind, it's similar to how we handle the QC's in GWASCatalog. A class that contains a list of predefined regions we concluded are potentially worth excluding. It won't be flexible, but it would ensure things are readable.

In any case, this is not critical for this PR

src/otg/dataset/summary_statistics.py
d0choa: Let's do it in another PR

d0choa (approved):
A few thoughts in the PR review we could come back later. Test coverage (100%).
There is a pending question about performance and scalability that we will resolve if problems arise.

PR #130, opened 2023-09-15
Docs build is now tested

d0choa:
- mkdocs build is now tested in strict mode, so we will now know if there is an issue in the docs before merging
- it includes a couple of fixes to ensure strict mode:
  - some links pointing nowhere
  - deprecated plugin removed


ireneisdoomed (approved):
Really nice, thanks a lot!!

PR #132, opened 2023-09-17
Pytest parallelisation using pytest-xdist

d0choa:
25% gain on testing performance, using fairly standard module pytest-xdist + some additional adjustments of the testing spark session.

Not very rigorous performance benchmark:

[main] Results (125.78s (0:02:05))
[xdist] Results (100.57s (0:01:40))


tests/conftest.py
```diff
@@ -47,6 +49,11 @@ def spark() -> SparkSession:
         .config("spark.worker.ui.retainedDrivers", "1")
         # fixed memory
         .config("spark.driver.memory", "2g")
+        .config("spark.sql.warehouse.dir", tmp_path_factory.mktemp("warehouse"))
+        .config(
+            "spark.driver.extraJavaOptions",
+            "-Dderby.system.home={tmp_path_factory.mktemp('derby')}",
```
ireneisdoomed, line 55: My understanding of this is that we are explicitly configuring Dderby as the metadata store for Spark. And by defining specific paths for Spark to write the metadata, we are avoiding clashes between the multiple Spark Sessions created by pytest-xdist.

ireneisdoomed (approved):
Really nice! Tested and worked. ~2min on a 4 cores machine.

For clarification, pytest-xdist works by creating as many testing environments as available nodes we have. Therefore, on 4 nodes, we'll create 4 instances of all fixtures. On an ideal scenario, I guess we'd have a single session parallelising all tests.
I've made some suggestions:
- Since we can specify how we want to run pytest at the project level, I've included all the parameters we use inside the TOML. This way, every time we do `pytest`, either locally or in CI/CD, we will use our custom config.
- Minor improve of `make test` to remove the local XML files we generate.

Thank you also for the tests clean up. I've checked that tests related with the GWAS catalog curation are covered in the current project. For the Coloc and V2G tests, it's sad to let them go but I'll make sure we bring those back when we go back to these components.

PR #133, opened 2023-09-18
fix: missing dataset schemas do to wrong hook setup

d0choa:
Mkdocs built-in hook functionality was not properly setup


ireneisdoomed (approved):
Thank you!

PR #140, opened 2023-09-22
Introducing hierarchy into dataclasses

DSuveges:
Code changes covering #3078

- [x] A new class was introduced to represent and validate genomic regions.
- [x] A new class was introduced to represent single point associations.
- [x] The single point association class has shared methods applicable to study locus and summary statistics dataset.
- [x] Summary stats inherits from single point association
- [ ] Study locus inherits from single point association
- [x] All tests must pass.


DSuveges:
Not relevant now. Closing. Keep the branch though.

PR #143, opened 2023-09-26
Correct formula for PICS standard deviation (duplicated)

ireneisdoomed:
Duplicated from https://github.com/opentargets/genetics_etl_python/pull/113


src/otg/method/pics.py
```diff
@@ -61,12 +61,12 @@ def _pics_standard_deviation(neglog_p: float, r2: float, k: float) -> float | No
...
             0.0
             >>> round(PICS._pics_standard_deviation(neglog_p=10.0, r2=0.5, k=6.4), 3)
-            0.143
+            1.493
             >>> print(PICS._pics_standard_deviation(neglog_p=1.0, r2=0.0, k=6.4))
             None
         """
         return (
-            (1 - abs(r2) ** 0.5**k) ** 0.5 * (neglog_p) ** 0.5 / 2
+            abs(((1 - (abs(r2) ** 0.5) ** k) ** 0.5) * (neglog_p**0.5) / 2)
```
DSuveges, line 69: OK. It might be a trivial question: is `r2` a squared number (the notation suggests)? Because in that case it should always be positive, so the `abs()` is useless. I might be wrong though.

tskir (approved):

src/otg/method/pics.py
ireneisdoomed: You're right, it's useless

PR #144, opened 2023-09-26
Preprocess with pluggable architecture, Airflow-based workflow framework, FinnGen summary stats ingestion

tskir:
This is an omnibus PR which introduces several major features.

### New, pluggable architecture for datasource ingestion
Under the new approach, if we want to ingest a new datasource, we just have to implement one fully self-contained module under `src/otg/preprocess`, and add an appropriate entry to the Airflow Preprocess DAG under `airflow/dags`. That's it! I hope that this will greatly speed up any subsequent work to ingest new datasets (see: UK Biobank, GWAS Catalog, various QTLs, FinnGen finemapping, potential future biobanks...)

### Airflow-based workflow framework for coordination of ingestion tasks
I would like to once again thank @ireneisdoomed and @d0choa for introducing me to Airflow. This is in my experience the first workflow manager which actually works and helps me develop workflows rather than actively trying to prevent me from doing so. Workflow layout currently looks like this (subject to future optimisations):
image

### FinnGen summary stats ingestion
I demonstrate the use of two features above to implement and execute full FinnGen summary stats ingestion. More datasets to follow, based on the same architecture.

----

The architectural solutions in this PR are still not completely final. For example, I can think of a more elegant way to pass parameters to Preprocess modules; DAG configuration code could be made even more elegant and clear; there are some optimisations to be made to ensure the maximum performance of Dataproc clusters.


.vscode/settings.json
```diff
@@ -1,4 +1,11 @@
 {
+  "black-formatter.args": [
+    "--line-length",
+    "9999"
+  ],
+  "flake8.args": [
+    "--max-line-length=9999"
+  ],
```
tskir, line 8: Note: I included this because ruff is configured to ignore long lines [which is actually also something we can discuss], but VS Code isn't, so it kept flagging the long lines as problems. At least that's what was happening on my setup.

tskir:
All changes from this PR haven been broken down into:
1. https://github.com/opentargets/genetics_etl_python/pull/150
2. https://github.com/opentargets/genetics_etl_python/pull/164
3. https://github.com/opentargets/genetics_etl_python/pull/166
4. https://github.com/opentargets/genetics_etl_python/pull/167

PR #146, opened 2023-09-27
feat: create new object in pics results

d0choa:
Create object instead of modifying the existing on PICS results


d0choa:
I'm closing this for now and working on something bigger

PR #148, opened 2023-09-28
feat: enable compatibility with google cloud storage connector

Daniel-Considine:
Update to allow pyspark/hail to read and write to google cloud buckets when working locally on the ETL. For example, running an ETL script with `spark.read.parquet('gs://...')` in it will now correctly read the parquet from a google cloud bucket. 

Make sure to run `gcloud auth application-default login` prior to `make setup-dev`. 

Code adapted from here to conform to ETL formatting: https://github.com/broadinstitute/install-gcs-connector/blob/master/install_gcs_connector.py

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ clean: ## Clean up prior to building
 setup-dev: SHELL:=/bin/bash
 setup-dev: ## Setup development environment
 	@. utils/install_dependencies.sh
-
+	@poetry run python utils/install_gcs_connector.py
 check: ## Lint and format code
 	@echo "Linting..."
 	@poetry run ruff src/otg .
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "otgenetics"
-version = "1.0.0+dc16"
+version = "1.0.0"
 description = "Open targets Genetics Portal Python ETL"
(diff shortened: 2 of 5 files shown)


d0choa:
I'm slightly concerned about this. If someone tries to run some of our steps locally (e.g. reading summary stats), it will incur a potentially massive bill due to egress on huge datasets. If we want to allow for this we are going to have to add some safety mechanisms somewhere else.

Makefile
```diff
@@ -19,7 +19,7 @@ clean: ## Clean up prior to building
 setup-dev: SHELL:=/bin/bash
 setup-dev: ## Setup development environment
 	@. utils/install_dependencies.sh
-
+	@poetry run python utils/install_gcs_connector.py
```
DSuveges, line 22: Hi @Daniel-Considine, I'm qiute concerned that this connector. I feel all our components are expected to either run on dataproc, where the connection to GCP buckets are managed at a deeper level or locally (mostly for debugging or prototyping), where a smaller chunk or sample of the input data is sufficient. 

I feel, having the logic in our repo to magange the gcs_connector.py is quite a burden and a bit out of scope.

Daniel-Considine: No worries - I thought it might be useful for testing smaller steps on the stored data without having to download to local/rewrite paths.

PR #149, opened 2023-10-01
Generalizing population mapping

DSuveges:
Backstory: prior PICS, all association needs to go through LD expansion, which has to be an acestry aware process. Up until this point, this process was only implemented for GWAS Catalog derived associations, which had a huge impact on how the standardization of the population name was implemented.

More details under [#3094](https://github.com/opentargets/issues/issues/3094)

**Major updates**:

- [x] population mapping is now a static method in the study index class.
- [x] It needs a list of structs with the ancestry labels and sample sizes. After mapping to the populations in the LD reference, samples from the same ancestries are merged and the relative sample size is returned.
- [x] As the Genetics team expressed their interest to propagate the results of the mapping and the relative sample sizes, this process is now moved to the generation of the study index.
- [x] Implemented for UKBB, GWAS Catalog and Finngen.
- [x] There's a number of smaller updates reflecting the source agnostic nature of the process in which we no longer bound to GWAS Catalog and GnomAD only.
- [x] New tests were added, old test got updated and now all passes.


src/otg/datasource/gwas_catalog/study_index.py
```diff
@@ -36,26 +31,6 @@ class GWASCatalogStudyIndex(StudyIndex):
 
     """
 
-    @staticmethod
```
DSuveges, line 39: This logic is moved to `StudyIndex`.

src/otg/datasource/gwas_catalog/study_index.py
```diff
@@ -117,53 +92,6 @@ def from_source(
             ._annotate_discovery_sample_sizes()
         )
 
-    def get_gnomad_population_structure(self: GWASCatalogStudyIndex) -> DataFrame:
```
DSuveges, line 120: The `df.transform(fun)` approach got dropped. Now this operation is column based.

d0choa (approved):
I would probably update the `update_study_id` function title. Feel free to adjust here or in another PR

PR #154, opened 2023-10-03
chore: unused dependency

d0choa:


ireneisdoomed (approved):
Good! Bugbear is a plugin with native support from Ruff ([issue](https://github.com/astral-sh/ruff/issues/389))

PR #155, opened 2023-10-03
fix: cleaning clumping

DSuveges:
There were a few minor details in the locus collecting branch of the clumping logic that introduces some problems. Needed to fix


d0choa (approved):
good

PR #157, opened 2023-10-03
Some studies just don't have population data causing LD annotation fail

DSuveges:
I'm not 100% sure this is the right thing to do... The LD set certainly can be resolved, but the subsequent r2Overall calculation fails. Adding the condition nullifies the LD set, but then we need to add an other QC metric to flag these special cases.

```python
(
    study_index.df
    .filter(f.col('ldPopulationStructure').isNull())
    .select('studyId', 'ldPopulationStructure')
    .show(20)
)
```

Out of 87,996 studies 12:
```
+------------+---------------------+
|     studyId|ldPopulationStructure|
+------------+---------------------+
|  GCST000218|                 null|
|GCST90093226|                 null|
|GCST90093235|                 null|
|  GCST000870|                 null|
|GCST90093223|                 null|
|GCST90102942|                 null|
|GCST90093229|                 null|
|GCST90093238|                 null|
|GCST90093232|                 null|
|GCST90093220|                 null|
|GCST90093217|                 null|
|  GCST000768|                 null|
+------------+---------------------+
```

### DIFF
--- a/src/otg/dataset/study_locus.py
+++ b/src/otg/dataset/study_locus.py
@@ -47,6 +47,7 @@ class StudyLocusQualityCheck(Enum):
     AMBIGUOUS_STUDY = "Association with ambiguous study"
     UNRESOLVED_LD = "Variant not found in LD reference"
     LD_CLUMPED = "Explained by a more significant variant in high LD (clumped)"
+    NO_POPULATION = "Study does not have population annotation to resolve LD"
 
 
 class CredibleInterval(Enum):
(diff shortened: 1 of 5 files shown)


DSuveges:
The process is failing, due to not all discovery-samples are parsed properly:

```
 discoverySamples      | [{African American or Afro-Caribbean, African unspecified, European, East Asian, Hispanic or Latin American, South Asian, 577663}]                                                                                                  
 ldPopulationStructure | [{null, null}] 
```

d0choa (approved):

PR #159, opened 2023-10-04
Enhancements around PICS and credible set annotation

d0choa:


d0choa:
@DSuveges ready for review

DSuveges:
`._qc_no_population()` - on StudyLocus never works: that function checks the presence of a column, it filters on, if that column is not in the dataframe the dataframe returned unchanged. However that column is not part of the StudyLocus schema. Realizing this made me think about generalizing the QC process. [#3101](https://github.com/opentargets/issues/issues/3101)

tests/dataset/test_study_locus.py
```diff
@@ -199,8 +213,13 @@ def test_annotate_credible_sets(
     )
     data_sl = StudyLocus(
         _df=spark.createDataFrame(observed, schema), _schema=StudyLocus.get_schema()
-    ).annotate_credible_sets()
```
DSuveges, line 202: How did it work before? It seems the same method is called on data_sl object twice...

DSuveges (approved):

PR #167, opened 2023-10-10
[Preprocess #4] Implement FinnGen Airflow DAG

tskir:
Closes https://github.com/opentargets/issues/issues/3106.

> **Note**
> This is **ready for review,** but please **do not merge.** The base branch is not main, but the previous PR in the series, [Preprocess #3](https://github.com/opentargets/genetics_etl_python/pull/166). This is done so that only changes which are related to this specific PR are displayed. I'll handle merging myself once this is approved.

---

## General considerations

While nested, highly flexible configuration facilities can be very beneficial for other parts of the pipeline, they introduce _a lot_ of friction for data ingestion and preprocessing tasks. This is because all those tasks have something in common:
* They are independent from each other and from the rest of the pipeline.
* They have very rigid configuration. Really, the only parameters we can choose are the output paths. Everything else, including input paths and processing parameters, is determined by the source data.
* We can expect them to be relatively numerous. We already have or are planning to have FinnGen, UK Biobank, GWAS Catalog, and eQTL catalog. This will certainly grow over time, added either by us or by other users of this code base.

## What happens under the current architecture

Under the current architecture, when adding each datasource, we end up writing a lot of boilerplate code; and once we define the configuration variables, we then keep passing them around:

### `config/datasets/gcp.yaml`
image

### `config/step/my_finngen.yaml`
image

### `src/otg/config.py`
image

### Register config
image

### Add entry to the DAG definition
image

The result is that every tiny action, such as adding or renaming a parameter, cascades into lots of changes all over the code. This is very distracting, creates friction, and slows down development.

## The new approach

The _central idea_ of Airflow is “configuration as code” [[1]](https://www.upsolver.com/blog/apache-airflow-when-to-use-it-when-to-avoid-it-while-building-a-data-lake), [[2]](https://big-data-test-infrastructure.ec.europa.eu/service-offering/apache-airflow_en). When applied properly, this approach can bring a lot of benefits. It is especially useful for orchestrating those ingestion tasks.

Under the new approach, you define everything directly in the DAG. That's it:
image

## Note on flexibility
Aside from being succinct and readable, the new architecture is extremely flexible and straightforward in terms of orchestration choices. It is possible to:
1. Manually run the workflow or any specific step in Airflow UI.
2. Run the workflow or any specific step from Airflow CLI.
3. Configure repeated or automatically triggered execution.
4. Run the individual ingestion script locally: `python3 src/otg/preprocess/finngen.py --finngen_release_prefix=FINNGEN_R9 [...]`.
5. Manually submit the same script with parameters as a PySpark job.
6. Import ingestion logic from another Python module: `from src.otg.preprocess.finngen import main; main ([...])`

All of these options are straightforward and don't require any additional boilerplate code; they just work.


tskir:
Most of the changes from this PR have already been implemented in https://github.com/opentargets/genetics_etl_python/pull/199 and https://github.com/opentargets/genetics_etl_python/pull/200. It's also way out of alignment with the recent changes, so I'll submit a new PR for this.

PR #169, opened 2023-10-13
Ds 3117 refactor clumping

DSuveges:
A trivial update in making window based clumping more intuitive.


d0choa (approved):
makes more sense

PR #170, opened 2023-10-13
docs: several fixes in docstrings

d0choa:
Trivial fixes. I thought we were picking up this kind of errors somewhere in the precommits but it doesn't seem to be the case.


src/otg/datasource/intervals/thurnman.py
```diff
@@ -19,7 +19,7 @@ class IntervalsThurnman(Intervals):
     """Interval dataset from Thurman et al. 2012."""
 
     @staticmethod
-    def read_thurnman(spark: SparkSession, path: str):
```
DSuveges, line 22: Is the person `Thurman` or `Thurnman`? Not a crutial detail though. I assume it's the former as the publication behind the dataset is most likely [this](https://www.nature.com/articles/nature11232). Also [here](https://genetics-docs.opentargets.org/our-approach/data-sources/others). It hink it would be great if we could take a look around the repo and fix everywhere.

d0choa:
Thurman instead of Thurnman changes implemented. Tests passing

DSuveges (approved):
Looks good. Thanks

PR #171, opened 2023-10-13
Fine-mapping

hlnicholls:
This PR introduces an outline of the code needed for fine-mapping. I have updated it to be divided into each method:
1.	Fine-mapping data extraction
    - Extracts LD matrix and summary statistics for +/- 500kb region
    - Filters both datasets for only matching/intersecting SNPs
2.	DENTIST outlier detection
    - Runs outlier detection using a [DENTIST calculation](https://github.com/mkanai/slalom/blob/1339bd4e8fcd0c0315dd64e19253a7fa4ec4b74b/slalom.py#L245C12-L245C12) provided by the slalom package (https://github.com/mkanai/slalom/)
2.	SuSiE fine-mapping
    - Runs SuSiE-inf fine-mapping-inf python package  (https://github.com/FinucaneLab/fine-mapping-inf/)

**Fine-mapping Data Extraction**

The fine-mapping data extraction uses a `fm_data_extraction.py` script in `/otg/datasets` which contains four functions:

1. `get_FM_study_locus()` – gets summary statistics and the lead SNP ID for a locus.
2. `get_gnomad_ld_matrix() `– uses the lead SNP ID to get the LD matrix within +/- 500kb
3. `get_matching_snps()` – Finds intersect of SNPs in summary statistics and LD matrix
4. `allele_flip_check()` - Checks alleles match between LD matrix and summary statistics (added in case we have conflicting input data in the future)

I have each of these functions running individually with some UKBB summary statistics taken from the GCP [here](https://github.com/hlnicholls/slalom-susie/blob/main/python-pipeline-tests/ld_matrix_extraction_GCP_test.ipynb) (the functions work, but I expect they need re-writing as I had conflicting pyspark and hail version errors throughout creating these functions both locally and on a dataproc cluster, for some reason only specific pyspark functions would give a version error). 

Also, when submitting a job for this step in the ETL I’ve been getting an error: `Error locating target 'otg.fm_data_extraction.FMDataExtractionStep'` - so I haven’t been able to test it further yet. I spent a couple of hours trying to address this, as I assume something small is missing in my step setup, but haven’t found the issue.

**DENTIST**

Dentist outlier detection has 2 functions:

1. `calculate_dentist()` – detects which SNPs are outliers and creates a binary column denoting which is or isn’t an outlier. This function takes the filtered summary statistics from the fine-mapping data extraction as input. It also has the option to switch to using CARMA outlier detection, this is unfinished and maybe not worth actually including, but in theory just the outlier detection of CARMA, aside from its fine-mapping, could be applied here.
2. `calc_nlog10p_dentist_s()` – this function is nested inside `calculate_dentist()` provides the final calculation for dentist before it is used for outlier detection.

I haven't tested DENTIST in the ETL but I have a test run of my function [here](https://github.com/hlnicholls/slalom-susie/blob/main/python-pipeline-tests/pyspark_dentist_func.ipynb). This function needs the summary statistics with an R2 column measuring LD for all variants with the lead SNP of the locus.

Also just to note, the DENTIST thresholds I used are the default in the SLALOM package `(df.R2 > r2_threshold) & (df.nlog10p_dentist_s > nlog10p_dentist_s_threshold)` however this rarely detects outliers and suggests we should test other thresholds. For example, the SLALOM paper also tests using just `df.R2 > df.t_dentist_s` - which in some examples I've run does successfully detect outliers (although usually still a low number).

**SuSiE**

- Currently `/otg/method/susie.py` is just a placeholder of code copied from the finemapping-inf package. This still needs testing and conversion into pyspark if it is the method we choose to use.
- I have also ran these SuSiE functions that I copied into the ETL on an example locus [here](https://github.com/hlnicholls/slalom-susie/blob/main/python-pipeline-tests/susie-inf-test/susie-inf-test.ipynb).

The code for running dentist and susie are completely untested in the ETL. However, I have run the functions individually (https://github.com/hlnicholls/slalom-susie/). I also have some expected output results of SuSiE [here](https://github.com/hlnicholls/slalom-susie/tree/main/example-results) for reference


src/otg/method/finemapping.py
```diff
@@ -0,0 +1,160 @@
...
+        susieinf_command = f"""python /Users/hn9/Documents/GitHub/fine-mapping-inf/run_fine_mapping.py \
+            --sumstats /Users/hn9/Documents/GitHub/fine-mapping-inf/susieinf/loci/{lead_SNP_ID}_locus_sumstat_with_dentist.txt.gz \
+            --beta-col-name beta \
+            --se-col-name se \
+            --ld-file /Users/hn9/Documents/GitHub/fine-mapping-inf/susieinf/loci/{lead_SNP_ID}_locus_ukbb_ld.txt.gz \
+            --n {n_sample} \
+            --method susieinf \
+            --save-tsv \
+            --eigen-decomp-prefix {lead_SNP_ID}_locus \
+            --output-prefix  {lead_SNP_ID}_locus """
```
DSuveges, line 157: I have only skimmed through the code, but having hard coded local path would certainly not work. It needs to be somehow parametrized.

hlnicholls: Yes I agree. Also I suspect, if we are happy to use this susie-inf package, that we can take their functions directly instead - but I’m not sure how this is best implemented in the ETL so I’ve left this command line code as a placeholder for now.

hlnicholls: Small update on this: I've updated the code so it's setup to try and run the susie functions copy-pasted from the finemapping-inf package directly (tested it [here](https://github.com/hlnicholls/slalom-susie/blob/main/python-pipeline-tests/susie-inf-test/susie-inf-test.ipynb) but not tested running in the ETL itself). I think this will still need a lot of work, but hopefully it's a better starting point

tskir:
@DSuveges Given that Hannah has left, do we plan on finalising this PR? Is this currently assigned to anyone? And as you reviewed it previously, do you think it'll require only cosmetic changes or a big overhaul?

DSuveges:
As this work is now contiuned on different branches, This PR is closed.

PR #172, opened 2023-10-16
Documentation structural changes and data source images

d0choa:
Several changes to documentation:

- Navigation bar to simplify navigation on different sections
- Top-level sections consolidated
- Logos added to data sources
- Dataset classes harmonised with schema first, class last (Irene's style)
- All images moved to assets/imgs/
- Contributing guidelines + Troubleshooting moved to the same top-level category

This PR does not include additional documentation, just structural changes. Once, this is agreed we can start adding more content.


DSuveges (approved):
All changes makes sense.

PR #173, opened 2023-10-16
bumping dependency versions

d0choa:
poetry was a bit lazy updating some libraries and causing issues due to older versions of dependencies being used.

This is a bump on all possible libraries to use the latest compatible version.

Tests are all passing, but if any issues emerge we should come back to what's changed here.


ireneisdoomed (approved):
I tested locally that tests ran and that the CLI set up with hydra was working.
However we should keep an eye open for breaking changes when running steps.

PR #174, opened 2023-10-16
automatically PR to bump poetry packages

d0choa:
This action should open PRs to update poetry packages, which is not happening at the moment


ireneisdoomed (approved):
🙏

PR #177, opened 2023-10-16
Github dependabot

d0choa:
Activates dependabot for poetry/pip and github actions

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,22 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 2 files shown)


ireneisdoomed (approved):
This should inform us about updates in the dependencies of our project.
- A concern was that we introduce breaking changes or incompatibilities between packages. I think we are quite safe because tests won't run on Github unless the new environment is properly set up. In https://github.com/opentargets/genetics_etl_python/pull/177/commits/1920c9d2dc314578814b906d002b583b7b649186 , I added an extra test that checks that `pyproject.toml` is valid and that there are no incompatibilities with the lock file.
- https://github.com/opentargets/genetics_etl_python/pull/177/commits/24be6de4c9d5263841b02068f65e3a38f5e96484 adds configuration for the commit message. If `package` goes from 1.2 to 1.5, the commit message will be `build(package): update from 1.2 to 1.5`

PR #189, opened 2023-10-19
Add missing columns in gene index

ireneisdoomed:
In the process of running V2G I ran into minor issues with the gene index.
This PR contains:
- Fix typo in the gene index output path
- Addition of missing columns of the target dataset that are used later in the project:
    -  approvedSymbol
    - approvedName
    - biotype - used in V2G
    - obsoleteSymbols
- with `obsoleteSymbols`, I propose a simplification of the schema. Impact is small. As a counterargument, we might not want to deviate from the target's original schema. Let me know your comments
    - In the target we have an array of structs with `label` and `source`
    - In my changes I propose to extract only the labels, so that we simply operate with an array of strings  

Output is herer `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/gene_index`

### DIFF
--- a/config/datasets/gcp.yaml
+++ b/config/datasets/gcp.yaml
@@ -27,7 +27,7 @@ ld_index_raw_template: gs://gcp-public-data--gnomad/release/2.1.1/ld/gnomad.geno
 ld_matrix_template: gs://gcp-public-data--gnomad/release/2.1.1/ld/gnomad.genomes.r2.1.1.{POP}.common.adj.ld.bm
 
 # Output datasets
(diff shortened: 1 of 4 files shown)


DSuveges:
Do we need `obsoletedSymbols` to normalize gene names? Where do we need to do this? Intervals?

ireneisdoomed:
@DSuveges Exactly, it's used in the intervals to build an ID/name LUT

DSuveges (approved):
All makes perfect sense. Thank you!

PR #190, opened 2023-10-20
feat: run variant to gene step

ireneisdoomed:
This PR includes minor fixes to generate a variant to gene dataset.
- Removal of the column `position` from V2G. We used this when I was trying to repartition the data by chromosome, position. Not needed anymore.
- Removal of the column `label` from V2G. We were showing different information in this very general field. For the VEP consequence, label contained the label of the SO term; for the polyphen/sift assignments, it contained an interpretation of the score. I don't think we need that.
- More complex logic in the LiftOver class to read the chain file. The python package cannot read remote files, so we need logic to download the data if a GCS location is provided. Chain file is just 500kb.
- + minor fixes and enhancements for the code to run

**Important notes**
- Variant to gene assignments from all sources: 18,533,638 (9,312,515 unique variant/gene pairs). This is quite small in comparison with what I was extracting last year. That is due to the size of the current variant index, which is rather a subset since it only contains Finngen and top hits study locus.
    - Data: `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/v2g` 
- [Job](https://console.cloud.google.com/dataproc/jobs/my_v2g-tkytt4xyxkhve?region=europe-west1&project=open-targets-genetics-dev) runs in ~13min


DSuveges:
The size of the variant index is can realistically expected to be ~1-1.5 orders of magnitude bigger as we are expecting data from ~100k studies, however Finngen is just 2k.

src/otg/common/Liftover.py
```diff
@@ -38,8 +40,17 @@ def __init__(
...
+        if chain_file.startswith("gs://"):
+            bucket_name = chain_file.split("/")[2]
+            blob_name = "/".join(chain_file.split("/")[3:])
+            with tempfile.NamedTemporaryFile(delete=True) as temp_file:
+                storage.Client().bucket(bucket_name).blob(
+                    blob_name
+                ).download_to_filename(temp_file.name)
+                self.lo = LiftOver(temp_file.name)
+        else:
+            self.lo = LiftOver(chain_file)
```
DSuveges, line 54: I'm wondering if it worth abstracting this functionality into the utils methods. OK, at this point this is the only application that cannot handle google storage, however I see this might be an issue in the future.

Also not 100% sure how this solution would be portable for aws and azure storages. (Just thinking about @addramir's promise to make the pipeline portable.) It's not a blocker in any ways, just thinking about design/portability.

src/otg/v2g.py
```diff
@@ -36,16 +38,24 @@ def run(self: V2GStep) -> None:
...
+                "variantFunctionalConsequenceId"
+            ),
+            f.col("Term").alias("label"),
+            f.col("v2g_score").cast("double").alias("score"),
         )
 
         # Variant annotation reduced to the variant index to define V2G variant universe
-        va_slimmed = va.filter_by_variant_df(vi.df, ["id", "chromosome"]).persist()
+        va_slimmed = va.filter_by_variant_df(
+            vi.df, ["variantId", "chromosome"]
```
DSuveges, line 57: Why do you need to specify the columns you join the two dataframes? Isn't it always the same? What other usecases are there?

src/otg/common/Liftover.py
ireneisdoomed: I also thought about this and agree with you. This code would crash if the path is pointing to a different cloud provider. 
I decided not to do anything about it because the former solution of using gcsfs would lead to the same results, and because in general we're quite locked in Google's cloud platform. Porting this to other cloud providers will have more ramifications. So far, this solution works with local files and with GCS files.

src/otg/v2g.py
ireneisdoomed: I don't think we use this function in any other way. It is more flexible to select the joining cols, but hard coding them to variantid and chromosome shouldn't cause problems. Thanks

ireneisdoomed:
> The size of the variant index is can realistically expected to be ~1-1.5 orders of magnitude bigger as we are expecting data from ~100k studies, however Finngen is just 2k.

Previously, the index contained 6 million variants when processing the old credible sets. However, with the more refined StudyLocus, I expect the final variant count to be much less.

ireneisdoomed:
@DSuveges I forgot to ask you something about this comment in the Javierre and Andersson logic https://github.com/opentargets/genetics_etl_python/blob/c021e6b538bed320bfba5a3b07d4fe5b3e5fc4f6/src/otg/datasource/intervals/andersson.py#L62
We have this hardcoded value to apply a threshold between a variant and the distance to a gene's TSS. What is it that needs to be done applying percentiles? Is this important?

DSuveges:
> oded value to apply a threshold between a variant and the distance to a gene's TSS. What is it that needs to be done applying percentiles? Is this important?

I have to admit, I can't remember what I was thinking <s>exactly</s> at all. So, it is certainly not important.

src/otg/v2g.py
```diff
@@ -62,22 +70,24 @@ def run(self: V2GStep) -> None:
             va_slimmed.get_plof_v2g(gene_index_filtered),
             # intervals
             IntervalsAndersson.parse(
-                IntervalsAndersson.read_andersson(self.session, self.anderson_path),
+                IntervalsAndersson.read_andersson(
```
DSuveges, line 73: I might be just a huge pain, but I feel, the list of the included intervals should not be hardcoded. I think we should have datasources to process interval datasets with consistent schema ready to be ingested by v2g, however, v2g should not be aware of what are the explicit list of the available datasets. This implementation makes it more complicated to introduce more intervals. What do you think?

src/otg/v2g.py
DSuveges: Thanks.

ireneisdoomed:
@DSuveges Thank you for your suggestions Daniel. I think it now looks much better. The list of intervals to include is now pulled from the configuration, the actual V2G logic is source agnostic. The job runs successfully.

DSuveges (approved):
Thanks for considering my recommendations.

PR #192, opened 2023-10-20
fix(docs): schemas path for gene and study index

ireneisdoomed:
Schemas for Gene and Study index were pointing to an obsolete filename.


DSuveges (approved):
Wow, good catch. I'm wondering if there could be some sort of tests to prevent such divergence in the future.

PR #199, opened 2023-10-26
Refactor step configuration

tskir:
Closes https://github.com/opentargets/issues/issues/3132.

For each of the existing steps:
* Moved the list of attributes and their descriptions into the main `*Step` class;
* Adjusted `my_STEP.yaml` files so that they populate the appropriate class directly via `_target_`;
* Removed the inheritance from a `*StepConfig` class;
* Since it is no longer required, removed the `*StepConfig` class and the accompanying `cs.store` call.

As can be seen from the diff (± lines), this reduces the code required to define the same configuration by about 35%, and removes two entities: `*StepConfig` classes and `cs.store` calls.

At the same time, we retain the full flexibility, as everything can still be configured through YAML files, overridden from the command line argument or in `omegaconf`'s `instantiate` call; and the `*Step` classes can still be inherited and modified (although we don't use this at the moment).

### DIFF
--- a/config/step/my_colocalisation.yaml
+++ b/config/step/my_colocalisation.yaml
@@ -1,2 +1,3 @@
+_target_: otg.colocalisation.ColocalisationStep
 study_locus_path: ${datasets.study_locus}
 coloc_path: ${datasets.colocalisation}
--- a/config/step/my_colocalisation.yaml
+++ b/config/step/my_colocalisation.yaml
@@ -1,6 +1,2 @@
-# Default config
-defaults:
-  - colocalisation
-# Additional config
 study_locus_path: ${datasets.study_locus}
 coloc_path: ${datasets.colocalisation}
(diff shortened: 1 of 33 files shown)


d0choa (approved):
Makes sense. I like that things are more self-contained within the steps. Also, documentation now provides information about what inputs are required to run the step.

I'm approving without running it. I would suggest @ireneisdoomed have a look as well because she has at least one clashing branch. Let's try to merge this as soon as she approves so we don't create more conflicts.

As a follow-up, perhaps we could remove the `run` function from each step and `step.run()` from the `cli.py`. There is not much more we would like to do with a step different than running it. Every step will run as part of the initialisation of the object and it will reduce the boilerplate even further.

ireneisdoomed (approved):
Thanks! I personally prefer having all configuration together, but this point from David kind of sold it 
> documentation now provides information about what inputs are required to run the step.

I've run the gene index step to validate the changes.

PR #200, opened 2023-10-26
Follow up changes to step configuration refactor

tskir:
> **Note**
> * **Ready for review,** but please do not merge yet: base branch is not `main`.
> * Based on top of [#199](https://github.com/opentargets/genetics_etl_python/pull/199).

Closes https://github.com/opentargets/issues/issues/3132.

This PR contains follow up changes and is submitted separately for easier review (as the previous one was quite long already):
* For each step, changed `run()` to `__post_init__()` and removed `step.run()` from cli.py as [suggested](https://github.com/opentargets/genetics_etl_python/pull/199#pullrequestreview-1699310247) by @d0choa
* Renamed all steps without the prefix, for example: `my_finngen` → `finngen`. Now that we don't register the configs in the config store, there isn't a conflict which warrants this name scheme. Updated YAMLs and DAGs accordingly.
* Made step docstrings uniform: all information is now in the class docstring, while the `__post_init__` docstring just says "Run step." (I would have removed it entirely, but pre-commit isn't happy about this idea...)


ireneisdoomed (approved):
Thank you!
I've added 2 changes:
- Python handler that excludes private methods to be rendered by mkdocs (we were including `__post_inits__`
- Renamed `my_config` in favour of `config`. Hydra picks `config` as the config name by default

tskir:
Thank you @ireneisdoomed! I like how the code is even cleaner and elegant now :)

PR #201, opened 2023-10-27
Streamline Airflow in Docker for GCP configuration

d0choa:
This is an attempt to minimise the number of steps required by a user to set up the local Airflow instance. Changes include:

- Versioned docker compose YAML and Dockerfile with all the required GCP adjustments (airflow provider, google credentials, etc.). This removes several steps of the previous configuration (e.g. adding variables in the Airflow UI or the Python DAG). All variables are parametrised through the `.env` file as suggested by the official YAML.
- Use Airflow `LocalExecutor` instead of `CeleryExecutor` for a more lightweight airflow setup. This also removes the need for some of the services (e.g. reddis) [[1](https://docs.astronomer.io/learn/airflow-executors-explained)] [[2](https://datatalks.club/blog/how-to-setup-lightweight-local-version-for-airflow.html#:~:text=3.-,Why%20this%20setup%20is%20lighter,-The%20main%20reason)]
- Revised documentation with the streamlined procedure
- The `src/airflow/` directory now includes `logs`, `plugins`, `config` and `dags`, so directories don't need to be created.
- Removal of legacy `/workflow` directory

This PR does not cover any logic in the DAG apart from removing the GCP credentials handling which is no longer required.


ireneisdoomed (commented):
I've managed to set up the Airflow server and send jobs!
Instructions are clear, however I've run into a few issues:
- Python version of the image is incompatible with Google Cloud SDK. When I build the extended image, it was raising into `AttributeError: module 'collections' has no attribute 'Mapping'`.
    - It worked when i downgraded it to 3.9. My suggestion is to use the slim Airflow version [apache/airflow:slim-latest-python3.9](https://hub.docker.com/layers/apache/airflow/slim-latest-python3.9/images/sha256-d920504e8f2b24393dc47e9f22e304469724ced184d2dbe15746c1b2ad4a6627?context=explore)
- The initialisation of the database was failing with the new image. The slim version doesn't have `psycopg2`, a dependency of postgresql.
    - Solved it just by adding it to `requirements.txt`. No more problems in the set up.
- With Airflow set up, I started a new DAG run but all steps were failing ([example](https://console.cloud.google.com/dataproc/jobs/1c6815c4-0cfd-4607-bbc3-0f70782ab85c/monitoring?region=europe-west1&project=open-targets-genetics-dev)), even the gene index which is what i wanted to test. I think this has to do with the latest Hydra changes in main.
    - Error `Cannot find primary config 'config'. Check that it's in your config search path.` 

Something that I'm not sure why it doesn't work now is the assignation of Pyspark job names based on the task ID. There is a `job_name` argument that is [meant to be based ](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_api/airflow/providers/google/cloud/operators/dataproc/index.html#airflow.providers.google.cloud.operators.dataproc.DataprocJobBaseOperator)on the `task_id`, but right now IDs are incomprehensible.

src/airflow/docker-compose.yaml
```diff
@@ -0,0 +1,232 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
tskir, line 1: Are we absolutely certain that we want to include the `docker-compose.yaml` as part of our code base? My issue with this is that this is an externally maintained file, with only a few custom changes introduced by us.

In the future, as Airflow continued to evolve, this config file might become obsolete in subtle ways, and in the worst case scenario cause all sorts of weird errors.

I'm still in favour of the approach where we fetch the fresh YAML file from Airflow, and then only introduce the changes we want. (This could even be automated using the `patch` command, so that the user doesn't have to do the changes manually.) In this case we can always be sure which changes are ours, and updates won't be a pain.

tskir (commented):
Thank you David! Great to see our Airflow implementation mature :upside_down_face:  I've added a few minor style suggestions and one conceptual discussion point.

src/airflow/.env
```diff
@@ -0,0 +1,8 @@
+AIRFLOW_UID=1126896676
+AIRFLOW_IMAGE_NAME=extending_airflow:latest
+GOOGLE_LOCAL_CREDENTIALS_PATH=~/.config/gcloud
+GOOGLE_DOCKER_CREDENTIALS_PATH=/.config/gcloud
+GOOGLE_APPLICATION_CREDENTIALS=/.config/gcloud/service_account_credentials.json
+AIRFLOW_CONN_GOOGLE_CLOUD_DEFAULT='google-cloud-platform://?extra__google_cloud_platform__key_path=/.config/gcloud/service_account_credentials.json'
+GCP_PROJECT_ID=open-targets-genetics-dev
+#GCP_GCS_BUCKET
```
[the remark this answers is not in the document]
d0choa: It was a reminder that it can be set up in the same way so it's available to every DAG. I'll remove it and we can keep this in mind for further parametrisation.

src/airflow/docker-compose.yaml
d0choa: I put quite a bit of thinking about this before implementing it and I checked online what the good practices are. My conclusions:

The main rationale is that you don't want every developer/user of the pipeline to repeat the same steps. In terms of overall benefit, we want to have as few steps as possible to spawn an airflow service.

Apparently, it's good practice to version the images/composer files, similar to why we version the `poetry.lock`. It keeps track of the version of the code/services that works and it can be tested. We could probably test more aspects of the airflow codebase but there are some tests already to ensure minimal functionality

My impression is that the alternative solution also has its own risks. Imagine a user "curling" the latest YAML and realising things don't run because the system has been changed in the latest YAML. Imagine this user under some pressure to get a run of the pipeline done. Also, we have introduced quite a few changes to the composer YAML (as anybody else is doing) which makes me think that the template YAML is just a very raw recipe that builds a very heavy general-purpose instance.

The cost is that we need to maintain the setup. One of the things, we added on top of the tests to ensure we keep things afloat is the dependabot rule that monitors the Dockerfile for version bumps. I don't think this covers everything, but it's a start and it will ensure PRs are opened and tested.

tskir: Thank you for the explanation David. The points you make do make sense.

I wonder if we can have the best of two worlds by having the following set up:

* Store all custom modifications we make in a patch file, `docker-compose.yaml.patch`. This is the only file we directly make changes to.
* By curling a _specific_ version of the base YAML file and applying the patch, we obtainn the final file `docker-compose.yaml`. **This patched file is also stored in the repository,** and is the one used in all instructions.

So a regular user won't have to curl and patch anything, and the configuration is there; however, for us as developers, we can always keep track of all changes. And if a new version comes out, we can quickly update instead of performing a manual three way merge (old version — our modified version — new version)

tskir: FYI **I don't think this is blocking** to merge this PR — in fact I'll go ahead and submit my approval. I'm just asking in general if you think this system makes sense, and if you do, I can implement it myself as a follow up PR

tskir (approved):

src/airflow/docker-compose.yaml
d0choa: it looks like a nice-to-have. It would improve the system. It's hard to know whether is needed without knowing how often there are breaking changes in Airflow composer yaml

ireneisdoomed (approved):
Re the Config error in the gene index step reported yesterday: this is just a result of an incompatibility between branches.
    - `my_config` was renamed to `config` in main, but the changes in`config.py` (changing `config` to `default_config`) were only made in the L2G branch...
ANYWAY, the proposed changes in this branch have nothing to do with this, and after sorting it out I was able to run the step.
I'll approve, you can also merge or close https://github.com/opentargets/genetics_etl_python/pull/214 before merging this one to main.

PR #202, opened 2023-10-28
Dependabot to monitor airflow docker

d0choa:
It should ensure the `src/airflow/Dockerfile` gets updates, but it hasn't been tested.


ireneisdoomed (approved):
LGTM

PR #212, opened 2023-10-31
feat: use `pydoclint` as a documentation linter

ireneisdoomed:
Checks between a function's arguments and types and its docstrings were lost in the transition to Ruff.
It is in Ruff's plans to bring this functionality ([issue](https://github.com/charliermarsh/ruff/issues/458)), but they haven't started yet.
For the time being I propose using `pydoclint`, a high performance alternative to darglint (the tool that flake8 relied on).

This PR contains:
- addition and configuration of `pydoclint` to the project
    - used their [cli options](https://github.com/jsh9/pydoclint/blob/main/pydoclint/main.py) as a reference
- addition of docstring linting to the `make check` rule
- addition to `pydoclint` as one of the pre commit hooks
    - used https://jsh9.github.io/pydoclint/#23-as-a-pre-commit-hook as a reference 
- several fixes of all issues raised in pydoclint in our project
- a change in `tests/airflow/test_dag.py` to comment out a failing test `test_no_import_errors`
    - this is raised due to an environment discrepancy between the test (which reads the module relative to the rest of the project) and airflow (which resolves the DAG independently of the project)


d0choa (approved):
Great

PR #214, opened 2023-10-31
feat(airflow): change docker image to `apache/airflow:slim-latest-python3.10` and minor changes

ireneisdoomed:
I've made a PR with the suggestions pointed out in https://github.com/opentargets/genetics_etl_python/pull/201
I am not sure they make sense so I prefer not to commit directly.
Summary:
- use the slim Airflow version [apache/airflow:slim-latest-python3.9](https://hub.docker.com/layers/apache/airflow/slim-latest-python3.9/images/sha256-d920504e8f2b24393dc47e9f22e304469724ced184d2dbe15746c1b2ad4a6627?context=explore) with a python version compatible with GCP
- Minor changes in the Dockerfile
- Addition of psycopg2 as a requirement on top of the slim image
- Versioning of the packages inside requirements

I've successfully launched the Airflow server, jobs were failing but that is independent.


d0choa:
I might be missing something, but why are you not using `apache/airflow:slim-latest-python3.10`

src/airflow/Dockerfile
```diff
@@ -1,11 +1,10 @@
...
+FROM apache/airflow:slim-2.7.2-python3.9
 
 # Install additional Python requirements.
 # --no-cache-dir is a good practice when installing packages using pip, because it helps to keep the image lightweight.
 COPY requirements.txt /requirements.txt
-RUN pip install --user --upgrade pip
-RUN pip install --no-cache-dir --user -r /requirements.txt # --no-cache-dir good practise when installing packages using pip. It helps to keep the image lightweight
-
+RUN pip install --user --upgrade pip && \
+    pip install --no-cache-dir --user -r /requirements.txt
```
tskir, line 7:
```suggestion
RUN pip install --quiet --user --no-cache-dir --upgrade pip setuptools && \
    pip install --quiet --user --no-cache-dir -r /requirements.txt
```

* Using `--quiet --user --no-cache-dir` for both commands
* When upgrading `pip`, it's usually a good idea to upgrade `setuptools` as well. Together, their update usually solves all installation problems

tskir (commented):
Looks good to me, added one minor suggestion

ireneisdoomed:
Re the Config error in the gene index step reported yesterday: this is just a result of an incompatibility between branches.
    - `my_config` was renamed to `config` in main, but the changes in`config.py` (changing `config` to `default_config`) were only made in the L2G branch...
ANYWAY, the proposed changes in this branch have nothing to do with this, and after sorting it out I was able to run the step.
I'll approve, you can also merge or close https://github.com/opentargets/genetics_etl_python/pull/214 before merging this one to main.

> I might be missing something, but why are you not using `apache/airflow:slim-latest-python3.10`?

It was because the GCP SDK version was not compatible with Python 3.10. It was hardcoded to v322, updating it to the latest (422) didn't raise any errors.

src/airflow/dags/common_airflow.py
ireneisdoomed, line 1: @tskir @d0choa This sets the cluster to be automatically deleted after 5 minutes. I don't think it is a good idea:
- If we intend to use Airflow for development, 5 mins is very little time
- It creates a sync problem between the autoelimination of the cluster and the DAG (because the deletion clask is not complete)
It is not this PRs intention to touch the Airflow logic, I hope it is ok because it is a small change.

PR #215, opened 2023-11-01
chore: move L2GFeature to datasets

d0choa:


ireneisdoomed (approved):
Thank you! Changes make sense. Ideally I want L2GFeature to be inside the `l2g_feature_matrix` module. At the moment this is not possible because there is a problem of circular dependencies between these 2 (see https://github.com/opentargets/issues/issues/3146)

Tested that the code runs.

PR #216, opened 2023-11-02
Implement Preprocess DAG following recent architecture updates

tskir:
Closes https://github.com/opentargets/issues/issues/3106.

* Added three new functions to the `common_airflow` module: `submit_step`, `read_yaml_config`, and `generate_dag`.
* Updated the ETL DAG to use these functions, making it even more readable and focused on logic.
* Implemented a new separate DAG for the Preprocess part, as we discussed.
* Made some refactoring changes specific to FinnGen ingestion.

### DIFF
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -24,16 +24,21 @@ zone = "europe-west1-d"
 image_version = "2.1"
 
 
-# Executable configuration.
+# Cluster init configuration.
 initialisation_base_path = (
     f"gs://genetics_etl_python_playground/initialisation/{otg_version}"
 )
-python_cli = f"{initialisation_base_path}/cli.py"
 config_tar = f"{initialisation_base_path}/config.tar.gz"
 package_wheel = f"{initialisation_base_path}/otgenetics-{otg_version}-py3-none-any.whl"
 initialisation_executable_file = [
     f"{initialisation_base_path}/install_dependencies_on_cluster.sh"
 ]
+
+
+# CLI configuration.
+cluster_config_dir = "/config"
+config_name = "config"
+python_cli = "cli.py"
 
 
 # Shared DAG construction parameters.
@@ -150,6 +155,20 @@ def submit_pyspark_job(
                 "spark.kryo.registrator": "is.hail.kryo.HailKryoRegistrator",
             },
         },
+    )
+
+
+def submit_step(cluster_name, step_id):
+    """Submit a PySpark job to execute a specific CLI step."""
+    return submit_pyspark_job(
+        cluster_name=cluster_name,
+        task_id=step_id,
+        python_module_path=python_cli,
+        args=[
+            f"step={step_id}",
+            f"--config-dir={cluster_config_dir}",
+            f"--config-name={config_name}",
+        ],
     )
 
 
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -216,3 +216,13 @@ def delete_cluster(cluster_name: str) -> DataprocDeleteClusterOperator:
         trigger_rule=TriggerRule.ALL_DONE,
         deferrable=True,
     )
+
+
+def generate_dag(cluster_name, tasks):
+    """For a list of tasks, generate a complete DAG."""
+    return (
+        create_cluster(cluster_name)
+        >> install_dependencies(cluster_name)
+        >> tasks
+        >> delete_cluster(cluster_name)
+    )
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import Any
 
 import pendulum
+import yaml
 from airflow.providers.google.cloud.operators.dataproc import (
     ClusterGenerator,
     DataprocCreateClusterOperator,
@@ -216,6 +217,13 @@ def delete_cluster(cluster_name: str) -> DataprocDeleteClusterOperator:
         trigger_rule=TriggerRule.ALL_DONE,
         deferrable=True,
     )
+
+
+def read_yaml_config(config_path):
+    """Parse a YAMl config file and do all necessary checks."""
+    assert config_path.exists(), f"YAML config path {config_path} does not exist."
+    with open(config_path, "r") as config_file:
+        return yaml.safe_load(config_file)
 
 
 def generate_dag(cluster_name, tasks):
--- a/src/airflow/dags/dag_preprocess.py
+++ b/src/airflow/dags/dag_preprocess.py
@@ -7,7 +7,7 @@ from airflow.models.dag import DAG
 
 from . import common_airflow as common
 
-CLUSTER_NAME = "workflow-otg-cluster"
+CLUSTER_NAME = "otg-preprocess"
 
 ALL_STEPS = [
     "finngen",
@@ -20,12 +20,8 @@ with DAG(
     default_args=common.shared_dag_args,
     **common.shared_dag_kwargs,
 ):
-    (
-        common.create_cluster(CLUSTER_NAME)
-        >> common.install_dependencies(CLUSTER_NAME)
-        >> [
-            common.submit_step(cluster_name=CLUSTER_NAME, step_id=step)
-            for step in ALL_STEPS
-        ]
-        >> common.delete_cluster(CLUSTER_NAME)
-    )
+    all_tasks = [
+        common.submit_step(cluster_name=CLUSTER_NAME, step_id=step)
+        for step in ALL_STEPS
+    ]
+    dag = common.generate_dag(cluster_name=CLUSTER_NAME, tasks=all_tasks)
--- a/src/airflow/dags/dag_preprocess.py
+++ b/src/airflow/dags/dag_preprocess.py
@@ -1,0 +1,31 @@
+"""Airflow DAG for the Preprocess part of the pipeline."""
+from __future__ import annotations
+
+from pathlib import Path
+
+from airflow.models.dag import DAG
+
+from . import common_airflow as common
+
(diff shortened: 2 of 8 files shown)


d0choa:
@ireneisdoomed and I failed to make the (commented) test `test_no_import_errors` in tests/airflow/test_dag.py` run.  The test is meant to validate that the DAG will be imported into Airflow without errors. It used to work before the split that created the `common_airflow.py` script, but I think it's now having issues with the fact that it needs to import the other script. 

It feels like this is a critical test to have from a CI/CD point of view.

tskir:
Thank you for the comment @d0choa! I wasn't aware of this commented out test. I'll look into it now.

d0choa:
I know this is not part of this part, but I felt you might know better what the issue might be. I reproduced the issue in a python console:

```python
In [1]: import airflow
In [3]: from airflow.models import DagBag
In [4]: DagBag(dag_folder='src/airflow/dags', include_examples=False)
[2023-11-02T11:08:01.788+0000] {dagbag.py:536} INFO - Filling up the DagBag from src/airflow/dags
/Users/ochoa/Library/Caches/pypoetry/virtualenvs/otgenetics-_BWSZ0mB-py3.10/lib/python3.10/site-packages/airflow/providers/google/cloud/links/dataproc.py:79 AirflowProviderDeprecationWarning: This DataprocLink is deprecated.
/Users/ochoa/Library/Caches/pypoetry/virtualenvs/otgenetics-_BWSZ0mB-py3.10/lib/python3.10/site-packages/airflow/providers/google/cloud/links/dataproc.py:128 AirflowProviderDeprecationWarning: This DataprocListLink is deprecated.
[2023-11-02T11:08:04.379+0000] {dagbag.py:346} ERROR - Failed to import: src/airflow/dags/dag_preprocess.py
  File "/Users/ochoa/Library/Caches/pypoetry/virtualenvs/otgenetics-_BWSZ0mB-py3.10/lib/python3.10/site-packages/airflow/models/dagbag.py", line 342, in parse
    loader.exec_module(new_module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "src/airflow/dags/dag_preprocess.py", line 8, in <module>
    from . import common_airflow as common
ImportError: attempted relative import with no known parent package
[2023-11-02T11:08:04.384+0000] {dagbag.py:346} ERROR - Failed to import: src/airflow/dags/dag_genetics_etl.py
  File "/Users/ochoa/Library/Caches/pypoetry/virtualenvs/otgenetics-_BWSZ0mB-py3.10/lib/python3.10/site-packages/airflow/models/dagbag.py", line 342, in parse
    loader.exec_module(new_module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "src/airflow/dags/dag_genetics_etl.py", line 8, in <module>
    from . import common_airflow as common
ImportError: attempted relative import with no known parent package
Out[4]: <airflow.models.dagbag.DagBag at 0x108dbccd0>
```

tskir:
@d0choa Thank you for the commands! I've created an issue to address this: https://github.com/opentargets/issues/issues/3140

tskir:
@d0choa I've now rebased this branch on top of the recently merged PR, so that we can take advantage of the re-enabled test to check these new changes (it passes here as well)

src/airflow/dags/common_airflow.py
```diff
@@ -200,3 +220,20 @@ def delete_cluster(cluster_name: str) -> DataprocDeleteClusterOperator:
...
+    """Parse a YAMl config file and do all necessary checks."""
+    assert config_path.exists(), f"YAML config path {config_path} does not exist."
+    with open(config_path, "r") as config_file:
+        return yaml.safe_load(config_file)
+
+
+def generate_dag(cluster_name, tasks):
+    """For a list of tasks, generate a complete DAG."""
+    return (
+        create_cluster(cluster_name)
```
ireneisdoomed, line 235: I am now trying to run LD index and have to tweak the cluster parameters.
I think this function could live in a more visible space, I expect that in common we have rather "static" logic.

ireneisdoomed:
@tskir Thanks! This fix is relevant if you want to test this PR https://github.com/opentargets/genetics_etl_python/pull/219

src/airflow/dags/common_airflow.py
tskir: The idea is that a single cluster, under a uniform configuration, will be able to support all kinds of jobs. Very shortly I'll submit another PR which enables autoscaling.

I was wondering if you could please list the tweaks you are applying to the cluster config to run LD index? Then I can make sure that the unified configuration supports that as well :upside_down_face:

ireneisdoomed: This is what i am trying:
```
create_cluster(
                CLUSTER_NAME,
                master_machine_type="n1-highmem-96",
                master_disk_size=5000,
                num_local_ssds=0,
            )
```

src/airflow/dags/dag_preprocess.py
```diff
@@ -0,0 +1,27 @@
...
+
+from pathlib import Path
+
+from airflow.models.dag import DAG
+
+from . import common_airflow as common
+
+CLUSTER_NAME = "otg-preprocess"
+
+ALL_STEPS = [
```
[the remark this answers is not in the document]
tskir: I see what you mean; but as it turns out, reading configuration from YAML is not a good idea in Airflow, in general, for two reasons.

**1. Conceptual.** The _foundation_ of Airflow is "configuration as code": [\[1\]](https://airflow.apache.org/docs/apache-airflow/2.7.1/#workflows-as-code), [\[2\]](https://big-data-test-infrastructure.ec.europa.eu/service-offering/apache-airflow_en). The idea is, because Python is more extensible, expressive and elegant than YAML or JSON, it just makes sense to define all concepts directly, rather reading them from external files.

**2. Technical.** As we discussed previously with Manuel and David, to ensure that the DAG is current, Airflow scans and re-interprets the Python file multiple times per second. Because of that, it is important to make each DAG file as lightweight as possible on importing external libraries and parsing files. And, well, defining a static list of several strings in Python is about _6,700 times faster_ than reading it from an external YAML file (note the change in units between the approaches):

```python
%%timeit -n 1000
steps = ["finngen", "eqtl_catalog", "gwas_catalog", "uk_biobank"]
```
**29.7 ns** ± 0.246 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

```python
%%timeit -n 1000
from pathlib import Path
import yaml
config_path = Path("test.yaml")
assert config_path.exists(), f"YAML config path {config_path} does not exist."
with open(config_path, "r") as config_file:
    steps = list(yaml.safe_load(config_file))
```
**201 µs** ± 2.61 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

So actually, a good idea would be to eventually move away from using YAML in the ETL DAG as well; this would make the entire setup faster and more expressive at the same time

tskir:
@ireneisdoomed Thank you for the fix! I've applied it and rebased this PR on top of it. Autoscaling PR is coming up separately as it's out of scope for this one :upside_down_face:

src/airflow/dags/common_airflow.py
ireneisdoomed: I generated a LD index yesterday based on these params.
When you try this with autoscaling, can you please change the output path to make sure we don't clash? I need this data for subsequent steps.
Also I'll shortly be submitting a PR that integrates this step as part of the preprocessing part.

src/otg/finngen.py
```diff
@@ -37,38 +37,33 @@ class FinnGenStep:
...
-
-        # Sort and partition for output.
-        summary_stats_df.sortWithinPartitions("position").write.partitionBy(
-            "studyId", "chromosome"
-        ).mode(self.session.write_mode).parquet(self.finngen_summary_stats_out)
+        # Process summary stats.
+        summary_stats_df = FinnGenSummaryStats.from_source(summary_stats_df).df
+        # Write summary stats.
+        (
+            summary_stats_df.sortWithinPartitions("position")
```
ireneisdoomed, line 65: Can i ask why the dataset has to be ordered by position?

ireneisdoomed (approved):
Thanks! Looks good.
Just a comment outside the scope of this PR: I hadn't realised that Finngen's study table generation is coupled to SS processing. Right now Finngen's table is static, and we wouldn't need to produce one unless there's a new version out. However we haven't dealt with trait mappings yet, and this process could break the static nature of this step.

d0choa (approved):
+1 to @ireneisdoomed  comments.
Let's try to merge this and we keep moving

src/airflow/dags/common_airflow.py
tskir: Noted about the paths, and thank you for the PR!

src/airflow/dags/dag_preprocess.py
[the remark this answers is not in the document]
tskir: Yes, certainly it is possible to use YAMLs and in some cases even inevitable — for example, if we don't know the structure of the workflow in advance and it is provided externally. Our workflows are, however, always predefined (at least at this point in the development).

As for this part:

> a system that a non expert can tweak

I think that this would be fantastic (really!), but at the same time, given the complexity of the entire project, I think it's pretty impossible :-)

src/otg/finngen.py
tskir: Sure! I made it following existing code for GWAS Catalog by @DSuveges, which partitions by chromosome and sorts by position. I think it's an instinctive move for genetic data to do that. I _think_ (but I'm not sure so don't quote me on this) that on writing the Parquet file Spark remembers the sorts which were done, so if in the future we want to any joins with other data, it should be quicker

tskir:
Thank you @d0choa @ireneisdoomed for constructive and thought-provoking comments, as usual!

PR #218, opened 2023-11-02
Re-enable and fix the `test_no_import_errors` Airflow DAG test

tskir:
Closes https://github.com/opentargets/issues/issues/3140.

For properly resolving all imports, we should change the working directory to `dags/` (which is the way Airflow does it in production), and since this wasn't done, the tests were failing.

Because we are using the `MonkeyPatch` class, directory is only changed for the import fixture and does not affect any other tests in the repository.


d0choa (approved):
Beautiful. Thanks @tskir 
cc @ireneisdoomed

PR #219, opened 2023-11-02
fix(airflow): job args are list of strings

ireneisdoomed:
This affects the way that args are provided for the pyspark job specification. 

DataprocSubmitJobOperator takes args as a list of strings. The previous function accepted a list of strings or a dictionary as inputs. When provided a dictionary, it'd parse it into a list of strings. In https://github.com/opentargets/genetics_etl_python/commit/930155765e7e7c474f877faad0018b329d977e79 I introduced a bug that incorrectly assumed that args were provided as a dict. If a list was provided, formatted_args would remain an empty array. This is actually the cause for the bug reported here https://github.com/opentargets/genetics_etl_python/pull/201#pullrequestreview-1706949353

I've changed it to only give support to have a list as input. It works now.

### DIFF
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -122,7 +122,7 @@ def submit_job(
 
 
 def submit_pyspark_job(
-    cluster_name: str, task_id: str, python_module_path: str, args: dict[str, Any]
+    cluster_name: str, task_id: str, python_module_path: str, args: list[str]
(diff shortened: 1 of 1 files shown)


tskir (approved):
Thank you Irene! Support for dictionary arguments was only there for an approach to radically reduce the number of configuration layers that I made a month ago. Since we're not going with it, only supporting lists is fine for now. Thank you for the fix!

PR #221, opened 2023-11-02
Airflow set up, part 1: Style updates for docs and DAG code

tskir:
First part of updates https://github.com/opentargets/issues/issues/3143, split for readability and easier review. This PR only contains style changes.

* Streamlined style and tone of the documentation
* Fixed formatting of the numbered list and code block under "Advanced configuration"
* Updated all variable names in `common_airflow.py` to upper case in line with PEP8
* Fixed numbered list formatting in the contributing checklist (while I'm at it)

### DIFF
--- a/docs/development/airflow.md
+++ b/docs/development/airflow.md
@@ -14,18 +14,18 @@ This section describes how to set up a local Airflow server which will orchestra
 ## Configure Airflow access to Google Cloud Platform
 
 !!!warning Specifying Google Cloud parameters
-    Run the next two command with the appropriate project ID and service account name to ensure the correct Google default application credentials are set up.
+    Run the next two command with the appropriate Google Cloud project ID and service account name to ensure the correct Google default application credentials are set up.
 
 Authenticate to Google Cloud:
 
 ```bash
-gcloud auth application-default login --project=open-targets-eu-dev
+gcloud auth application-default login --project=<PROJECT>
 ```
 
 Create the service account key file that will be used by Airflow to access Google Cloud Platform resources:
 
 ```bash
-gcloud iam service-accounts keys create ~/.config/gcloud/service_account_credentials.json --iam-account=open-targets-genetics-dev@appspot.gserviceaccount.com
+gcloud iam service-accounts keys create ~/.config/gcloud/service_account_credentials.json --iam-account=<PROJECT>@appspot.gserviceaccount.com
 ```
 
 
--- a/docs/development/airflow.md
+++ b/docs/development/airflow.md
@@ -1,65 +1,67 @@
 # Airflow workflow
 
-The next section describes how to run Airflow workflows locally while performing computation in Google Cloud Platform. This is useful for testing and debugging, but for production use, we recommend running Airflow on a dedicated server.
+This section describes how to set up a local Airflow server which will orchestrate running workflows in Google Cloud Platform. This is useful for testing and debugging, but for production use, it is recommended to run Airflow on a dedicated server.
 
-## Pre-requisites
+## Install pre-requisites
 
 - [Docker](https://docs.docker.com/get-docker/)
 - [Google Cloud SDK](https://cloud.google.com/sdk/docs/install)
 
 !!!warning macOS Docker memory allocation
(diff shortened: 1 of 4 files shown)


docs/development/airflow.md
```diff
@@ -1,65 +1,67 @@
...
 
-Run the next command with the appropriate <PROJECT_ID> to ensure you have Google default application credentials set up:
+!!!warning Specifying Google Cloud parameters
+    Run the next two command with the appropriate project ID and service account name to ensure the correct Google default application credentials are set up.
+
+Authenticate to Google Cloud:
 
 ```bash
-gcloud auth application-default login --project=<PROJECT_ID>
+gcloud auth application-default login --project=open-targets-eu-dev
```
ireneisdoomed, line 22: Why is it better to hard code the project?
In any case, wouldn't it be `open-targets-genetics-dev`?

d0choa, line 22: if we want to hardcode a project in the documentation, we should be using `open-targets-genetics-dev`

docs/development/airflow.md
```diff
@@ -89,13 +91,13 @@ More information on running Airflow with Docker Compose can be found in the [off
...
 1. **Increase Airflow concurrency**. Modify the `docker-compose.yaml` and add the following to the x-airflow-common → environment section:
 
-```yaml
-AIRFLOW__CELERY__WORKER_CONCURRENCY: 32
-AIRFLOW__CORE__PARALLELISM: 32
-AIRFLOW__CORE__MAX_ACTIVE_TASKS_PER_DAG: 32
-AIRFLOW__SCHEDULER__MAX_TIS_PER_QUERY: 16
-AIRFLOW__CORE__MAX_ACTIVE_RUNS_PER_DAG: 1
-```
+    ```yaml
```
d0choa, line 94: None of this will work now because we are not using CeleryExecutor anymore.

src/airflow/dags/common_airflow.py
```diff
@@ -164,11 +164,11 @@ def submit_step(cluster_name, step_id):
     return submit_pyspark_job(
         cluster_name=cluster_name,
         task_id=step_id,
-        python_module_path=python_cli,
+        python_module_path=PYTHON_CLI,
```
ireneisdoomed, line 167: I'd rather pass here the full path to the Python module, rather than building it inside the function. I was struggling yesterday to debug which module exactly we were running, and it was because the logic was buried inside `submit_pyspark_job`
So I suggest something like this:
```suggestion
        python_module_path=f"{INITIALISATION_BASE_PATH}/{PYTHON_CLI}",
```
 or `PYTHON_CLI = f"{INITIALISATION_BASE_PATH}/cli.py`

ireneisdoomed (changes requested):
Thanks! I had some minor comments

docs/development/airflow.md
tskir: Actually, you know what, you're right. I only did this for consistency, because the service account (open-targets-genetics-dev@appspot.gserviceaccount.com) was specified, and project ID wasn't.

However, a better way to achieve consistency is to not hardcode anything in the first place. So this is what I did now. Fixed in https://github.com/opentargets/genetics_etl_python/pull/221/commits/d42fd268f781ef31aaf56911349995d7c64615d1.

docs/development/airflow.md
tskir: Oh, you are right. We are moving so fast, I totally forgot that we god rid of CeleryExecutor in https://github.com/opentargets/genetics_etl_python/pull/201.

However, only the first parameter (`AIRFLOW__CELERY__WORKER_CONCURRENCY`) was actually specific to CeleryExecutor. All of the rest ones apply to Airflow core or scheduler components which are independent from the executor.

I've now clarified this section in https://github.com/opentargets/genetics_etl_python/pull/221/commits/c17fbd239b88bf8515ed629ba88605a315566537.

docs/development/airflow.md
tskir: You're right on both counts. Not hardcoded anymore: https://github.com/opentargets/genetics_etl_python/commit/d42fd268f781ef31aaf56911349995d7c64615d1

src/airflow/dags/common_airflow.py
tskir: Fair enough! Thank you for the suggestion, I'll accept it now

ireneisdoomed (approved):
Ty for the changes!

PR #222, opened 2023-11-02
Airflow set up, part 2: Fixing issue with setting `AIRFLOW_UID`

tskir:
Second part of updates for https://github.com/opentargets/issues/issues/3143, split for readability and easier review.

It is important that `AIRFLOW_UID` in `src/airflow/.env` is the same as the UID of the user running the host operating system. If they don't match, all sorts of "Permission denied" errors start popping up in my setup, for example when Airflow tried to read the credentials file.

In Linux, a standard UID of the first user on the machine is 1000. However, on Mac it's different. So I added a command to populate this dynamically for every given user, and removed it by default. (It's better to see a "user ID not set message", rather than the cryptic permission errors if it is set incorrectly, which I observed in my case.)

@ireneisdoomed @d0choa can you please check that the `id`/`sed` commands work on your setup as intended?


docs/development/airflow.md
```diff
@@ -46,6 +46,16 @@ cd src/airflow
...
 ```
 
+### Set Airflow user ID
+
+!!!note Setting Airflow user ID
+    This command allows Airflow running inside Docker to access the credentials file which was generated earlier.
+
+```bash
+USER_ID=$(id -u)
+sed -Ei "s|^AIRFLOW_UID=.*|AIRFLOW_UID=${USER_ID}|" .env
```
d0choa, line 56: Same thing different style. I have no preference:

```bash
echo -e "AIRFLOW_UID=$(id -u)\nAIRFLOW_GID=0" >> .env
```
The risk of this is accidentally running the command twice. Even with that error that the risk is very low, I'd say.

d0choa: `sed` command runs on my environment. I had bad experiences with sed in the past because the mac implementation wasn't the same as the GNU sed

src/airflow/.env
```diff
@@ -1,4 +1,4 @@
-AIRFLOW_UID=1126896676
+AIRFLOW_UID=
```
d0choa, line 1: having an empty variable looks a bit weird I have to say.

d0choa (approved):
Look at the styling comments. It works for me, so I approve.

docs/development/airflow.md
ireneisdoomed: The proposed command wasn't working for me because of `-i`. As David says, this works differently. Apparently a good xplatform solution is:
```
USER_ID=$(id -u)
sed "s|^AIRFLOW_UID=.*|AIRFLOW_UID=${USER_ID}|" .env > .env.tmp && mv .env.tmp .env
```

tskir: @ireneisdoomed @d0choa Based on your comments, and using some of your suggestions, I've designed a new solution which:
* Does not use `sed`
* Does not include an empty variable declaration in the base `.env` file
* Is safe when you run it multiple times
* Is quite simple

I've updated the PR, please do check that it also works on a Mac, and let me know what you think.

d0choa (approved):
This looks good @tskir. Try to merge it as soon as you can

PR #223, opened 2023-11-02
Airflow set up, part 3: Fixing issue with `docker compose` syntax

tskir:
Closes https://github.com/opentargets/issues/issues/3143.

Third and final part of updates for https://github.com/opentargets/issues/issues/3143, split for readability and easier review.

For me, on Ubuntu, `docker-compose` (with hyphen) works, while `docker compose` (with space) does not:
```
docker: 'compose' is not a docker command.
See 'docker --help'
```

I installed it in a standard way via apt (`sudo apt install docker.io`).

It looks like this distinction has a long history: https://github.com/docker/compose/issues/8630; and certainly there are fixes and workarounds about this.

However, I was wondering: @d0choa @ireneisdoomed, does `docker-compose` (with hyphen) work for you on your machines?
* If yes, we can leave this option as seemingly the most compatible.
* If no, then I'll amend this PR and specify this discrepancy and potential workarounds.

### DIFF
--- a/docs/development/airflow.md
+++ b/docs/development/airflow.md
@@ -65,13 +65,13 @@ mv .env.tmp .env
 Before starting Airflow, initialise the database:
 
 ```bash
-docker compose up airflow-init
+docker-compose up airflow-init
 ```
 
 Now start all services:
 
 ```bash
-docker compose up -d
+docker-compose up -d
 ```
 
 Airflow UI will now be available at `http://localhost:8080/`. Default username and password are both `airflow`.
@@ -90,13 +90,13 @@ docker ps
 To stop Airflow, run:
 
 ```bash
-docker compose down
+docker-compose down
 ```
 
 To cleanup the Airflow database, run:
 
 ```bash
-docker compose down --volumes --remove-orphans
+docker-compose down --volumes --remove-orphans
 ```
 
 ### Advanced configuration
--- a/docs/development/airflow.md
+++ b/docs/development/airflow.md
@@ -65,13 +65,13 @@ mv .env.tmp .env
 Before starting Airflow, initialise the database:
 
 ```bash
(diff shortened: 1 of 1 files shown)


d0choa:
I have no issue using `docker-compose`:

``` bash
❯ ls /usr/local/bin/ | grep docker
com.docker.cli
docker
docker-compose
docker-compose-v1
docker-credential-desktop
docker-credential-ecr-login
docker-credential-gcloud
docker-credential-osxkeychain
docker-diagnose
docker-image-tool.sh
docker-index
kubectl.docker
```

tskir:
Thank you very much for the pointers @d0choa! I investigated further, and apparently the issue was that my Docker installation on Ubuntu was done years ago, before it even had official Ubuntu packages. So all this time I was using an unofficial package `docker.io`, which, while it kept updating the actual Docker versions, didn't port the `compose` command to new syntax.

I've reverted syntax back to "docker compose" and added a note in case anyone encounters this in the future.

d0choa:
We can live without this warning. We can probably assume someone is using a moderately modern version of Docker.

tskir:
@d0choa Here's the thing though: I did have the _latest_ version of Docker installed and still ran into this error. It's not about the version, it's about the packages which were used to install it.

Docker only started officially providing Linux packages [two years ago.](https://docs.docker.com/desktop/release-notes/#400) So everyone who installed it on Debian/Ubuntu more than two years ago using the `docker.io` package (which was the recommended way to install at the time), and then _kept updating to the latest version_ as I did, will still run into this problem.

My opinion remains that having this warning, which might apply to a sizable portion of Debian/Ubuntu users, would be helpful. Especially seeing as we do keep a platform specific warning about macOS memory issues :upside_down_face:

PR #224, opened 2023-11-03
feat(airflow): move `variant_annotation` and `ldindex` steps to preprocessing dag

ireneisdoomed:
Moving steps between DAGs is trivial.
This PR includes:
- relocation of `variant_annotation` and `ldindex` steps to preprocessing dag, these are datasets that only need changing when a major gnomad version is released. They also happen to be the steps that require the most time to run
- this way we can tweak the ETL DAG without worrying about messing our input data
- On the flip side, it becomes less obvious to know the dependencies of our ETL steps

**Current ETL DAG:**
image

**Current Preprocessing DAG:**
image

### DIFF
--- a/src/airflow/dags/configs/dag.yaml
+++ b/src/airflow/dags/configs/dag.yaml
@@ -1,10 +1,8 @@
 - id: "gene_index"
 - id: "variant_annotation"
-- id: "ld_index"
 - id: "gwas_catalog"
   prerequisites:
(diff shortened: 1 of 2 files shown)


d0choa (approved):
Makes sense. In the future we can think about whether all preprocess steps belong to the same DAG 🤔

tskir (approved):
Thank you Irene! Makes sense

PR #226, opened 2023-11-05
Simplify configuration and remove config store

tskir:
Closes https://github.com/opentargets/issues/issues/3145.

### Root cause of the bug
I traced the issue back to the new config syntax introduced recently in https://github.com/opentargets/genetics_etl_python/pull/67:

```yaml
defaults:
  - session: dataproc
session:
  start_hail: true
```

Even though it looks intuitive (to me, too!), unfortunately it doesn't work as intended.

### Explanation (maybe)
It's hard to pinpoint exactly why this happens. Hydra is complex, and while I've read a lot of documentation and discussions in GitHub issues, I don't think I have the full picture.

I _think_ it happens because configuration for `session` is provided in two separate places: both `dataproc.yaml` / `local.yaml`; and inside the step configs. And it looks like Hydra just does not support injecting values into a class from multiple locations. Anyway, I couldn't find an example in their documentation which would match the syntax above.

### Fix
This is the way I reimplemented things:
1. As before, default dataset (`gcp`) and session (`dataproc`) configurations are specified in the main `config.yaml`.
2. If a step wishes to specify any additional modifications to the session, they are provided inside a special field called `extra_session_params`. They are applied _on top of_ the values from the session configuration specified in `config.yaml`,  and override them if needed.
3. At the time of instantiating everything, `cli.py` checks for this field and passes the parameters to the Session constructor.
4. After a Session object has been constructed, the additional parameters are removed from the config, and a Step object is constructed using the Session object from before.

I verified that, when running things like I described in the original issue, the parameters are indeed passed to the Session object as expected.

P.S. I also removed two parts of the code which are no longer needed:
* `SessionConfig` from the config store, as we can inject it directly, same as any other configs;
* `start_hail` param from inside LDIndexStep, as at that point it is not used and cannot affect anything.


tskir:
@d0choa @DSuveges @ireneisdoomed I've now resolved conflicts against main again, and also removed an additional unused `start_hail` parameter in `VariantAnnotationStep` (as it's a session adjustment, not a step one, and doesn't have any effect when defined there)

ireneisdoomed (commented):
Thank you for picking this up, it was a great finding.
I think Hydra is composing the configuration as expected, the problem is that we are overwriting it in each step when we do:
```
class Step:
    session: Session = Session()
``` 
Calling a default Session won't connect to yarn, it will create a local session.
If you remove this from the step, configuration is built as intended.

Some general comments related to this approach. Initially I applied the changes following a similar pattern, that is, adding logic about how a session would look like inside `cli.py`. However, we agreed that this is probably not the best place to include it:
- Technically we are calling `cli.py` for each step. That means that we are instantiating the session class as many times as steps we run.
- I believe that when we need to extend the default Spark config it is because a specific step requires it, therefore it makes sense that fine tuning the session is a configuration specific to the step.
- For example, you mention that “start_hail” is a session adjustment. It is, yes, but it is also a step requirement that only applies in the context of the variant annotation step and a few others.


In sum, I am more in favour of the approach that we had in place, we just need to remove the default “Session” from each step to avoid the configuration override.

This is what I did [here](https://github.com/opentargets/genetics_etl_python/commit/d021e6cdb73b85efb73998ed223b3cad211efa76) for the Variant Annotation, there's more context in the [draft PR](https://github.com/opentargets/genetics_etl_python/pull/237).

tskir:
@ireneisdoomed 

> the problem is that we are overwriting it in each step when we do

Of course! I can't believe I didn't see it myself when investigating. Now it all makes sense.

Yes, I agree that your approach without the extra logic in `cli.py` is more elegant.

In this case please do finalise your draft PR with the changes and I'll review once ready.

(However, please also keep this PR open for now, as some of my other work is built on top of it. Once we merge yours and I rebase my work on top of it, I'll close it.)

And thank you for taking the time to look into this, implement, and explain :upside_down_face:

ireneisdoomed:
@tskir Daniel encountered that the same bug was causing his tests in https://github.com/opentargets/genetics_etl_python/pull/236 to fail and fixed all the steps.
Once this is merged (I just had minor comments), you can pull from main.

tskir:
@d0choa @DSuveges @ireneisdoomed I have rebased this PR against main and updated based on my idea, which ended up working after a few tweaks. Essentially, now we:
* Don't need to specify `defaults → session: dataproc` in every step
* Don't need the config store and registering configs _at all_
* And the entire thing is ~100 lines of code lighter

This all comes down to how Hydra composes configuration. The main thing which was required was to specify the composition in this exact order:
* `- datasets: gcp` → load input/output file paths
* `- step/session: dataproc` → apply default session parameters
* `- step: ???` → apply step specific configuration, including session parameter overrides

I have verified that the entire thing works, and the config is composed the way we expect to.

config/step/session/local.yaml
```diff
@@ -1,3 +1,3 @@
-defaults:
-    - session_config
+_target_: otg.common.session.Session
+spark_uri: local[*]
 hail_home: .venv/lib/python3.10/site-packages/hail/backend/hail-all-spark.jar
```
ireneisdoomed, line 3: This was wrong, local hail_home doesn't have a fixed location like in Dataproc.
We usually find it after running: `poetry run pip show hail | grep Location | awk -F' ' '{print $2 "/hail"}'`
I suggest that this is required when in local model
```suggestion
hail_home: ???
```

ireneisdoomed (changes requested):
These changes make sense to me and reduce the number of layers.
A summary of how I understand this works:
- `config.yaml` is now the "parent" config, instead of `config.py`. We lose config validation, but we gain simplicity
- Here we set that all steps by default will extend the Session config for Dataproc. This is cool, because it means that we don't have to call a default config per step.
- At the step level, we define extra parameters for Session, and Hydra will compose them. These extra parameters don't need to be part of the default config, they just have to be arguments to the Session class (`start_hail`, for example)

I've tested the changes in several steps:
- Removing `config.py` doesn't break our CLI. I still see the list of available steps when I do `otg --config-dir="/config" --config-name="config"`
- Extending the Session to provide extra spark config in L2G works
- A normal step without extra parameters works after fixing a bug in the Spark write mode
- Steps dependent of Hail don't work, but that is because of the bugs fixed in  https://github.com/opentargets/genetics_etl_python/pull/237

I've requested a minor change in the local config, after that I think it is good to go.

config/step/session/local.yaml
tskir: Thank you! This makes sense

tskir:
@ireneisdoomed Thank you for such an extensive review & testing!

ireneisdoomed (approved):

PR #227, opened 2023-11-05
Enable autoscaling, tune cluster parameters, add docs

tskir:
Closes https://github.com/opentargets/issues/issues/3142.

Now that https://github.com/opentargets/issues/issues/3145 is addressed, enabling autoscaling is easy ;-)

I have verified that it works using LD index as an example. I didn't run it to completion, but it has a very characteristic CPU load profile where we can immediately see the difference. On a single 96 core VM, graph width = 6 hours:

And on an a autoscaled cluster of 20 VMs (plot width = same 6 hours):

(@ireneisdoomed Sorry again for the clash, I didn't immediately realise you were running the step at the same time as I did — will be more vigilant going forward & change the output paths, etc.)

The cluster was very stable at 2 primary + 20 secondary workers (I even tested randomly killing a few secondary workers — it recovers beautifully), so I increased the maximum number of secondary VMs to 50 in the autoscaling policy. I'll monitor performance and adjust as needed going forward.

I also added some basic documentation on the high level organisation of the pipeline, including which DAGs we have, and how autoscaling works. (Feel free to suggest alternative places where this piece of documentation can be moved if you have some ideas)

### DIFF
--- a/docs/development/airflow.md
+++ b/docs/development/airflow.md
@@ -1,4 +1,4 @@
-# Airflow workflow
+# Airflow configuration
 
 This section describes how to set up a local Airflow server which will orchestrate running workflows in Google Cloud Platform. This is useful for testing and debugging, but for production use, it is recommended to run Airflow on a dedicated server.
 
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -23,6 +23,7 @@ GCP_PROJECT = "open-targets-genetics-dev"
 GCP_REGION = "europe-west1"
 GCP_ZONE = "europe-west1-d"
 GCP_DATAPROC_IMAGE = "2.1"
+GCP_AUTOSCALING_POLICY = "otg-etl"
 
 
 # Cluster init configuration.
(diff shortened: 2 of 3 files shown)


src/airflow/dags/common_airflow.py
```diff
@@ -45,7 +46,7 @@
 # Shared DAG construction parameters.
 shared_dag_args = dict(
     owner="Open Targets Data Team",
-    retries=3,
+    retries=1,
```
tskir, line 49: BTW, the reason for this change: when using PySpark + Dataproc, in my experience, 99% of errors are due to mistakes in the code and not some transient failures. In this case, restarting the step 3 times just wastes time & money, instead of failing immediately and allowing to start investigations as to the cause of failure

docs/development/airflow.md
```diff
@@ -10,6 +10,9 @@ This section describes how to set up a local Airflow server which will orchestra
 !!!warning macOS Docker memory allocation
     On macOS, the default amount of memory available for Docker might not be enough to get Airflow up and running. Allocate at least 4GB of memory for the Docker Engine (ideally 8GB). [More info](https://airflow.apache.org/docs/apache-airflow/stable/howto/docker-compose/index.html#)
 
+!!!warning Making sure the Docker set up is up to date
```
d0choa, line 13: this was in another PR

d0choa (approved):

d0choa:
@tskir try to merge when you can

docs/development/airflow.md
tskir: Yep — now resolved. I added a comment in the closed PR about this: https://github.com/opentargets/genetics_etl_python/pull/223#issuecomment-1794523985

PR #234, opened 2023-11-06
GnomAD variants within class configuration

d0choa:
Along the same lines as #233, refactor of variant annotation to move configuration that it's datasource-specific to the class itself.

It also includes the chain file within the project, so there is one less external dependency


DSuveges (approved):
All makes sense.

PR #236, opened 2023-11-07
Small update in functionality to the GnomAD datasource toolset

DSuveges:
- A method is added to extract an arbitrary square slice of the LD sparse matrix of a given population.
- 🙀tests to make sure the function returns the right data in the right format.
- A (hopefully) temporary fix to allow hail to be used in the testing environment.


tests/datasource/gnomad/test_gnomad_ld.py
```diff
@@ -65,3 +69,63 @@ def test_resolve_variant_indices(
...
+    @pytest.fixture(scope="class")
+    def _setup(self, spark: SparkSession) -> None:
+        """Initialize GnomADLDMatrix object."""
+        hl.init(sc=spark.sparkContext, log="/dev/null")
+
+        self.gnomad_ld_matrix = GnomADLDMatrix(
+            ld_matrix_template="tests/data_samples/example_{POP}.bm"
+        )
+
+    def test_ld_matrix_slice(self) -> None:
```
ireneisdoomed, line 89: Ideally I'd break this logic into different tests. 
What you can do:
1. Add `matrix_slice` to `_setup` so that it can be used as a fixture within the class
```
self.matrix_slice = (
            self.gnomad_ld_matrix
            # Extract a 2 by 2 squary matrix:
            .get_ld_matrix_slice(
                gnomad_ancestry="test-pop", start_index=1, end_index=2
            ).persist()
        )
```
2. Break it into different tests. E.g.:
- Type assertion of `self.matrix_slice`
- Count assertion of `self.matrix_slice`
- Assertion that transversing of `self.matrix_slice` worked 

Just a suggestion if you think it's a good idea. The main point is to make tests as unitary as possible.

src/utils/spark.py
```diff
@@ -10,6 +11,7 @@ def get_spark_testing_conf() -> SparkConf:
     Returns:
         SparkConf: SparkConf with settings for testing.
     """
+    hail_home = hl.__file__.replace("/__init__.py", "")
```
ireneisdoomed, line 14: To make more obvious that what we want here you can do this (Path has to be imported):
```suggestion
    hail_home = Path(hl.__file__).parent.as_posix()
```

I did something very similar yesterday, and added hail_home as a fixture. If you think it's a good idea, you can copy this https://github.com/opentargets/genetics_etl_python/pull/237/commits/6b737fc318646ed33990e530d6986f0cff4d9dc7

tests/datasource/gnomad/test_gnomad_ld.py
```diff
@@ -65,3 +69,63 @@ def test_resolve_variant_indices(
...
+            ),
+            on=["idx_i", "idx_j"],
+            how="inner",
+        )
+
+        # Is the matrix complete:
+        assert compared.count() == matrix_slice.count()
+
+        # Is the matrix symmetric:
+        assert compared.filter(f.col("r") == f.col("r_sym")).count() == compared.count()
```
ireneisdoomed, line 131: Nice!

ireneisdoomed (approved):
Thank you!
I think the logic is good, just added some details.
As a note: this PR also fixes the bug reported here https://github.com/opentargets/issues/issues/3145

tests/datasource/gnomad/test_gnomad_ld.py
DSuveges: Thanks, now the single test method was broken down to 4.

src/utils/spark.py
DSuveges: Yeah, much nicer.

PR #237, opened 2023-11-07
fix(session): hail config was not set unless extended_spark_conf was provided

ireneisdoomed:
This PR includes:
- bugfix in `_create_merged_config` where the hail configuration was not set unless  `extended_spark_conf` was not null (https://github.com/opentargets/genetics_etl_python/commit/995392b38d2d98943b8bc54c4d41305d312bf641). This caused that, at the time of running variant annotation, it would fail because the necessary Hail config was not provided.
- Added `test_hail_configuration` (https://github.com/opentargets/genetics_etl_python/commit/6b737fc318646ed33990e530d6986f0cff4d9dc7)
- bugfix in `submit_pyspark_job` where the path to the cli was being resolved again (https://github.com/opentargets/genetics_etl_python/commit/3faec29df216003265520c02bee4594c1cbd3cca)
- bugfix in the step, where the default session was overriding the parameters provided in the config YAML as reported [here](https://github.com/opentargets/issues/issues/3145) (https://github.com/opentargets/genetics_etl_python/commit/d021e6cdb73b85efb73998ed223b3cad211efa76) 
- ~bugfix in the session configuration, where the `start_hail` parameter of the step was not being picked up~ (https://github.com/opentargets/genetics_etl_python/pull/237/commits/69cb5c13fba97fc0c3a73f51a306f74c099e5d42)
    - ~as a downside, currently `start_hail` has to be explicitly set in every step's config~
- bugfix that goes back to using the remote chain file because of what i think it is a bug in Hail reported here https://github.com/hail-is/hail/issues/13993

**With these fixes, variant annotation's job completed in 33 mins thanks to the autoscaling policy ([job](https://console.cloud.google.com/dataproc/jobs/65e83f76-be1f-4e7b-84f5-5d770be216cf/monitoring?region=europe-west1&project=open-targets-genetics-dev))**


src/airflow/dags/configs/dag.yaml
```diff
@@ -1,24 +1,20 @@
...
+# - id: "ukbiobank"
+# - id: "study_locus_overlap"
+#   prerequisites:
+#     - "gwas_catalog"
+#     - "ukbiobank"
+# - id: "locus_to_gene"
+#   prerequisites:
+#     - "variant_index"
+#     - "v2g"
+#     - "study_locus_overlap"
```
tskir, line 20: I'm not sure if this change is intended?

src/otg/datasource/gnomad/variants.py
```diff
@@ -95,7 +92,8 @@ def as_variant_annotation(self: GnomADVariants) -> VariantAnnotation:
         # Select relevant fields and nested records to create class
         return VariantAnnotation(
             _df=(
-                ht.select(
+                ht.head(30)
```
tskir, line 95: Same as above; this looks like a temporary change made for testing?

tskir (changes requested):

src/airflow/dags/configs/dag.yaml
ireneisdoomed: Thank you! Now fixed in https://github.com/opentargets/genetics_etl_python/pull/237/commits/d651a033e1ebd5df6ed6031279c54bc2ae64b42d

config/step/ld_index.yaml
```diff
@@ -2,5 +2,6 @@ _target_: otg.ld_index.LDIndexStep
 
 session:
   start_hail: true
-
+grch37_to_grch38_chain_path: ${datasets.chain_hail_37_38}
+ld_index_raw_template: ${datasets.ld_index_raw_template}
```
tskir, line 6: This change doesn't look right in relation to the changes in `config/datasets/gcp.yaml`: while we are removing `chain_hail_38_37` from `gcp.yaml`, we are adding `datasets.chain_hail_37_38`. This looks like it wouldn't work because the line in `gcp.yaml` is removed

tskir (changes requested):

config/step/ld_index.yaml
ireneisdoomed: Great catch! Thanks. These params are not used after https://github.com/opentargets/genetics_etl_python/pull/233/files

PR #239, opened 2023-11-08
Type enhancements

d0choa:
Context:
We have used a very relaxed mypy setup on our local environment and the CI/CD pipelines. Consequently, several bugs have been introduced that could have been picked up by a proper type validator.

This PR introduces several enhancements that should ensure better type-testing in the future:
- `py.typed` file introduced for stubs
- [semi-strict](https://emplocity.com/en/about-us/blog/mypy_not_enough/) configuration of mypy
- minor differences between local mypy setup and mypy in the pre-commits. Local mypy setup goes further and validates against the dependency types. This is something that pre-commit can't handle without installing the packages. I discarded this option because it just takes too long and pre-commit.ci timeouts.  
- All changes required in the current `main` to comply with type validation

What this PR doesn't do:
- it doesn't try to alter the business logic. Only in the case in which obvious mistakes were committed I tried to fix them.

Beware, once this PR is merged the bar will raise and typing issues can emerge on other branches that were so far passing validation


src/otg/dataset/dataset.py
```diff
@@ -58,7 +60,7 @@ def schema(self: Dataset) -> StructType:
 
     @classmethod
     @abstractmethod
-    def get_schema(cls: type[Dataset]) -> StructType:
+    def get_schema(cls: type[Self]) -> StructType:
```
DSuveges, line 63: What is the advantage of using `Self` from the typing_extension library over just adding the name of the class?

src/otg/dataset/intervals.py
```diff
@@ -72,8 +72,8 @@ def from_source(
             raise ValueError(f"Unknown interval source: {source_name}")
 
         source_class = source_to_class[source_name]
-        data = source_class.read(spark, source_path)
-        return source_class.parse(data, gene_index, lift)
+        data = source_class.read(spark, source_path)  # type: ignore
```
DSuveges, line 75: Do these comments `# type: ignore` have any meaning?

src/otg/datasource/gwas_catalog/study_index.py
```diff
@@ -243,7 +243,7 @@ def from_source(
         catalog_studies: DataFrame,
         ancestry_file: DataFrame,
         sumstats_lut: DataFrame,
-    ) -> StudyIndex:
+    ) -> GWASCatalogStudyIndex:
```
DSuveges, line 246: Great catch! objects of class `GWASCatalogStudyIndex` are also instances of the `StudyIndex` as well. Is this something mypy enforces or you have just seen it?

src/otg/datasource/ukbiobank/study_index.py
```diff
@@ -49,7 +49,7 @@ def from_source(
         Returns:
             UKBiobankStudyIndex: Annotated UKBiobank study table.
         """
-        return StudyIndex(
+        return UKBiobankStudyIndex(
```
DSuveges, line 52: I don't really understand this. I understand we have a class to generate UKBB study index. However there's no ukbb study index datatype. When saying `UKBiobankStudyIndex.get_schema()` it's just `StudyIndex.get_schema()` because of the inheritance.

DSuveges (commented):
Most of the modifications are straightforward, some are a bit obscure to me. Overall, it seems things are changing for the better.

src/otg/dataset/dataset.py
DSuveges: If this practice is better, shouldn't we change everywhere?

ireneisdoomed: The particularity here is that `get_schema` is called from the child classes of Dataset.
Here it is saying that cls is not just any Dataset type, but the type of the child class that is currently being defined. 
For example, StudyLocus is technically of type Dataset, but mypy would complain saying that the returned StructType is generic to the Dataset class, not StudyLocus's.

.pre-commit-config.yaml
```diff
@@ -53,13 +53,27 @@ repos:
...
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.6.1'
+    rev: "v1.6.1"
     hooks:
       - id: mypy
+        args: [
+            # --ignore-missing-imports is required to run mypy on virtual
+            # environments without installed dependencies
+            "--ignore-missing-imports",
```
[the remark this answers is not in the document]
d0choa: pre-commit does not install the environment so it can't chase the dependency types (e.g. hail). This setting is different in the TOML that feeds the vscode warnings/errors. VScode is therefore more strict than the pre-commit

.pre-commit-config.yaml
```diff
@@ -82,4 +96,3 @@ repos:
     rev: 0.3.8
     hooks:
       - id: pydoclint
-        args: [--style=google, --check-return-types=True, --allow-init-docstring=True, --config=pyproject.toml src]
```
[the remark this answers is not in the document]
d0choa: They are being picked up from the TOML. They were duplicated. I confirmed with @ireneisdoomed

src/otg/common/schemas.py
```diff
@@ -67,49 +64,3 @@ def flatten_schema(schema: t.StructType, prefix: str = "") -> list:
         else:
             fields.append(Field(name, dtype))
     return fields
-
-
-def _get_spark_schema_from_pandas_df(pdf: PandasDataFrame) -> t.StructType:
```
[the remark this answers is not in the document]
d0choa: it's not used anywhere. @ireneisdoomed confirmed

src/otg/dataset/intervals.py
d0choa: it's telling mypy not to handle this line. This code would need a proper refactor but I thought it was not worth the effort.

src/otg/datasource/gwas_catalog/study_index.py
d0choa: the problem starts when you start using the resulting object. Methods for the child class are not "typed" to be available if you return the parent type. In these cases the more specific the better.

src/otg/datasource/ukbiobank/study_index.py
d0choa: I think it's the same comment as the above, but let me know if it's not clear

DSuveges (approved):
Thanks for the explanation

PR #240, opened 2023-11-09
Access gnomad matrix slice with resolved variant identifiers

DSuveges:
The new function returns a dataframe with a melted square ld matrix with resolved variant identifiers. As the test passes, I assume things are alright. 

**HOWEVER**: to avoid mypy meltdown, I had to make sure the function always return a dataframe with the same type. Which is kind of alright, however, if data is not available, I had to initialize an empyt spark dataframe out of thin air. To do so, I had to call `SparkSession.bulider.getOrCreate()` In [theory this is safe to do so](https://stackoverflow.com/a/44504213).


src/otg/datasource/gnomad/ld.py
```diff
@@ -313,6 +315,103 @@ def as_ld_index(
...
+                & (f.col("position") <= end)
+            )
+            .select("chromosome", "position", "variantId", "idx")
+            .persist()
+        )
+        # +----------+---------+--------------------+--------+
+        # |chromosome| position|           variantId|     idx|
+        # +----------+---------+--------------------+--------+
+        start_index = self._get_value_from_row(
+            ld_index_df.orderBy(f.col("position").asc()).first(), "idx"
```
d0choa, line 370: Perhaps use window? https://github.com/opentargets/genetics_etl_python/blob/1b9a37b88cedaee2e62e02900d463292f794d283/src/otg/common/spark_helpers.py#L157

DSuveges: I don't think it helps: I don't need to window, I need the minimum from the entire dataframe and also when I get the value from the returned dataframe I might as well fall back to `None`.

d0choa (approved):

PR #241, opened 2023-11-09
feat: hydra full error on dataproc

d0choa:
Use HYDRA_FULL_ERROR=1 in dataproc by default to have more verbose errors. Otherwise, errors are occasionally so succinct that it's hard to know what's happening.

After failing with a few setups, this produces the intended results.


ireneisdoomed (approved):
This is great, thank you!!!

PR #242, opened 2023-11-09
fix: wrong python file uri in airflow

d0choa:
Python module path was malformed. It looked like `gs:.../.../gs:../.../.../` due to excessive concatenation of variables.

This fix should rescue all airflow runs. @ireneisdoomed came out with the same fix separately.


tskir (approved):
Thank you @d0choa!

PR #244, opened 2023-11-10
fix(gwas_catalog): clump associations, remove hail and style fixes

ireneisdoomed:
This PR includes:
- I've added calling `clump()` which got removed accidentally
- Removal of hail initialisation (on a side note, the fact that Hail was initialising despite having set `start_hail=False` makes me think that sessions are not overwritten.
- Minor style changes

The step runs in 20'


tskir (approved):
Looks good to me, thank you Irene!

PR #252, opened 2023-11-14
fix: persist raw gwascat associations to return consistent results

ireneisdoomed:
This PR surpasses the behaviour described in https://github.com/opentargets/issues/issues/3152 

I've checked various studies and they look correct. The one reported in the issue:
```python
df = session.spark.read.parquet("gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/credible_set/catalog_curated")
df.filter(f.col("studyId") == "GCST000058").select("studyId", "variantId").show()
+----------+--------------+                                                     
|   studyId|     variantId|
+----------+--------------+
|GCST000058|6_38473194_A_G|
+----------+--------------+
```

### DIFF
--- a/src/otg/gwas_catalog.py
+++ b/src/otg/gwas_catalog.py
@@ -60,7 +60,7 @@ class GWASCatalogStep:
         )
         catalog_associations = self.session.spark.read.csv(
             self.catalog_associations_file, sep="\t", header=True
-        )
+        ).persist()
         ld_index = LDIndex.from_parquet(self.session, self.ld_index_path)
 
         # Transform


DSuveges (commented):
I think it worth explore why the problem occured in the first place and address it in the business logic level, however if it works, let's go with this solution for now. 

I have only one question before approving: Is there a way we can test that variants and studies are not mixed up, eg. the `studyLocusId` is good? Eg.

```
+-------------+-----------------+----------------+
|      studyId|   older_variants|   newer_variant|
+-------------+-----------------+----------------+
|   GCST000002|  [5_9332169_T_A]|  4_18031865_T_C|
+-------------+-----------------+----------------+
```
The ids in old and the new runs can not be the same, right? Assuming the problem happens after the ID generation, meaning the id is correct, this can be detected at the last step. Does it make sense?

DSuveges:
> @DSuveges studyLocusId is indeed not the same. This is assigned at the end of generating a studyLocus when splitting. So if the variant is shuffled, studyLocusId will differ. This is in fact how I identified the error.

Oh, you are right, you need to have the actual study id and resolved variant id to generate the study-locus id, so this step must be at the end of the proccess. I find this behaviour of spark very unsettling and it would be great to find the best solution and also knowing the source would allow us avoid such mistake later. 

OK, if this solution works, let's go with it, but at some point we need to get back to it.

DSuveges (approved):

PR #254, opened 2023-11-14
Adjust running parameters to increase stability

tskir:
Closes https://github.com/opentargets/issues/issues/3154, please see the detailed description there.


tskir:
@d0choa 

NB: I have tested all changes in this PR _except_ for `.repartition(len(input_filenames) * 16)`, which is provisional.

As we discussed, it might be better to use `coalesce` to avoid triggering a reshuffle? Or alternatively we may want to  `repartition` by studyId + chromosome from the start, seeing as this is the output partitioning we want. I'm still trying to understand exactly how repartitioning and reshuffling work, so not sure which solution is necessarily the best

tskir:
Actually, thinking about the above, I've removed the `.repartition(len(input_filenames) * 16)` update from this PR. Reasoning:
* I've not tested it, so it may potentially cause more harm than good
* It is performance related, not stability related. We can evaluate performance later

d0choa (approved):
100% agree with all these changes

DSuveges (approved):
All changes looks solid.

PR #256, opened 2023-11-17
Enable local SSDs for worker nodes

tskir:
@d0choa @ireneisdoomed @DSuveges Please see the results of evaluating various performance approaches in https://github.com/opentargets/issues/issues/3155.

This PR implements the only tangible performance improvement that actually helps: attaching local SSDs to instances. Airflow's `ClusterGenerator` does not support this natively, and the existing `num_local_ssds=1` which is passed to it does nothing.

However, by editing the resulting config directly we can achieve this.

Closes https://github.com/opentargets/issues/issues/3155.


d0choa (approved):
looks good. I think the maths of having this by default are reasonable, but please double check in case I missed something.
P.S. the fact that now it works as expected is also positive ;)

PR #257, opened 2023-11-18
Extra parametrisation on airflow common

d0choa:
This PR adds parameters to airflow common functions in order to be able to:
- Parametrise autoscaling policy - different DAGs might have very different requirements
- Add task_id (airflow) different than the `step_id`. This allows the reuse of steps for different processes
- `other_args` in the submission of pyspark allows to parametrise the submission of a step (e.g. inputs/outputs)
- `trigger_rule` allows to specify that pyspark jobs might run on "ALL_DONE" instead of "ALL_SUCCESS". This is very powerful while developing. Combined with the spark `writemode`, it allows the running of subsequent steps even though the upstream tasks failed because the outputs were already available.

The logic exploiting these new parameters will come in different PRs


tskir (approved):
Thank you @d0choa! All looks very sensible and useful

PR #266, opened 2023-11-21
Updating summary stats schema and ingtesion

DSuveges:
Details under this ticket: [3039](https://github.com/opentargets/issues/issues/3039)

- Removing upper and lower confidence interval of beta
- Adding support for variant wise sample size.
- Sample size parsed from GWAS Catalog summary stats.
- Sample size added to study locus (collected field after the window based clumping)
- Sample summary stats dataset updated (columns removed)
- Sumstats mock logic updated.
- New filter applied: dropping associations with beta = 0
- Ingestion updated for GWAS, Finngen.

There are a handful of logic defined under the the above mentioned ticket, that I haven't addressed yet. I feel they have marginal importance in terms of scale:

```
 For the rest of snps if se==0 and p-value!=1 and p-value!=0 -> infer se from p-value and beta, otherwise drop
 For the rest of snps if p-value==1 or p-value==0 - infer it form beta and se
 If p-value is too small (0) even after inferring -> replace zero with 2e-308
```


d0choa (approved):
hopefully we won't change this in a long time :)

PR #267, opened 2023-11-21
Removing odds ratio, and confidence intervals from the schema

DSuveges:


tskir (approved):
Thank you! Makes sense in line with the previous changes

PR #268, opened 2023-11-22
fix: multiply standard error by zscore in `calculate_confidence_interval`

ireneisdoomed:
The function that calculated the confidence interval was not taking into account the 95% z score.
This bugfix has no impact on the current data, as this is a function we were not using at the moment.


DSuveges (approved):
Good catch! Thanks.

PR #269, opened 2023-11-22
Fix local SSD attachment syntax

tskir:
My apologies, there's a missing field in the chain which slipped through refactors and rebases. I've verified that this now works and the cluster is created without an issue.

Please feel free to merge this PR immediately once approved.

### DIFF
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -106,7 +106,9 @@ def create_cluster(
             # Create a disk config section if it does not exist.
             cluster_config[worker_section].setdefault("disk_config", dict())
             # Specify the number of local SSDs.
-            cluster_config[worker_section]["num_local_ssds"] = num_local_ssds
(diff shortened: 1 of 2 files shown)


src/airflow/dags/common_airflow.py
```diff
@@ -106,7 +106,9 @@ def create_cluster(
             # Create a disk config section if it does not exist.
             cluster_config[worker_section].setdefault("disk_config", dict())
             # Specify the number of local SSDs.
-            cluster_config[worker_section]["num_local_ssds"] = num_local_ssds
+            cluster_config[worker_section]["disk_config"][
```
DSuveges, line 109: How did this work before?

tskir: This did work at first; then I was rebasing a previous PR to modify the list of changes included with it and accidentally broke it; and now after the fix it works again

DSuveges (approved):

d0choa:
Thanks for the quick fix!

PR #271, opened 2023-11-24
Il gwascat dag

ireneisdoomed:
This PR defines the tasks that need to happen to process GWASCatalog data in a DAG. The process starts with the ingested harmonized summary stats flat files and finishes with the definition of credible sets.

It is built on top of 'do_clump_step'. It is WIP, as I don't love the current solution of splitting both clumping methods into different steps.
Writing the studies before splitting is terrible as well. 
I prototyped a combined clumping step in https://github.com/opentargets/genetics_etl_python/pull/271/commits/ce28e0449fd32df1b58f58531f517a22e8a89fdf but I don't like it either. I need to think about it.


d0choa:
There is a bunch of potential clashes with other PRs. Let's agree on the steps we want to have. Treat them as individual PRs and then create the separate DAGs as independent PRs.

Once all the steps have been merged, we can see if the DAGs are still working before merging.

Step-wise, we either need to:
- split the 2 clumps in 2 steps
- treat gwas catalog curation and gwas catalog SS as different streams with some but not many repeated calls to similar functions.

I'm more inclined for the second because we don't need to write useless temporary files (e.g. window-based clumped study-locus before LD clumping)

ireneisdoomed:
Superseded by https://github.com/opentargets/genetics_etl_python/pull/291 and https://github.com/opentargets/genetics_etl_python/pull/288

PR #272, opened 2023-11-24
feat: finngen preprocess prototype

d0choa:
DAG responsible for Finngen preprocessing:
- harmonisation + study index
- distance-based clumping + LD annotation + LD clumping
- PICS

A full compute of this DAG was completed taking 3h + 13 min + 1min on the specified infrastructure. This DAG should only be needed once per Finngen release or the last 2 steps in case LD index changes.

**Notes:**
- The harmonisation of the summary stats is the bottleneck here. Input files are read from `tsv.gz` and processed in one single node. One study takes somewhere between 45 and 60 min to run.  Even with 1 core per phenotype, ~10% jobs fail because of the load. This starts another round of dealing with the failed jobs and subsequently until the task is finished. Due to autoscaling, machines are decommissioned and cost goes down.
- In line with other data sources, semi-static configuration now lives in the data source class. The expectation is that the data source parser will only work using default parameters with the latest public release of finngen. This has some implications on the way the functions are tested but this has been addressed.


config/datasets/gcp.yaml
ireneisdoomed, line 1: Inputs in `gcp.yaml` have been refined in https://github.com/opentargets/genetics_etl_python/pull/326
We need to consolidate this config by adding the finngen related SS.

src/airflow/dags/common_airflow.py
```diff
@@ -86,7 +86,7 @@ def create_cluster(
         zone=GCP_ZONE,
         master_machine_type=master_machine_type,
         worker_machine_type=worker_machine_type,
-        master_disk_size=500,
+        master_disk_size=2000,
```
ireneisdoomed, line 89: Do we need this size by default or is it only useful for Finngen?

src/otg/datasource/finngen/study_index.py
```diff
@@ -24,35 +22,45 @@ class FinnGenStudyIndex(StudyIndex):
...
+
     @classmethod
     def from_source(
         cls: type[FinnGenStudyIndex],
-        finngen_studies: DataFrame,
-        finngen_release_prefix: str,
-        finngen_summary_stats_url_prefix: str,
-        finngen_summary_stats_url_suffix: str,
+        spark: SparkSession,
+        finngen_studies_json: str = "",
```
ireneisdoomed, line 36: I don't love setting an empty string as a default.
Any downside to treating it as None and doing:
````
if finngen_studies_json:
            with open(finngen_studies_json) as finngen_studies:
                json_data = finngen_studies.read()
        else:
            json_data = urlopen(cls.finngen_phenotype_table_url).read().decode("utf-8")
        rdd = spark.sparkContext.parallelize([json_data])
        raw_df = spark.read.json(rdd)
```

ireneisdoomed (approved):
Looks good! Changes to the logic make sense, only a minor comment.
I've checked that input files to the preprocessing DAGs were up to date. As a general comment, we need to extract preprocessing datasets like Finngen's SS out of the release related bucket.

src/airflow/dags/common_airflow.py
d0choa: parametrised in a new commit

PR #282, opened 2023-11-27
fix: proper parsing of gwas catalog study accession from filename

DSuveges:
I have added a function to parse the GWAS Catalog study accession from the path/filename. I think we should be agnostic on the path, however the assumption is reasonable that the filename has to have the id.


d0choa:
if you prefer to parametrise the study id that's ok with me. We can  externalise the logic to a different layer (airflow), so no info is encoded in the path

DSuveges:
> if you prefer to parametrise the study id that's ok with me. We can externalise the logic to a different layer (airflow), so no info is encoded in the path. It used to be the case

It's not really external to the process: you throw the path/filename to the function and the function parses the study id. What I prefer about it is that this function asserts if the filename does contain a valid gcst identifier and fails if not. I think spark would just leave an null column. (but I might be wrong with this)

d0choa (approved):

PR #288, opened 2023-11-29
feat: add `clump` step

ireneisdoomed:
This PR is coauthored with @d0choa and based on his branch https://github.com/opentargets/genetics_etl_python/pull/258.

What makes the most sense to me is having a single clumping step where we call each method upon request depending on the stage of the pipeline that we are at.
This is essentially moving away from source-centric steps into process-centric steps.

The way that I thought about it in the first place was by flagging which clumping method we want to use depending on the input data. However I shrinked the logic because as of today:
- we only do window based clumping on a `SummaryStatistics` dataset
- we only do LD based clumping on a `StudyLocus` dataset

Depending on the input data, we choose one journey or the other.


.pre-commit-config.yaml
d0choa, line 1: this doesn't belong here

d0choa:
I'm having second thoughts about all this. @DSuveges can we get your input here?

The point here is to make a clump step that doubles for LD clumping and window-based clumping. In order to do this, @ireneisdoomed has done some work on handling the 2 types of inputs and deciding what needs to be done for them. I think the intention is to have one step instead of two. Although I thought it could be a good idea, I'm not so positive anymore.

My main problems now are 2:
- There is a "guessing" stage that I don't like. When an input is provided, the step is trying to guess what that input is. It's a bit artificial because we already know what we are providing. I suggested @ireneisdoomed to do an alternative approach in which we provide the 2 paths `studylocus_path` or `summary_stats_path`. But I understand is still a bit messy to check that  the logic needs to check that only one is provided, etc. This guessing stage also might run into not some (not very serious) compute waste. Reading all parquet GWAS catalog harmonised summary stats as it's done here takes some time (1-2min). Not to read the content, just to deal with all the files in the ~half-million partitions. Forcing the step to guess and having to read the parquet schemas twice looks unnecessary when we already know what we are providing.
- The other problem is related to the return. Although both will be returning a `studyLocus` the content of the `studyLocus` would be very different. I find it a bit obfuscated to explain that 2 things might come out of the same step, they look alike but they are not the same.

All things considered. My recommendation is to break this into 2 steps: window-based and ld clumping. If the 2 have different inputs and different outputs. And we want to use them in some cases but not in others. I think it's justified to use 2 different steps. It will make every step easier to understand. What comes in and what comes out.

ireneisdoomed:
Why I decided to do it this way:
- Clump is the same concept for me, having different methods just mean different applications. Ideally, I'd integrate both methods into StudyLocus/Summary statistics.clump() with a switch deciding which to use
- Having a common step deduplicates everything: the step itself, its configuration and it's documentation
- Although this is not the case at the moment, both methods should be susceptible of being applied on both datasets. To me, this is a reason for having an agnostic step. 
- I don't consider that having non fixed inputs/outputs is a problem. This is an artifact of having the same dataset representing different stages. 
- Regarding the part where we look at the data type, I don't think that reading the metadata is an expensive task. As far as I know, when you read you don't load the whole dataset, Spark will only read its metadata. 

In sum, to me it makes sense having a recipe for pie and letting the ingredients define its flavour, than dealing with a recipe for lemon pie, apple pie,...
At the same time the implications are not so terrible, so I can always split it if you think the losses outweigh the gains.

DSuveges:
Conceptually speaking, summary statistics and study locus aren't that much different: to me, window based clumping should be applicable on both types. This is what is also aligned with @addramir's idea, that both datasets should be pushed through the same process. I would think about how to write the window based clumping in a way that allows this. Then this method could be inherited by both datasets. (This effort is somewhat related to these other tickets: 
- [Generalize window based clumping to SL](https://github.com/opentargets/issues/issues/3130)
- [introducing single point associations with shared methods](https://github.com/opentargets/issues/issues/3087)

If this is done, one step could take care of all the datasets to PICS.

src/otg/clump.py
```diff
@@ -0,0 +1,70 @@
...
+                )
+            self.data = StudyLocus.from_parquet(self.session, self.input_path)
+            ld_index = LDIndex.from_parquet(self.session, self.ld_index_path)
+            study_index = StudyIndex.from_parquet(self.session, self.study_index_path)
+
+            clumped_study_locus = self.data.annotate_ld(
+                study_index=study_index, ld_index=ld_index
+            ).clump()
+        else:
+            self.data = SummaryStatistics.from_parquet(self.session, self.input_path)
```
DSuveges, line 63: Correct me if I'm wrong, but it seems, summary statistics aren't LD clumped. Is it correct?

src/otg/dataset/study_locus.py
```diff
@@ -368,6 +369,22 @@ def annotate_credible_sets(self: StudyLocus) -> StudyLocus:
...
+        """Annotate LD information to study-locus.
+
+        Args:
+            study_index (StudyIndex): Study index to resolve ancestries.
+            ld_index (LDIndex): LD index to resolve LD information.
+
+        Returns:
+            StudyLocus: Study locus annotated with ld information from LD index.
+        """
+        from otg.method.ld import LDAnnotator
```
DSuveges, line 384: Having an import statement in a function definition seems quite uncanonical to me. Is it intentional?

src/otg/clump.py
ireneisdoomed: Yes. The step performs one clumping method only. If you want to clump twice, you call the method twice. 

DAG for summary stats is (assuming these are harmonized already):

Distance Clumping (SS automatically return SL) -> LD Clumping (SL automatically return SL) -> PICS

As you mentioned, when we give support to perform both methods on both data types, you'll have to choose which method you want to use. This is how I wrote it originally here 

https://github.com/opentargets/genetics_etl_python/commit/317380222fa460ffb6368aad4c6f996970cd46bb

src/otg/dataset/study_locus.py
ireneisdoomed: It's a common practice to avoid the problem of dependency circularity

LD annotator imports from StudyLocus
StudyLocus imports from LD annotator 

Importing LDAnnotator only when it's needed removes this. You're right in that in general it's more pythonic having them at the top.

d0choa:
@ireneisdoomed let's discuss this briefly later and we can merge it. We don't want to get stuck on this and we can iterate in the future

ireneisdoomed:
Thank you both for reviewing

PR #291, opened 2023-12-01
feat: add gwas_catalog_preprocess dag

ireneisdoomed:
This PR includes:
- bugfix in `submit_step` where the ID for the job was derived from the step, and not the task_id that we define. I had clashes when adding several calls to the clump step
 - changes to the gwas_catalog.py step to stop the journey when the associations table is converted into a StudyLocus; and when the studies table is parsed into a split StudyIndex
 - addition of a new DAG that processes all GWASCatalog data. The new DAG is meant to be run as part of the preprocessing and consists of 2 groups of tasks:
     1. Processing summary statistics
         - Conversion of harmonised summary stats to a StudyLocus after window based clumping
         - LD clumping of the SS StudyLocus
         - Fine mapping with PICS of clumped study locus
    2. Processing of GWAS Catalog curation
        - Creation of a StudyLocus from curation and a StudyIndex
        -  LD clumping of the curated StudyLocus
         - Fine mapping with PICS of clumped study locus

image

**In the future, I'd like to see these 2 streams of data consolidated right before running LD clumping.** That way, we could operate with a single dataset in all downstream processes.


d0choa (approved):
Good stuff.

PR #295, opened 2023-12-03
feat: implement the new `resilient_fetch` module

tskir:
This is the first PR in a series of three for https://github.com/opentargets/issues/issues/3158.

### Problem
When you are reading a URL (HTTP, HTTPS, FTP) as a byte stream using `urllib.request.urlopen` or similar, sometimes there are intermittent network errors. This happens especially often when you open multiple connections to the same server, which happens in our use case. When this happens, the entire pipeline fails.

### Solution
Implement a module which reads the input URL block by block. If a block fails, it is retried without having to download or re-process the entire file. The retry operations have delays between them, and are handled completely seamlessly. Downstream processing can just open `ResilientFetch(uri) as bytes_stream` and not worry about anything.

### Additional notes
Before writing this, I of course checked if any existing modules achieve this functionality. Unfortunately, there are none. The closest one, https://github.com/berdario/resumable-urlretrieve, has not been updated for 4+ years, only supports fetching into a file on disk rather than into a byte stream, and only supports HTTP and not FTP.

Once we have some time, I believe we could and should publish this as a separate PyPi module and make the wider community aware of its existence.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.deptry]
-extend_exclude = ["src/conftest.py", "src/airflow", "src/utils"]
+extend_exclude = ["src/conftest.py", "src/airflow", "src/utils", "src/batch"]
 
 [tool.deptry.per_rule_ignores]
 DEP001 = ["gentropy"]
@@ -125,7 +125,7 @@ color = true
 exclude = ["dist"]
 
 [tool.pytest.ini_options]
-addopts = "-n auto --doctest-modules --cov=src/ --cov-report=xml"
+addopts = "-n auto --doctest-modules --cov=src/ --cov-report=xml --ignore ./src/batch"
 pythonpath = [".", "./src/airflow/dags"]
 
 # Semi-strict mode for mypy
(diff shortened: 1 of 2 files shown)


ireneisdoomed (commented):
This tool allows you to load batches of bytes stream from any HTTP/HTTPS/FTP urls, with the particularity that if any network problems occur, only failed batches are retried.

I don't have comments about the code (maybe the @opentargets/be-team should take a look?), but I have tested that it worked:
```
ftp_url = "ftp://ftp.ebi.ac.uk/pub/databases/opentargets/platform/latest/input/target-inputs/tractability/tractability.tsv"
http_url = "http://ftp.ebi.ac.uk/pub/databases/opentargets/platform/latest/input/target-inputs/tractability/tractability.tsv"

with ResilientFetch(ftp_url) as f:
         df = pd.read_csv(f, sep="\t")
         print(df.head())
>>> 
```

In general, i think this is a fantastic solution that is useful outside this project. Considering the seemingly simple architecture of the module, I'd suggest publishing it independently. It doesn't seem like a huge task and we'd avoid ""polluting"" the project. What do you think?

DSuveges:
> I'd suggest publishing it independently. It doesn't seem like a huge task and we'd avoid ""polluting"" the project. What do you think?

I agree. Just thinking about our tasks for the upcoming releases, we'll likely need to do things around PIS, which tool is already dealing with fetching data from ftp/http, so it would look nice to have the approach unified across our resources. This needs to be investigated though. We can even consider publishing on pypi for the sake of tidyness. 

On the other hand I that would mean yet and other repository to maintain, but I feel the gain is big enough. And yes, this is not quite genetics related.

PR #306, opened 2023-12-05
test: try to fix Codecov configuration

tskir:
There was this dependabot PR which I confirmed was working and merged: https://github.com/opentargets/genetics_etl_python/pull/301

The tests were passing in the PR, but after merging they failed in the main branch with the following error message:

`[2023-12-04T20:52:25.943Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Could not find a repository associated with upload token ***', code='not_found')}`

Our project indeed has a `CODECOV_TOKEN` secret, which is picked up by the workflow. It was updated 15 hours ago, and it looks like the tests have started failing since then.

Codecov documentation [mentions](https://docs.codecov.com/docs/frequently-asked-questions) (first question, last paragraph) that this token is not required for public repositories, so my idea is that by removing this entirely we might resolve the problem.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -35,7 +35,6 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage.xml
           fail_ci_if_error: true
           verbose: true


tskir:
Ah, no, I think I've got it. _Despite what the documentation says_ (unexpected right?), it looks like starting with Codecov v4 upload token [is actually required](https://github.com/codecov/codecov-action/blob/main/README.md#v4-beta-release) even for public repositories. I've now updated it and restarted the test for the main branch, see if it works

tskir:
Yes, that helped!

image

Closing this PR as no longer necessary.

d0choa:
Indeed, I updated the token yesterday to see if that resolved the issues. The Codecov upload is a bit of a hit-and-miss at the moment. They have been recently acquired and it doesn't seem to be in a stable state. I've been monitoring whether people will be jumping out or using any alternative, but there is some inertia at the moment.

PR #310, opened 2023-12-06
perf: improve join between gold standards and overlaps

ireneisdoomed:
Overlaps is used to build the gold standard in the step where redundant associations are removed. With the current approach, the size of the overlaps dataset represents a bottleneck in the L2G pipeline.
This PR contains:
1. Joining operation is now split into two steps. This only impacts performance, business logic should remain the same:
- Filtering overlaps to include only associations present in the gold standard. Overlaps are on the left side of the join, and gold standard is broadcasted.
- Joining of the overlapping associations onto the gold standards. Gold standards are on the left side of the join, overlapping associations is broadcasted.
2. `_convert_to_square_matrix` only returns relevant columns, this is to minimise data that needs to be shuffled

I've run this solution and I don't see the problem anymore.

### DIFF
--- a/src/otg/dataset/l2g_gold_standard.py
+++ b/src/otg/dataset/l2g_gold_standard.py
@@ -117,7 +117,19 @@ class L2GGoldStandard(Dataset):
         Returns:
             L2GGoldStandard: L2GGoldStandard updated to exclude false negatives and redundant positives.
         """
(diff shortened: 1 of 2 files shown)


ireneisdoomed:
The solution here didn't fix the issue.

PR #313, opened 2023-12-08
refactor: stop inheriting datasets in parsers

d0choa:
Attempt to stop using inheritance in the parsers (`otg.datasource`). As @DSuveges proposed, parsers should not inherit the dataset classes.

During the refactoring, I realised we were using the parsers both as parsers and as a way to iterate on instances of the resulting object. This was only the case for the GWAS Catalog in which there is some amount of manipulation of the dataset after parsing that it's still data source specific (e.g. splitting). To circumvent this issue, I created 2 extra dataset classes (StudyIndexGwasCatalog and StudyLocusGWASCatalog) that inherit from their respective parent classes and contain helper functions to iterate over the resulting data source-specific dataset.

I think the result is reasonable


d0choa:
As suggested by @DSuveges, The datasource-specific datasets belong to the `datasource` not to the `dataset`

DSuveges (approved):

PR #322, opened 2023-12-11
feat: ruff as formatter

d0choa:
This PR intends to use ruff as a formatter, as we were already using it as linter.
As a consequence, this PR deprecates black and isort from the project, making dependencies/extensions/pre-commits a bit lighter.
I also updated some of the ruff lintr rules by taking them from [here](https://github.com/home-assistant/core/blob/dev/pyproject.toml)


ireneisdoomed (approved):
Thank you for the changes!!!

PR #326, opened 2023-12-12
chore: review study locus and study index configs

ireneisdoomed:
This PR includes several changes to the steps configurations that are primarily derived from harmonising the locations of study index and study locus.
More specifically:
- All study indices are now below `study_index`
- Recursive reading of StudyIndex and StudyLocus
- input data in `pics.yaml` is set as required, without the default paths
- Renaming of `study_locus` to `credible_set` to be more specific about the dataset that we are referring to. This implies doing things like `credible_set = StudyLocus...`, which is not wrong but it doesn't look very explicit.
- In `l2g.py`, removed the saving step as this is something handled by the trainer when providing `model_path`

It's the config setup that I've used to run the latest L2G version.


d0choa (approved):
nice cleanup

PR #366, opened 2023-12-19
feat(eqtl): add preprocessing dag

ireneisdoomed:
This PR includes:
 - new DAG of steps in order to process the whole preprocessing of eQTL Catalog
 - redefined `eqtl_catalogue` step to `eqtl_catalogue_ingestion`. This is the script that produces a study index and harmonises their raw summary statistics.

When running the step, I hit a bottleneck at the `eqtl_catalogue_ingestion` level and the job was cancelled.
I want to test it with the recent fixes introduced in https://github.com/opentargets/genetics_etl_python/pull/369, it might work now


ireneisdoomed:
I don't think the changes here are relevant any more.
- The plan is to define credible sets from eQTL Catalogue fine mapping results directly, and not from fine mapping summary stats for PICS.
- This PR had also a technical limitation in terms of how the inputs were handled in the DAG, so I never managed to do a full run.

Since we don't have any business logic here and we only touch the Airflow layer, let's just close this.

PR #367, opened 2023-12-20
fix: incorrect parsing of `app_name` in makefile

ireneisdoomed:
## Context
I was not able to build the package because `APP_NAME` was incorrectly set. This variable is parsed from the TOML. After recent changes in this file, APP_NAME was rendering additional lines beyond the intended name line:
```
otgenetics
true
B007,
B023,
ICN001,
N804,
N805,
N815,
PLC0414,
T100,
PLW2901,
```

## Solution
This PR includes:
- Change in the assignment of `APP_NAME` to only look for the first line that starts with "name".
- Minor improvements in other rules:
    - There was an explicit line in `make build` to remove the content of `dist/`. This is redundant, because we call clean when we build.
    - Properly set up the `build` dependency of `make create-dev-cluster` and `make update-dev-cluster`.


d0choa (approved):
I was experiencing the same issue. Thanks for solving it

PR #371, opened 2023-12-21
chore(study_index): change numeric columns from long to integers

ireneisdoomed:
This PR contains:
- change in the datatype of `nSamples`, `nControls`, `nCases`, `discoverySamples.sampleSize`, `replicationSamples.sampleSize` **from long to integer**
- adjustment of different Study Index parsers to follow the new schema

## Rationale
This came out as a result of the failing tests in [another PR](https://github.com/opentargets/genetics_etl_python/pull/369) that parsed eQTL'sCatalogue study index.
I don't see any reason why these numeric fields have to be of LongType, conceptually they're just integers.
On top of that, integers are more memory efficient and in general more compatible with datatypes in other libraries.


DSuveges (approved):
All trivial.

DSuveges:
This update totally makes sense.

PR #383, opened 2024-01-03
feat(study_locus): remove statistics after conditioning from schema

ireneisdoomed:
This PR includes:
- deletion of the variant statistics after conditioning from the StudyLocus schema

## Context
Inside a `locus`, we have a couple of statistics that describe the association between a variant and the trait after doing conditional analysis.
COJO does conditioning, and that is why we were maintaining these fields. However our current implementation does not include adjusting the p values based on other signals in the locus.

Given the current size of our locus object, I think we're better off without these null fields polluting the codebase.

### DIFF
--- a/src/otg/assets/schemas/study_locus.json
+++ b/src/otg/assets/schemas/study_locus.json
@@ -166,18 +166,6 @@
             },
             {
               "metadata": {},
-              "name": "pValueMantissaConditioned",
-              "nullable": true,
-              "type": "float"
-            },
-            {
-              "metadata": {},
-              "name": "pValueExponentConditioned",
-              "nullable": true,
-              "type": "integer"
-            },
-            {
-              "metadata": {},
               "name": "beta",
               "nullable": true,
               "type": "double"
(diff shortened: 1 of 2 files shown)


src/otg/assets/schemas/study_locus.json
addramir, line 1: I agree with the changes. But if we want to get rid of COJO stats we need to also rename - "logABF" to "logBF".

addramir (changes requested):
I agree with the changes. But if we want to get rid of COJO stats we need to also rename - "logABF" to "logBF" in study_locus.json

ireneisdoomed:
> if we want to get rid of COJO stats we need to also rename - "logABF" to "logBF"

@addramir Thank you for the comment, one question about colocalisation. The COLOC implementation i know uses ABFs to do the hypothesis testing (see [here](https://github.com/opentargets/genetics_etl_python/blob/8f962094dbb11282ee835a360b0b625dcb0d842e/src/otg/method/colocalisation.py#L176)). Can I double check that we can just exchange ABFs with BFs and achieve the same?

addramir:
Yes, you can. It is basically the same, just some terminological issues.

addramir (approved):
As discussed with Irene, for changing ABF to BF it is better to open another PR.

PR #384, opened 2024-01-04
feat: redesign `StudyLocusOverlap` model to include overlapping variants in struct

ireneisdoomed:
This PR includes significant modifications to the `StudyLocusOverlap` dataset schema:
1. The schema has been restructured to consolidate `leftLocus` and `rightLocus` information. Instead of having separate entries for each variant in overlapping `StudyLocusId` pairs, the new schema represents these in a unified manner*.
    - Each `StudyLocusId` pair now corresponds to a single row, simplifying the overlap representation between `leftStudyLocusId` and `rightStudyLocusId`.
     - Changed the `statistics` field, that included some of the stats both left and right for `leftLocus` and `rightLocus`. These follow same schema and have the same content as `locus`, with the only exception that the variants in these arrays are just the overlapping ones (therefore, both arrays contain the same variants).
2. Adaptation of other components that depended on overlaps:
    - `Coloc` and `Ecaviar` now anticipate the grouped schema
    - `L2GGoldStandard.filter_unique_associations` is now simpler with the grouped schema

*Roughly, we have shifted from this schema:
```
leftStudyLocusId
rightStudyLocusId
tagVariantId
statistics
--- leftPosteriorProbability
--- rightPosteriorProbability
...
```
to this one
```
leftStudyLocusId
rightStudyLocusId
leftLocus
--- variantId
--- posteriorProbability
rightLocus
--- variantId
--- posteriorProbability
...
```

**Pending to be run ATM**


ireneisdoomed:
Proposed changes are memory intensive and not scalable by the time each pair of overlapping studyloci carried thousands hundreds of common variants + their statistics.
Example of a failing [job](https://console.cloud.google.com/dataproc/jobs/045f8f69-939f-444b-88b5-fd97b9ed5cb6?region=europe-west1&project=open-targets-genetics-dev). By not exploding data, I generate tasks that are unbalanced and cannot scale by raising more infrastructure.
On top of that, overlaps require to be exploded to be used downstream by eCAVIAR and COLOC.
For all these reasons, we have decided to keep the schema in an exploded version as it is now.

PR #386, opened 2024-01-05
feat: application configuration

d0choa:
This is yet another revised version of the application configuration.

Since the removal of the config store, the application could not run independently of a configuration YAML. This was a major limitation, as we could not expect a user of the package, to create a YAML every time they want to run a step.

The intention of this PR is to allow running the application in at least 3 ways:

a) as a CLI providing the appropiate parameters necessary to run a step, or overriding the default parameters
b) as a CLI in combination with YAML configuration to provide parameters
c) imported as a package

This PR implements this by recovering the configstore and the step configuration dataclasses. They contain all the necessary knowledge about what is required to run every step, what's required and what's not and some default parameters. By using, the config store, the application is aware of what are the available steps and what are the necessary parameters to run each of the steps. The application complains otherwise.

Our layer of YAML configuration `/config` is therefore no longer a dependency of the package. It only contains, the necessary information to populate/override the parameters that we are interested on for our own use-case of the package. This layer is getting slimer and slimer and I can see how the airflow layer should completely take over as it became our interface with the python package. Now that the YAML is no longer a dependency of the package, we could try to migrate the config to Airflow.

To further document how to use the package I added 2 how-to guides:

- how to run a step using the CLI
- how to run a step using the CLI + YAML configuration

I also added some additional tests for the CLI. I hope they become living documentation that ensure new steps are properly included and available to the CLI.

I also updated the airflow DAGs. This is not really tested so I hope I made no mistakes.


docs/howto/run_step_using_config.md
```diff
@@ -0,0 +1,49 @@
...
+
+    ``` yaml
+    defaults:
+        - gene_index
+
+    target_path: /path/to/target
+    gene_index_path: /path/to/gene_index
+    ```
+
+    This config file will inherit the default configuration for the `gene_index` step and overwrite the `target_path` and `gene_index_path` parameters.
```
ireneisdoomed, line 43:
```suggestion
    This config file will inherit the default configuration for the `gene_index` step (the name in `register_config`) and overwrite the `target_path` and `gene_index_path` parameters.
```

tests/test_cli.py
ireneisdoomed, line 1: 2 comments:
- `unittest.mock` is in the standard library and it has a similar functionality to pytest_mock. Maybe we can avoid using the new dependency
- I think that `expected` doesn't need to contain the full expected error. It's valid if it's just a substring, so that we can avoid having to update all possible steps.

I've tested sth very similar to this, and the test passed. Give it a try if you want:
```
"""Test the command-line interface (CLI)."""
from unittest.mock import patch

import pytest
from hydra.errors import ConfigCompositionException
from omegaconf.errors import MissingMandatoryValue
from otg.cli import main


def test_main_no_step() -> None:
    """Test the main function of the CLI without a valid step."""
    override_key = "step"
    expected = f"You must specify '{override_key}"

    with patch("sys.argv", ["cli.py"]), pytest.raises(
        ConfigCompositionException, match=expected
    ):
        main()
```

ireneisdoomed (approved):
Great addition. I still haven't tried it in Airflow, but all components are there. I have added a few suggestions.
My summary after review:
- We now register a configstore where each node is a dataclass for each step. For our use case, YAMLs take the parameters in the config store as reference and extend them.
- Every YAML has the same attributes as the respective config store, so that they can be overriden.
- The session class is registered at the time of instantiating the step.
- Steps are not dataclasses anymore. All the logic has been moved to the generator method.
- `local.yaml` (as opposed to `dataproc.yaml`) has been removed. Default params are provided in the config class.
- `cli.py` and the DAGs have been adapted to accommodate new changes

src/otg/clump.py
```diff
@@ -1,9 +1,7 @@
 """Step to run clump associations from summary statistics or study locus."""
 from __future__ import annotations
 
-from dataclasses import dataclass, field
-
-from omegaconf import MISSING
+from typing import Optional
```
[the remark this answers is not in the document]
d0choa: step has been removed by subsequent PRs

docs/howto/run_step_using_config.md
d0choa: I think you are seeing this too much from the perspective of a developer. This tutorial is about someone who wants to run the package, after PIP-install and they might not know anything about how it's written. I believe `register_config` will mean nothing to them.

d0choa:
I've been thinking about your points @tskir. I think it's important to think twice about these kind of changes.

The problem that this PR comes to solve is the fact that the YAML was a dependency of the package. Someone could not pip install the package and run a step with `otg step=XXX` without having a yaml along it. This comes to protect the user of the package rather than the developer. Now the CLI will tell you what are the required parameters necessary to run any step and complain if you don't provide them. This is something that the YAML-only version can't do. I know the CLI use case is one you are not particularly keen but I think it's important to simplify how users can run the code beyond our use cases. This also adds another benefit which is a better separation of concerns between what the package allows (configstore) and what we use the package for (YAML).

Talking about the problems. I also see a bit of redundancy. The configs and the steps both have a very similar set of parameters (not exactly the same). I think there is room to think about how to use the config classes to parametrise the steps without having to write/document every parameter again. I think this is the only layer that could be considered redundant and imply some maintanance in my opinion. The other bit to reduce and potentially eliminate are the YAML configurations for the steps. A much more slimmer Airflow configuration with a list of inputs/outputs could be enough and we could superseed the current configs/ directory. Instead of the current `Airflow -> configs -> otg` we would like a `configs -> Airflow -> otg` setup.

As for hydra, I had similar thoughts in the past about whether is worth it for our case. But after thinking quite a lot about it, I concluded we are using it for quite a lot of things: composition, interpolation, instantiation, cli,... If we wouldn't use hydra we would probably require quite a lot of code to a reasonably similar functionality.

src/otg/config.py
```diff
@@ -0,0 +1,315 @@
...
+    start_hail: bool = False
+    extended_spark_conf: Optional[dict[str, str]] = None
+    _target_: str = "otg.common.session.Session"
+
+
+@dataclass
+class StepConfig:
+    """Base step configuration."""
+
+    session: SessionConfig = SessionConfig()
```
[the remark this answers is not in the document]
d0choa: this was indeed an issue. Now it's solved

tests/test_cli.py
d0choa: Migrated to `unittest.mock` and dependency removed. Good tip

PR #396, opened 2024-01-09
feat(colocalisation): new step runs ecaviar on credible sets

ireneisdoomed:
This PR includes:
- Changes in the colocalisation step to run eCAVIAR on all credible sets.
- For the time being, I've switched COLOC off.

## QC
Results are here: `gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/colocalisation`
- Total: 10,474,884
- Mean CLPP: 0.134
- # GWAS associations colocalising w/ eQTL: 2,387,864
- 0.8 cutoff:
    - count: 746,515
    - mean: 0.95
    - Results with just 1 colocalising variant: 491,595

**Question:**
Do we need millions of colocalisation results where PP~0.1? I don't think these are relevant for a data consumer. It could help L2G deprioritise genes with low PP, but it's not very different than not having them as the feature'd be 0. We can test the impact.

## Context
The step doesn't have the overlaps dataset as starting point, we simply start off credible sets for which we calculate the overlaps on the go. In a conversation with @d0choa, we agreed that the overlaps set needed to run COLOC and eCAVIAR aren't the same. For eCAVIAR we simply need information about the variants that are common between diff studyLocus, whereas in COLOC we need all variants from both pairs. That is why we conveyed that instead of writing the dataset in an optimised format, we would compute an exploded version (potentially huge, but parallelizable) whenever these are needed.


DSuveges (approved):
All looks pretty straightforward. The 0.8 cutoff you reported was also applied on production?

ireneisdoomed:
@DSuveges No, I'm pretty certain the coloc dataset wasn't filtered in production either. I was just raising the question.

d0choa:
Coloc in production required 100 overlapping tags which resulted in a relatively high baseline H3/H4. This was discussed as too stringent and something to avoid. Of course, this comes at the cost of a lot of weak signals as you are observing.

If you think handling the data is not an issue and we won't pay reasonable performance costs, I would keep all the data points. There might be a marginal value in differentiating overlapping locus with high metrics vs low metrics vs non-overlapping locus.

PR #398, opened 2024-01-09
feat(finemapping): susie method

Daniel-Considine:
Adding SUSIE Class and methods for fine-mapping.


d0choa (approved):
I think it's good to go. We'll see how things go once we start using the method.
@Daniel-Considine feel free to merge when you are ready

PR #400, opened 2024-01-10
feat(l2g): add coloc based features

ireneisdoomed:
This PR includes changes in the L2G pipeline to bring eCaviar based features.
- Business logic to extract the feature is practically unchanged, with the exception that the maximum CLPPs in the neighborhood are in logarithmic scale. L2G should learn to prioritise genes where this value is more negative.
- Bug fixes in how features are built into the feature matrix
- Removal of the overlap dependency. This adds extra logic to extract overlaps only for the study loci in the curation set that doesn't look very pretty.


src/otg/method/l2g/feature_factory.py
```diff
@@ -96,57 +96,93 @@ def _get_max_coloc_per_study_locus(
...
-        # Max LLR calculation per studyLocus AND type of QTL
+        # Max PP calculation per studyLocus AND type of QTL
         local_max = get_record_with_maximum_value(
             colocalising_study_locus,
             ["studyLocusId", "right_studyType", "geneId"],
             "coloc_score",
-        )
+        ).persist()
+
+        intercept = 0.0001
```
ireneisdoomed, line 106: The value of the maximum PP in the neighborhood is the log of the difference between the maximum PP for the gene and the maximum PP across genes. I add this intercept so that when these 2 match up I can calculate the log.

src/otg/l2g.py
```diff
@@ -145,7 +162,7 @@ def __post_init__(self: LocusToGeneStep) -> None:
                 study_locus=credible_set,
                 study_index=studies,
                 variant_gene=v2g,
-                # colocalisation=coloc,
+                colocalisation=coloc,
             )
 
             # Join and fill null values with 0
```
addramir, line 168: Do we add 0 instead of Null in features?

src/otg/method/l2g/feature_factory.py
addramir: I don't fully understand the logic here. The difference can be negative.

src/otg/l2g.py
ireneisdoomed: That's right. Most models can't handle missing values so you need to treat them somehow at the preprocessing step. This is what we used to do in production. However, XGBoost has a smart strategy to handle nulls and it's been in my "to try" list. XGBoost is like an ensemble of decision trees, and theoretically when there's a null in a node it evaluates the performance in each possible direction and it goes in the direction where the loss is smaller. 


I haven't experimented with this yet, so I can't estimate the impact. Implementation wise, it should be as easy as retraining without filling nulls. Should I give it a go?

A smarter way to go around nulls is to infer them. But I think we'd need different imputation strategies per feature.

d0choa: my previous experience with gradient boosting was as well to fill nulls with 0s. I see why the original L2G had to do the same and why this is our default strategy. Imputation is usually not recommended when the proportion of nulls is too big (as is our case for some features).

Exploring the impact of dealing differently with the nulls is potentially useful but goes beyond this PR. We can queue it with all the other L2G improvements that @addramir has in mind and revisit it after first release.

d0choa (approved):
Getting there...

PR #402, opened 2024-01-11
chore(studylocus): rename logABF statistics to logBF

ireneisdoomed:
This PR includes:
- Renaming of `locus.logABF` to `locus.logBF` in the `StudyLocus` schema. SuSiE outputs the statistical measure between a variant and a trait in the form of bayes factors. We used to capture the approximation of this value because of GCTA-COJO, our previous fine mapping algorithm.
- Updates in the documentation and codebase


d0choa (approved):
it's crazy the amount of changes required to rename a column in this case

ireneisdoomed:
@d0choa In the schema, we could abstract the name of the statistic we use. Or capture both and just abstract COLOC. It is equally valid regardless of the fine mapping results you come with. But it's easier like this for now.

PR #404, opened 2024-01-12
feat: remove ukbiobank and overlaps steps

d0choa:
Deprecated steps.

- Logic for ukbiobank study index is still available in the datasource for future implementation but the step is not ready for main.
- Overalps are now temporarily generated in other steps, but not as a step itself


ireneisdoomed (changes requested):
We need to remove the step from the config store as well (`OverlapsIndexConfig`)

ireneisdoomed (approved):
Great, thansk!

PR #406, opened 2024-01-12
feat: flex python version to make all 3.10 versions compatible

d0choa:
Requiring a very specific version of python package 3.10.8 might be very stringent. This PR attempts to allow the package installation in any 3.10.X version as a first step.


d0choa:
I changed my local python version to 3.10.13 (pyenv + poetry), built a wheel, uploaded it to the bucket, and installed in dataproc (3.10.8). Everything seems to be working without problems.

ireneisdoomed:
You need to change how we parse the Python version in `install_dependencies.sh` (what we run when we do `make setup-dev`).
It currently parses the TOML, I think we can simply read from `.python-version`. Sth like:
```
echo "Activating Pyenv environment with a Python version required for the project..."
if [ -f ".python-version" ]; then
    PYTHON_VERSION=$(cat .python-version)
else
    echo ".python-version file not found."
    exit 1
fi
```
For consistency and simplicity, I think it is a good idea to lock the version we use for development (defined in `.python-version`) and still support different Python subversions for our users (defined in the TOML).
Let me test this a bit more.

ireneisdoomed (approved):
Thanks for the changes!

PR #407, opened 2024-01-12
test(study_locus): add semantic test for find_overlaps

ireneisdoomed:
To mimic the investigation I did at https://github.com/opentargets/issues/issues/3191 with artificial data.

Summary:
- `StudyLocus.find_overlaps` extracts all variants in the `locus` of 2 associations provided that there is at least one common variant between two loci.
- No overlaps are extracted if there is no common variant.


d0choa (approved):
Semantic tests are always welcome 😄

PR #408, opened 2024-01-12
feat(l2g): limit l2g predictions to gwas-derived associations

ireneisdoomed:
... as discussed with @d0choa and @addramir


src/otg/dataset/l2g_prediction.py
```diff
@@ -62,9 +62,17 @@ def from_credible_set(
         Returns:
             L2GPrediction: L2G dataset
         """
+        gwas_study_locus = StudyLocus(
```
d0choa, line 65: this could be a reasonable StudyLocus method. Most likely we will use it again `study_locus.filterByStudyType(study_index, ["gwas"])`.

It's a refactor so it could go in a different PR

d0choa (approved):
Sounds good. Look at the comment for a potentially useful refactor

PR #411, opened 2024-01-15
chore: update pre-commit hook versions

d0choa:
I ran the hooks against all the files locally without problems.

If there is one to monitor it's prettier. There has been a major update.


ireneisdoomed (approved):
Thank you!

PR #421, opened 2024-01-16
fix: deprecate __version__ in __init__.py

d0choa:
This is falling out of fashion when using modern build systems (e.g. poetry). More info here:
https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module


ireneisdoomed (approved):
Great! Found a [nice ticket in poetry](https://github.com/python-poetry/poetry/issues/1652) about this. With info to extract the version from the package metadata using `importlib.metadata`

PR #433, opened 2024-01-17
docs: several enhancements on docs including index and installation

d0choa:


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "gentropy"
 # !! version is managed by semantic_release
 version = "0.0.0"
-description = "Open targets Genetics Portal Python ETL"
+description = "Open Targets python framework for post-GWAS analysis"
 authors = ["Open Targets core team"]
```
DSuveges, line 6: Is the genetics team part of the "core team"? If not, we should also mention them.

DSuveges (approved):
I had only one quesiton, which is also not tightly related to this PR.

pyproject.toml
d0choa: they are

d0choa:
there is another branch from @addramir. I'll consolidate before merging

PR #436, opened 2024-01-18
chore: improvements to generate 2401 data release

ireneisdoomed:
This PR includes changes in the ETL process to generate all outputs for the data release. It's easier to track changes looking at the commit history, but overall:
- Changes to L2G:
    - Separation between train and test steps in the DAG
    - Inclusion of `vepMeanNeighborhood` and `vepMean` in the features list
    - Revert of https://github.com/opentargets/gentropy/commit/4eafaf297b85067704c327b88d5fdc361c22e14f: this introduced a bug in the generation of coloc features. I have changed the approach.
    - Added specific Spark configuration to allocate driver and executor memory
    - Other bugfixes
- Inclusion of the colocalisation step in the DAG
- Addition of `gene_index` to the data release bucket. I need to write this dataset to extract V2G. Conceptually it's very similar to `target_index` but it's just 3Mb. 
- Infrastructure changes: `n1-highmem-8` was not enough to run the L2G steps, because most of the operations happen in the node. `create_cluster`'s default machine is now `n1-highmem-16`.

--- Added later:

- Added `interaction` dataset to the data release bucket. Similar case as `target_index`. This is an input of L2G to train the model.
- Rearranged the inputs for V2G and L2G so that they all point to files in `static_assets` or in the release bucket


config/datasets/ot_gcp.yaml
```diff
@@ -64,4 +63,5 @@ colocalisation: ${datasets.release_folder}/colocalisation
 study_index: ${datasets.release_folder}/study_index
 variant_index: ${datasets.release_folder}/variant_index
 credible_set: ${datasets.release_folder}/credible_set
+gene_index: ${datasets.release_folder}/gene_index
```
DSuveges, line 66: I still don't understand why the target dataset is in the output. The ETL does nothing with genes: we are not aggregating information for genes, we are not enriching the gene dataset (as far as I know). All gene related information is either in the v2g or variant annotation dataset. Again, if there would be an actual product downstream that would require gene metadata I would understand the placement.

d0choa: I think we are all thinking the same from slightly different angles. No right or wrong. Whatever we do now is likely to change again

ireneisdoomed: It's simply because the gene index is a dependency of V2G.
If it's a big deal, I can delete it after V2G is done.

ireneisdoomed: No. Intervals are a static asset, target is not.

d0choa (approved):

PR #440, opened 2024-01-18
chore: minor updates to gwas catalog pathnames

ireneisdoomed:
Some pathnames are updated to reflect the structure @DSuveges has suggested in [#3193](https://github.com/opentargets/issues/issues/3193)


DSuveges (approved):
Thanks for tidying up! :D

PR #449, opened 2024-01-22
docs: add gentropy first steps guide

ireneisdoomed:
coauthored with @d0choa 

This PR includes changes in the documentation to provide the user with a guide on how to perform basic operations with the package. Inside `How-tos/Python API`, we have covered:
- Creating a Session
- Initialising any Dataset, taking Summary Statistics as an example
- Applying a method on a Dataset, taking window based clumping and fine mapping as an example
- Inspecting and writing the content of a Dataset

This will help the user in their journey with the package.

Next steps is to design a more advanced exercise in a Google Colab notebook, with the advantage that it would be run interactively.


d0choa (approved):
Thank you @ireneisdoomed. I think this will be very valuable for quite a while.

PR #451, opened 2024-01-23
test(docs): testing doc snippets proof of concept

d0choa:
This PR goes against `il-howtos` and presents a proof of concept to test the python snippets in the documentation using pytest.

### DIFF
--- a/tests/docs/test_creating_spark_session.py
+++ b/tests/docs/test_creating_spark_session.py
@@ -1,0 +1,19 @@
+"""Testing creating spark session docs."""
+from gentropy.common.session import Session
+
+from docs.src_snippets.howto.python_api.creating_spark_session import (
+    custom_session,
+    default_session,
+)
+
+
+def test_default_session() -> None:
+    """Test default session."""
+    session = default_session()
+    assert isinstance(session, Session)
+
+
+def test_custom_session() -> None:
+    """Test custom session."""
+    session = custom_session()
+    assert isinstance(session, Session)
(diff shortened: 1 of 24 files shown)


tests/docs/test_creating_spark_session.py
```diff
@@ -0,0 +1,19 @@
...
+)
+
+
+def test_default_session() -> None:
+    """Test default session."""
+    session = default_session()
+    assert isinstance(session, Session)
+
+
+def test_custom_session() -> None:
```
d0choa, line 16: I tried an assert to evaluate that the session is using 4G and it isn't. I think `session` contains the pytest session (probably through `getOrCreate`).

I'm not too concern about this, just saying is not an extremely canonical test, but it should be able to pick if something changes with the Session class.

DSuveges: So the code inside the how-tos are filled from an actual python scripts that are tested. Great.

ireneisdoomed (approved):
This approach entails a considerable effort (externalise the snippets to a separate module, think how to organise them to duplicate as little as possible, and finally test these functions in another module)... but the truth is that despite the simple documentation we had, I have already caught a couple of bugs.
I don't think the tutorials section is going to be very vast, so I think it's worth it and makes the documentation very solid.

DSuveges:
> This approach entails a considerable effort (externalise the snippets to a separate module, think how to organise them to duplicate as little as possible, and finally test these functions in another module)

If we consider providing a examples and how-tos, we should really make sure these documenations won't get obsoleted. For this purpose the above solution works.

PR #453, opened 2024-01-24
feat(LD): include leads with R2=1 in the ldSet when not found in LDIndex

d0choa:
As previously discussed with @ireneisdoomed, @DSuveges, @Daniel-Considine, @addramir, around 15% of the study-locus is not predicted using L2G because the variants are not found in the LD matrix. For these associations, the "PICS-road" will have an empty LdSet which would result in an empty credible set (locus) that would not be exploited by L2G.

This PR implements the logic discussed. When the study-locus is annotated with LD, and assuming the lead variant is available, a ldSet containing one variant (the lead) is created (R^2=1) in cases in which there is no LD annotation available.

This PR would impact all datasets in which we currently perform `clumping` (and downstream data) but it has no structural changes.


ireneisdoomed (approved):
All makes sense, thanks for the changes!
The only thing I'd like to highlight is that with this modification we go from overlooking these credible sets to taking them more significantly than those that do have the lead in the LD index and whose studies represent more than one population. In any case, this is by no means a blocker.

d0choa:
Thanks for the feedback.

On @DSuveges we are indeed dealing with quality checks differently in `StudyLocus` and `StudyIndex`. It all goes down to how liberal we are about allowing the addition of quality checks. Do we want a dictionary of controlled, documented checks or rather prefer to keep a common layer that allows you to add checks to almost any `Dataset`. Let's keep this in mind.

On @ireneisdoomed point, the same lead/tag in multiple populations should also have an R that approximates to 1 if they are available in all populations. The problem arises with how the current formula deals with linked variants that are not available in a fraction of the LD matrixes. Eventually, we need to come back to @addramir's suggestions to deal with this properly.

PR #462, opened 2024-02-01
ci: ignore patch versions in dependabot

d0choa:
This PR only ignores patch releases from Dependabot. (not tested)

I didn't change the time of the week, I'll leave it up to you @ireneisdoomed.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -12,6 +12,9 @@ updates:
     commit-message:
       prefix: "build"
       include: "scope"
+    ignore:
+      - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


ireneisdoomed (approved):
Thank you for the changes! Just one hypothetical question. Imagine two dependencies: A (1.1.0) and B (0.1.0), where B depends on A. A has a minor update to version 1.2.0. Then, dependency B updates its version of A and releases a patch update to 0.1.1. Could we face a compatibility issue because we are updating A and not B?
I don't know to what extent this may bring us problems. If the number of PRs is considerably reduced, I vote for trying this approach until it gives us any problems.

PR #463, opened 2024-02-01
feat: LD index and block matrix extraction for a studyLocus

Daniel-Considine:
Adds get_gnomad_matrix() function to study_locus.py. This function takes a study locus, finds the relevant study ID, extracts the major ancestry, and then returns the ancestry specific hail ld index and block matrices for that window.


addramir:
Is it ready for the review?

src/gentropy/dataset/study_index.py
```diff
@@ -183,3 +183,14 @@ def has_summarystats(self: StudyIndex) -> Column:
...
+    def get_major_pop(self: StudyIndex) -> DataFrame:
+        """Extract major population from ldPopulationStructure rows with multiple ancestries.
+
+        Returns:
+            DataFrame: Columns studyId and the extracted major population from ldPopulationStructure.
+        """
+        return self.df.select("studyId", "ldPopulationStructure").withColumn(
+            "ldPopulationStructure",
+            f.array_max(f.col("ldPopulationStructure")).getItem("ldPopulation"),
+        )
```
DSuveges, line 196: I would rather do something like this:
```suggestion
        return self.df.select(
            "studyId",
            f.array_max(f.col("ldPopulationStructure"))
            .getItem("ldPopulation")
            .alias("majorPopulation"),
        )
```
The name of the returned column should be different from what it was originally called. Also I would not use abbrevation in the function name. (`get_major_pop` -> `get_major_population`)

src/gentropy/dataset/study_locus.py
```diff
@@ -453,6 +457,60 @@ def clump(self: StudyLocus) -> StudyLocus:
         )
         return self
 
+    def get_gnomad_matrix(
```
DSuveges, line 460: I don't think this method conceptually belongs here. There's a very low-level operations (touching block-matrices), that more suitable for the GnomAD data-source. (you can consider how things would change if more possible LD matrices were added not just GnomAD. If this model of implemenation stays, all subsequent method were expected to be in this same module.

src/gentropy/dataset/study_locus.py
```diff
@@ -453,6 +457,60 @@ def clump(self: StudyLocus) -> StudyLocus:
...
+        Returns:
+            tuple[DataFrame, np.ndarray]: Returns the index of the gnomad matrix and the gnomad block matrix
+        """
+        _df = (
+            self.df.join(study_index.get_major_pop(), on="studyId", how="left")
+            .withColumn("start", f.col("position") - (window_size / 2))
+            .withColumn("end", f.col("position") + (window_size / 2))
+            .alias("_df")
+        )
+        _pop = _df.collect()[0]["ldPopulationStructure"]
```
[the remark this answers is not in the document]
Daniel-Considine: When I originally wrote the code to extract the Gnomad LD indices & block matrices for benchmarking, I kept running into issues trying to parallelise the step of extracting the block matrices for each locus. So I thought it was best to get it working for a single study-locus first. I might have have just been doing something wrong though.

src/gentropy/dataset/study_locus.py
Daniel-Considine: Ah ok, originally I thought it might best fit into method/ld.py, but wasn't sure.

src/gentropy/dataset/study_locus.py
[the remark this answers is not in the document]
addramir: Not sure that this is relevant but we disused a lot the these issues (within OTG and with @d0choa ) and concluded the following:
1) All functions related to FM (LD block matrix extraction, Z-score extraction, FM itself) will be applicable to one study-locus at a time. Of course, there is a room for optimisation, but we need to get it working on one study-locus first. The parallelisation is possible only on the level of study-loci (one study locus per cpu).
2) For each study-locus we will assume only ONE major population to extract LD (the most prevalent population), or, in case of absence the information, nfe.

Daniel-Considine:
@DSuveges  I've split the function and added some defaults to make the code clearer/nicer for when I write the wrapper function, and moved them to GnomADLDMatrix as you suggested. Also updated the get_major_population() function as you suggested. 

To extract the square NumPy matrix for a random study locus (5_59546948_C_T) using a 1 mega-base window in non-finish European populations takes 1m 11.4s running on my laptop.

DSuveges:
> To extract the square NumPy matrix for a random study locus (5_59546948_C_T) using a 1 mega-base window in non-finish European populations takes about 40s running on my laptop.

Nice. I'll review this PR together with the Z-score one today.

src/gentropy/datasource/gnomad/ld.py
```diff
@@ -448,3 +449,80 @@ def get_ld_matrix_slice(
...
         )
+
+    @staticmethod
+    def get_locus_index(
+        session: Session,
+        study_locus_row: DataFrame,
+        ld_index_path: str,
+        window_size: int = 1_000_000,
+        major_population: str = "nfe",
+    ) -> DataFrame:
```
DSuveges, line 460: Very-very theorietical problem I have with this function: this logic has nothing GnomAD specific in it. It uses a dataframe (`_matrix_index`) that we lift over from the original ld index. In this sense, I'm not sure if this is the right place for this function.

src/gentropy/datasource/gnomad/ld.py
```diff
@@ -448,3 +449,80 @@ def get_ld_matrix_slice(
...
+                _matrix_index.alias("matrix_index"),
+                (f.col("df.chromosome") == f.col("matrix_index.chromosome"))
+                & (f.col("df.start") <= f.col("matrix_index.position"))
+                & (f.col("df.end") >= f.col("matrix_index.position")),
+            )
+            .select(
+                "matrix_index.chromosome",
+                "matrix_index.position",
+                "referenceAllele",
+                "alternateAllele",
```
DSuveges, line 496: Are these columns coming from `df.` or `matrix_index.`?

src/gentropy/datasource/gnomad/ld.py
```diff
@@ -448,3 +449,80 @@ def get_ld_matrix_slice(
...
+        )
+
+        _index_joined = (
+            _df.alias("df")
+            .join(
+                _matrix_index.alias("matrix_index"),
+                (f.col("df.chromosome") == f.col("matrix_index.chromosome"))
+                & (f.col("df.start") <= f.col("matrix_index.position"))
+                & (f.col("df.end") >= f.col("matrix_index.position")),
+            )
```
DSuveges, line 491: If all the columns for the resulting table can be selected from the matrix_index table, there's no need to join, it can be a filter. Even if the alleles are not in the columns, we can say:

```python
(
    _matrix_index.
    .filter(
        (f.col("chromosome") == "chr") &
        (f.col("df.position") >= position - (window_size / 2)) &
        (f.col("df.position") <= position + (window_size / 2))
    )
    .select(
        "chromosome",
        "position",
        f.split(f.col("variantId"), "_")[2].alias("referenceAllele"),
        f.split(f.col("variantId"), "_")[3].alias("alternateAllele"),
    )
)
```

DSuveges (changes requested):
My only concern is the layout of the locus index retriever function. Which functionality is more or less already in this function: `get_ld_variants` [here](https://github.com/opentargets/gentropy/blame/b9b817d979b9964eb2dbf2702612f264612e8a60/src/gentropy/datasource/gnomad/ld.py#L320C9-L320C24)

DSuveges (approved):
OK, I think all makes sense now. In a separate effort we can update the other function that the ld index generation uses.

PR #470, opened 2024-02-06
feat(dataset): add `Dataset.filter` to create a new dataset instance with filtered data

ireneisdoomed:
The new `Dataset.filter` util was a request of @DSuveges. It is useful to apply filters on the data while maintaining the datatype.

Important to note that this solutions creates different versions (instances) of the dataset. More specifically, as many instances as filters we apply. 

I've added the method to the User Guide docs.


DSuveges (approved):
All makes sense! Thanks @ireneisdoomed

PR #475, opened 2024-02-08
feat(CARMA): adding time limited CARMA

addramir:
This PR adds time restriction on CARMA using 10 min as a default. If it exceeds the computation time it returns empty list of outliers.


tskir (approved):
Everything looks perfectly sensible to me!

PR #478, opened 2024-02-08
feat: get major population function

Daniel-Considine:


Daniel-Considine:
Adds get_locus_sumstats as a staticmethod under the SummaryStatistics class. 
Input is window size, and then either a single studyLocus, or you manually provide it with a chromosome, position, and studyId.  
Running on my laptop takes less than 10 seconds even for Finngen studies.

src/gentropy/dataset/summary_statistics.py
```diff
@@ -114,3 +117,65 @@ def exclude_region(self: SummaryStatistics, region: str) -> SummaryStatistics:
             ),
             _schema=SummaryStatistics.get_schema(),
         )
+
+    @staticmethod
+    def get_locus_sumstats(
+        session: Session,
```
d0choa, line 123: The logic of this function is datasource specific. I believe this logic should live in the `datasource` not in the `dataset`. There are 2 red flags.

- We never pass the `session` to any `Dataset` object. A `dataset` represents the data model and doesn't need to know anything about the data session.
- You need to add conditionals that are datasource specific: `if study is not None and study.startswith("FINNGEN_")`

src/gentropy/dataset/pairwise_ld.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        return np.array(
+            self.df.select(
+                f.split("variantIdI", "_")[1].cast(t.IntegerType()).alias("position_i"),
+                f.split("variantIdJ", "_")[1].cast(t.IntegerType()).alias("position_j"),
+                "r",
+            )
+            .orderBy(f.col("position_i").asc(), f.col("position_j").asc())
+            .select("r")
+            .collect()
```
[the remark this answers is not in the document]
d0choa: Don't worry too much in this PR if you don't want to. We can tackle it later.

src/gentropy/dataset/summary_statistics.py
d0choa: After discussing further with @DSuveges. This function is probably a bit of an overkill. You just need to make sure you have the summary stats object when you want to use it and filter it by chromosome and position.  Basically the majority of the logic should live upstream. If you want, you can generalise the filtering of summary statistics based on chromosome and position.

src/gentropy/datasource/gnomad/ld.py
```diff
@@ -448,3 +451,78 @@ def get_ld_matrix_slice(
...
             )
         )
+
+    @staticmethod
+    def get_locus_index(
+        session: Session,
+        study_locus_row: DataFrame,
+        window_size: int,
+        major_population: str = "nfe",
+        ld_index_path: str = "gs://genetics_etl_python_playground/input/ld/gnomad_r2.1.1.{POP}.common.ld.variant_indices.parquet",
```
d0choa, line 461: please remove the hardcoded path. We want no default here.

src/gentropy/datasource/gnomad/ld.py
```diff
@@ -28,12 +30,14 @@ class GnomADLDMatrix:
...
         ld_index_raw_template (str): Template for the LD index path. Defaults to "gs://gcp-public-data--gnomad/release/2.1.1/ld/gnomad.genomes.r2.1.1.{POP}.common.ld.variant_indices.ht".
+        ld_index_38_template (str): Template for the LD index path in build 38. Defaults to "gs://genetics_etl_python_playground/input/ld/gnomad_r2.1.1.{POP}.common.ld.variant_indices.parquet"
         grch37_to_grch38_chain_path (str): Path to the chain file used to lift over the coordinates. Defaults to "gs://hail-common/references/grch37_to_grch38.over.chain.gz".
         ld_populations (list[str]): List of populations to use to build the LDIndex. Defaults to ["afr", "amr", "asj", "eas", "fin", "nfe", "nwe", "seu"].
     """
 
     ld_matrix_template: str = "gs://gcp-public-data--gnomad/release/2.1.1/ld/gnomad.genomes.r2.1.1.{POP}.common.adj.ld.bm"
     ld_index_raw_template: str = "gs://gcp-public-data--gnomad/release/2.1.1/ld/gnomad.genomes.r2.1.1.{POP}.common.ld.variant_indices.ht"
+    ld_index_38_template: str = "gs://genetics_etl_python_playground/input/ld/gnomad_r2.1.1.{POP}.common.ld.variant_indices.parquet"
```
d0choa, line 40: to remove the hardcoded path

d0choa:
The new dataset PairwiseLD needs to be added to the docs

src/gentropy/dataset/study_index.py
```diff
@@ -183,3 +183,16 @@ def has_summarystats(self: StudyIndex) -> Column:
             Column: True if the study has harmonized summary statistics.
         """
         return self.df.hasSumstats
+
+    def get_major_population(self: StudyIndex) -> DataFrame:
```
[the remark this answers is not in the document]
Daniel-Considine: @addramir What do you think? Probably not a priority for now

src/gentropy/dataset/summary_statistics.py
Daniel-Considine: This makes sense - so instead I should move this code into datasource specific functions? 
datasource/gwas_catalog/summary_statistics.py for example?

src/gentropy/dataset/study_index.py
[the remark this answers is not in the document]
addramir: not a priority now, but we need to tackle the problem later

src/gentropy/dataset/summary_statistics.py
```diff
@@ -103,3 +104,25 @@ def exclude_region(self: SummaryStatistics, region: str) -> SummaryStatistics:
             ),
             _schema=SummaryStatistics.get_schema(),
         )
+
+    def get_locus_sumstats(
+        self: SummaryStatistics,
+        locus: DataFrame,
```
DSuveges, line 110: I don't really like this. Expecting a dataframe is a very vague requirement. It might won't have chromosome and position column in which case the function  fails. I believe it is more robust if we need studyid, chromosome, position. Yes, I think studyid should be mandated, otherwise the returned data might cause surprises.

src/gentropy/dataset/summary_statistics.py
```diff
@@ -103,3 +104,25 @@ def exclude_region(self: SummaryStatistics, region: str) -> SummaryStatistics:
...
+
+        Args:
+            locus (DataFrame): studyLocus DataFrame
+            window (int): Window size
+
+        Returns:
+            DataFrame: Summary statistics for the studyLocus region
+        """
+        chr = locus.select(f.first("chromosome")).collect()[0][0]
+        pos = locus.select(f.first("position")).collect()[0][0]
```
DSuveges, line 123: I don't really like this, so, the input dataframe might have multiple rows, but only the coordinates of the first row is used, ritht? It is not really intuitive and I am afraid also a potential risk for surprises. This would not be an issue if the function would need chr, pos, study.

DSuveges (changes requested):
This function essentially returns a dataframe with a filter applied, the dataframe has identical columns as the source summary statistics. I'm not saying the function was redundant, but have you considered using the `.filter()` method for our datasets?

It would work like this:

```python
(
    # SummaryStatistics object:
    summary_stats
    # Applying filter:
    .filter(
        (f.col('studyId') == study_id) &
        (f.col('chromosome') == chr) &
        (f.col("position") >= pos - (window / 2)) &
        (f.col("position") <= pos + (window / 2))
    )
    # Convert the returned SummaryStatistics object to dataframe:
    .df
)
```

DSuveges (approved):

Daniel-Considine:
No longer needed: 
- Logic for datasource specific handling will be moved upstream
- `order_array_of_structs_by_field` can fulfil purpose of `get_major_population` function
- filtering of sumstats is simple enough to include logic directly into fine mapper

PR #479, opened 2024-02-09
ci: exclude changelog.md from precommit

d0choa:


ireneisdoomed (approved):
... to avoid the problem observed in https://github.com/opentargets/gentropy/pull/477 where the pre-commit hooks were changing the CHANGELOG and these changes weren't part of the committed files.
Thanks!!!

PR #481, opened 2024-02-09
Dc susie finemapper

Daniel-Considine:


src/gentropy/datasource/gnomad/ld.py
```diff
@@ -448,3 +451,98 @@ def get_ld_matrix_slice(
...
+        """Extract the LD block matrix for a locus.
+
+        Args:
+            locus_index (DataFrame): hail matrix variant index table
+            gnomad_ancestry (str): GnomAD major ancestry label eg. `nfe`
+
+        Returns:
+            np.ndarray: LD block matrix for the locus
+        """
+        _locus = locus_index
```
DSuveges, line 537: Why is this needed? Can't we say `locus_index.select("idx").collect()`?

DSuveges (changes requested):
- I think the `notebooks/dev/null` file was added accidentally.
- I would rename this file: `notebooks/notebook.ipynb` if needed to be committed to the repo.

Daniel-Considine:
The logic for building credible sets from the susie outputs need to be changed to make it more robust for outlier cases and for when it hasn't resolved any credible sets.

PR #492, opened 2024-02-18
perf(clump): refactored window based clumping

d0choa:
After different failing attempts to implement the window-based clumping prioritising windows over joins. I had a longer discussion with @DSuveges on rescueing the current implementation based on joins. This seems to be the best solution and it's very conservative with the current implementation. My personal conclusion is that the task of clumping is not well suited for windows, because we don't really care about the vast majority of the summary statistics and windows are good for computing on all rows not a sub-selection of them.

At the end, the only problem with the current implementation was that the broadcast does not work in a "right" join. When the workload increased, the join had to deal with lots of shuffling and eventually failed due to the lack of resources. This PR minimally modifies this part of the window-based clumping to resolve that join.

Using the logic in this PR, clumps on all NFE GWAS Catalog summary statistics were calculated. The [job](https://console.cloud.google.com/dataproc/jobs/f270a358052d427cacf2ff11f4ce6fec/monitoring?region=europe-west1&project=open-targets-genetics-dev) took 18 min to compute and resulted on clumps with collected locus `gs://ot-team/dochoa/sl_11_3_24.parquet` (~25Gb). `spark.sql.shuffle` was set to 3200 for this run.

On top of the business logic, I also modified slightly the interface with the function and the arguments required (default). I think it makes a bit easier to understand the function. These changes are a lot more subjective, so happy to revert if we don't find an agreement.


d0choa:
I ran the window-based clumping collecting the locus in +/- 500_000 in 500 studies. It took 2.1 hours to compute with all parallelisation working properly. The worst chromosome in a busy study (GCST90002374, 40M variants, 1k+ clumps) took  1.6 hours. Clumping a simpler study (GCST006907, 8M variants, 4 clumps) takes seconds. Parallelisation seems to be working properly as long as `spark.sql.shuffle.partitions": str(NUM_STUDIES)`. This ensures that any data shuffle squashes a crazy amount of data in 200 partitions before any autoscaling can kick in.

The algorithm here is resource-intensive. If we conclude this is the way, we should probably implement it so that clumping happens incrementally. A change in the study-locus schema would imply recomputing all the clumps.

docs/src_snippets/howto/python_api/c_applying_methods.py
```diff
@@ -23,7 +23,7 @@ def apply_class_method_clumping(summary_stats: SummaryStatistics) -> StudyLocus:
     from gentropy.method.window_based_clumping import WindowBasedClumping
 
     clumped_summary_statistics = WindowBasedClumping.clump(
-        summary_stats, window_length=500_000
```
DSuveges, line 26: @addramir are we on board with this window size? All data processing uses the default value, changing this would mean changing the distances between the semi indices in the next release.

src/gentropy/dataset/summary_statistics.py
```diff
@@ -59,34 +58,30 @@ def window_based_clumping(
...
-                p_value_significance=gwas_significance,
-            )
+        from gentropy.method.window_based_clumping import WindowBasedClumping
+
+        return WindowBasedClumping.clump(
+            self,
+            distance=distance,
+            gwas_significance=gwas_significance,
+            collect_locus=collect_locus,
+            collect_locus_distance=collect_locus_distance,
```
DSuveges, line 84: I agree there should be one single interface for window based clumping regardless the locus collection.

src/gentropy/method/window_based_clumping.py
```diff
@@ -175,161 +180,128 @@ def clump(
...
+            # Adding study-locus id:
+            .withColumn(
+                "studyLocusId",
+                StudyLocus.assign_study_locus_id(f.col("studyId"), f.col("variantId")),
+            )
+            # Initialize QC column as array of strings:
+            .withColumn("qualityControls", f.array().cast(t.ArrayType(t.StringType())))
+        ).persist()
+
+        if collect_locus:
```
DSuveges, line 247: To my tast the indentation is a bit too much, I would move this logic to a different funciton.

```python
if collect_locus:
    if not collect_locus_distance:
        collect_locus_distance = distance
    locus_df = cls._locus_collection( clumped_dataframe, summary_statistics, collect_distance)
```

But this is mostly stylistic.

src/gentropy/dataset/summary_statistics.py
```diff
@@ -59,34 +58,30 @@ def window_based_clumping(
...
         Returns:
-            StudyLocus: Clumped study-locus containing variants based on window.
+            StudyLocus: Clumped study-locus optionally containing variants based on window.
         """
-        return (
-            WindowBasedClumping.clump_with_locus(
-                self,
-                window_length=distance,
-                p_value_significance=gwas_significance,
-                p_value_baseline=baseline_significance,
```
DSuveges, line 81: I am not sure if it is a good idea to drop `baseline_significance` at this point. OK, we can filter data later, but I assume it is cheaper to do it now than filtering the nested data structure of locus. We can keep the default 1 meaning no drop is done though.

src/gentropy/window_based_clumping.py
```diff
@@ -13,17 +13,21 @@ def __init__(
         session: Session,
         summary_statistics_input_path: str,
         study_locus_output_path: str,
+        distance: int = 500_000,
+        collect_locus: bool = False,
+        collect_locus_distance: int = 500_000,
```
DSuveges, line 18: So it seems this clumping step is flexible to do or not do locus collection.

DSuveges (approved):
I don't see anything that would block the merge, however I have a few minor comments, questions.

src/gentropy/method/window_based_clumping.py
d0choa: This function is a bit too long for my taste as well. I'll add the function

src/gentropy/dataset/summary_statistics.py
d0choa: I was thinking about this. In the [current function](https://github.com/opentargets/gentropy/blob/1095c67a8d9d713615bc8890a5c1926b53815e48/src/gentropy/method/window_based_clumping.py#L182), `baseline_significance` is used to filter the SummaryStatistics at the very top using `summary_stats.pvalue_filter(p_value_significance)`. The `p_value_significance` is no longer used anywhere down the function.

In a sense, this would be the same as if we filter the input summary statistics using `summary_stats.pvalue_filter(p_value_significance)` and pass them to the `clump` function. Everything should work the same. I have a perception that the 2 things are independent and we will increase modularity by keeping them separate, but I am happy to revert if you see a tight coupling between the 2 things.

src/gentropy/window_based_clumping.py
d0choa: Without having more pieces of the puzzle, I struggle to imagine whether we would run the clumping only once or twice (with and without collection).

src/gentropy/method/window_based_clumping.py
d0choa: When you have time have a quick look at the new structure. I have doubts about the `_annotate_locus_using_summary_statistics` (which refers to the same logic you were referring to as `_locus_collection`). Do you think this function belongs to `StudyLocus`? It doesn't look very clumping-specific anymore.

DSuveges: That's a very good point. Because in a sense, if we have the semi indices (which are SL), the locus collection is applied on SL. It also provides some extra flexibilty in terms of allowing a more granular control over which SL you want to collect for finemapping. This would be a great addition.

```python
(
	summary_statistics
	.window_based_clumping() # Returns SL
	.collect_locus(summary_statistics) 
)
```

src/gentropy/dataset/summary_statistics.py
DSuveges: Good point and in a sense it would even work better if the locus collection is a method for StudyLocus

src/gentropy/clump.py
```diff
@@ -27,7 +27,6 @@ def __init__(
         clumped_study_locus_path: str,
         study_index_path: Optional[str] = None,
         ld_index_path: Optional[str] = None,
-        locus_collect_distance: Optional[int] = None,
```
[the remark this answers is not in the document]
d0choa: You were right. The step has been removed.

d0choa:
All the discussions have been implemented as discussed with @DSuveges during the review. I think it's very modular and it should be a good template for what's about to come. @Daniel-Considine you might want to look at the new function `annotate_locus_statistics` in `StudyLocus`.

PR #500, opened 2024-02-20
fix: pre-commit autoupdate message

d0choa:
The default message from `pre-commit.ci` when running `pre-commit autoupdate` does not comply with conventional commits. This is causing all the PRs to fail.

This PR aims to configure `pre-commit.ci` to include conventional messages. Documentation for the configuration is here https://pre-commit.ci/#configuration.

I haven't fully tested this approach. We will need another CI trigger from `pre-commit.ci`


ireneisdoomed (approved):
Thanks!

PR #501, opened 2024-02-20
feat: dataflow decompress prototype

d0choa:
Includes working prototype of DAG step to decompress the eQTL catalog `.gz` datasets using DataFlow's template.

PR #502, opened 2024-02-20
test: increase modularity of test directories

d0choa:
Trying to decouple airflow tests from gentropy tests.


ireneisdoomed (approved):
Some context for this PR changes: in https://github.com/opentargets/gentropy/pull/501 we have a situation where the newly introduced Airflow dependency is incompatible with the dependency tree of Gentropy.
This is only a problem when testing, where we need an environment with all dependencies.
This PR is the first step to modularise the testing environments so that tests for gentropy run independently than the tests for Airflow.
The implication of these changes is that we are excluding `tests/airflow` when we run tests by default. The new rule to run Airflow tests will be implemented in another PR.

PR #507, opened 2024-02-27
chore: small updates to accomodate GWAS Catalog for feb release

DSuveges:
Some details on the modifications:

- [ ] To make sure pandas can write gs:// filesystem, we have to force fsspec version, so been added to as dependency
- [x] GWAS Catalog download files are updated as the expected files are no longer on ftp.
- [x] As the available files on ftp doesn't have `COHORT(S)` column, relevant logic need to be commented. I don't want to completely remove them, as I believe it was not a concious decision on the GWAS Catalog side and this field will likely be back later.


utils/update_GWAS_Catalog_data.sh
ireneisdoomed, line 1: Nice refactor with `fetch_from_ftp`.
Since the script basically places data from FTP into GCS, this could be easily reproduced in Airflow, right? The logic to extract EFO and ENSEMBL versions is only for logging purposes.

ireneisdoomed (approved):
Changes make sense, only that `fsspec` still seems problematic. Good to go once tests pass.

utils/update_GWAS_Catalog_data.sh
DSuveges: Yes, this is something I've been also working on.

PR #510, opened 2024-02-29
fix: variant annotation config

RobinM-code:


d0choa (approved):
Thanks @RobinM-code for the bugfix!
We need to find a way to unit test the configs.
I updated the PR title to comply with conventional commits. Perhaps adding a template for PRs could help with this as well.
If you can bring the PR up-to-date with `dev`, we can merge it.

RobinM-code:
Hi @d0choa,
Thanks for accepting the PR. I have rebased the branch on the `dev` branch.

Also, I think you mention some good suggestions that help in contributing to the code base.
Regarding the Semantic PR check for example: I could see it failed but I was unable to see a error message that explained what was wrong. A PR template would definitely help.

d0choa:
Thanks again!

PR #511, opened 2024-02-29
feat(coloc): single SNP case

xyg123:
no longer filtering out single snp studyLocusOverlaps.

when: f.col("sumlogsum") == f.col("logsum12") == 0 :
 This causes problems when we have to log the result.
 Instead of 0, replace with a very small number.
 
 Works when one causal snp is overlapping, returns high h4 value very close to 1.


addramir:
What is the status here?

xyg123:
Added additional tests for other edge cases where the overlapping SNPs are not causal/ only causal in one study.  The fix returns sensible results in both cases(high h0 or high h1).

This doesn't happen in practice as we take the union of the overlaps between credible sets, in Finngen r10 for example there are 3.4k/39k studyLocusOverlaps which are single SNP, all of which are causal (min.PP = ~0.95, avg.PP=0.99).

src/gentropy/method/colocalisation.py
```diff
@@ -188,12 +188,10 @@ def colocalise(
                 .withColumn("lH2bf", f.log(f.col("priorc2")) + f.col("logsum2"))
                 # h3
                 .withColumn("sumlogsum", f.col("logsum1") + f.col("logsum2"))
-                # exclude null H3/H4s: due to sumlogsum == logsum12
-                .filter(f.col("sumlogsum") != f.col("logsum12"))
                 .withColumn("max", f.greatest("sumlogsum", "logsum12"))
                 .withColumn(
                     "logdiff",
-                    (
+                    f.when(f.col("sumlogsum") == f.col("logsum12"), 1e-10).otherwise(
```
d0choa, line 194: You could specify `PSEUDOCOUNT:float=1e-10` at the beginning of the class.  
```suggestion
                    f.when(f.col("sumlogsum") == f.col("logsum12"), ECaviar.PSEUDOCOUNT).otherwise(
```
The result is exactly the same but it would make easier to understand why the `1e-10` is there for anyone reading this code.

d0choa (approved):
Everything looks healthy. One small comment to make more clear why things are done in this way but nothing critical. Feel free to merge when appropiate

PR #526, opened 2024-03-07
ci: precommit checks for poetry TOML and lock

d0choa:
## ✨ Context

At the moment, nothing ensures that the project TOML is valid and in agreement with the poetry.lock in the repository. It would be great to ensure the sanity of these files, especially in the context of CI/CD and merging different branches with conflicting locks. 

It implements the official [poetry pre-commit hooks](https://python-poetry.org/docs/pre-commit-hooks/) to perform `poetry check` and `poetry lock --no-update` in the files.

I ran `poetry run pre-commit run --all-files` to ensure checks are passing.

**Update**: `poetry lock --no-update` takes a bit of time. The CI throws a timeout, so I had to switch it off in CI. It would still run locally. If we see it slows down commits we will need to switch it off. It's very often skipped so it should not damage the overall experience.

**Update 2**: The pre-commit.ci now supports more modern python versions which tries to use by default. So I had to configure that it keeps using 3.10 

## 🙈 Missing

- I'm still not sure when this is triggered locally. It would be good to see in which scenarios the hook is triggered (not Skipped). Only in TOML changes, TOML and lock?

## 🚦 Before submitting


.pre-commit-config.yaml
```diff
@@ -99,3 +102,10 @@ repos:
     rev: 0.4.1
     hooks:
       - id: pydoclint
+
+  - repo: https://github.com/python-poetry/poetry
+    rev: "1.8.2"
```
d0choa, line 107: I don't think we hardcode the poetry version anywhere. The devcontainer does specify a version but in our local environments, we might all have different versions of poetry with different behaviours. This is not good

ireneisdoomed: I found this long living [thread](https://github.com/python-poetry/poetry/issues/3316) about enforcing Poetry versions, still unsolved.
We could specify it in `install_dependencies.sh`, which is what we trigger when we do `make setup-dev`:
`$ curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.8.2 python3 -`

.pre-commit-config.yaml
```diff
@@ -1,6 +1,9 @@
+default_language_version:
+  python: python3.10
```
ireneisdoomed, line 2: great!

.pre-commit-config.yaml
```diff
@@ -1,6 +1,9 @@
...
 ci:
   autoupdate_commit_msg: "chore: pre-commit autoupdate"
   autofix_commit_msg: "chore: pre-commit auto fixes [...]"
+  skip: [poetry-lock]
```
ireneisdoomed, line 6: Just double checking, this skips running the `poetry-lock` hook, but not the action of generating a new lock file when dependabot upgrades a dependency, right?

ireneisdoomed (approved):
Really nice, thanks again!

.pre-commit-config.yaml
d0choa: correct

.pre-commit-config.yaml
d0choa: The [recommended](https://python-poetry.org/docs/) way of installing poetry these days is using `pipx`.

PR #528, opened 2024-03-07
feat(dag): add data transfer task group for release process

DSuveges:
## ✨ Context

Some updates were made for the next data release + minor improvements in the Airflow layer. Scope defined under [#3238](https://github.com/opentargets/issues/issues/3238).

- Updated configurations for release 24.03
- Targest and interactions are considered as static assests (the airflow service account has no access to platform buckets.)
- The ETL DAG now transfers required datasets from the datasource specific buckets.
- FINNGEN susie ingestion outputs to the canonical location.

## 🚦 Before submitting


src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+    },
+    # GWAS Catalog manifest files:
+    "gwas_catalog_manifests": {
+        "source_bucket": GWAS_CATALOG_BUCKET_NAME,
+        "source_object": "manifests",
+        "destination_bucket": RELEASE_BUCKET_NAME,
+        "destination_object": f"releases/{RELEASE_VERSION}/manifests",
+    },
+    # eQTL Catalog study index:
+    "eqtl_catalog_study_index": {
```
ireneisdoomed, line 58:
```suggestion
    "eqtl_catalogue_study_index": {
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+        "source_bucket": EQTL_BUCKET_NAME,
+        "source_object": "study_index",
+        "destination_bucket": RELEASE_BUCKET_NAME,
+        "destination_object": f"releases/{RELEASE_VERSION}/study_index/eqtl_catalog",
```
ireneisdoomed, line 62:
```suggestion
        "destination_object": f"releases/{RELEASE_VERSION}/study_index/eqtl_catalogue",
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+    },
+    # eQTL Catalog SuSiE credible sets:
+    "eqtl_catalog_susie_credible_set": {
```
ireneisdoomed, line 65:
```suggestion
    "eqtl_catalogue_susie_credible_set": {
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+        "source_bucket": EQTL_BUCKET_NAME,
+        "source_object": "credible_set_datasets/susie",
+        "destination_bucket": RELEASE_BUCKET_NAME,
+        "destination_object": f"releases/{RELEASE_VERSION}/credible_set/eqtl_catalog_susie",
```
ireneisdoomed, line 69:
```suggestion
        "destination_object": f"releases/{RELEASE_VERSION}/credible_set/eqtl_catalogue_susie",
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+    "finngen_susie_credible_set": {
+        "source_bucket": FINNGEN_BUCKET_NAME,
+        "source_object": f"{FINNGEN_RELEASE}/credible_set_datasets/finngen_susie_processed",
+        "destination_bucket": RELEASE_BUCKET_NAME,
+        "destination_object": f"releases/{RELEASE_VERSION}/credible_set/finngen_susie",
+    },
+    # L2G gold standard:
+    "gold_standard": {
+        "source_bucket": "genetics_etl_python_playground",
+        "source_object": "input/l2g/gold_standard/curation.json",
```
ireneisdoomed, line 95: This file is stored in Github, do you think it makes more sense loading it from remote?
Loading remote files is not super straightforward with Spark (SparkFiles solution doesnt work in Dataproc) but we do something similar here https://github.com/opentargets/gentropy/blob/06c611780e3ecb95b7d4b8b40a192dd8b984560b/src/gentropy/datasource/eqtl_catalogue/study_index.py#L46

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+        "destination_bucket": RELEASE_BUCKET_NAME,
+        "destination_object": f"releases/{RELEASE_VERSION}/locus_to_gene_gold_standard.json",
+    },
+}
+
+
+# Test if release folder exists:
+def test_release_folder_exists() -> bool:
```
ireneisdoomed, line 103: This util might be useful as a sensor in other instances.
You could parametrise it and move it to `common_airflow.py`

```python
def test_release_folder_exists(bucket: str, path: str) -> bool:
    """This function tests if the release folder exists.

    Args:
        bucket (str): Name of the bucket in GCS
        path (str): Object name in GCS to check 

    Returns:
        bool: False if the folder exists, True otherwise.
    """
    hook = GCSHook(gcp_conn_id="google_cloud_default")
    return not hook.exists(bucket, path)
```
Minor, but the GCSHook connector is interesting

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+    """This function tests if the release folder exists.
+
+    Returns:
+        bool: False if the release folder exists, True otherwise.
+    """
+    hook = GCSHook(gcp_conn_id="google_cloud_default")
+    return (
+        False
+        if hook.exists(RELEASE_BUCKET_NAME, f"releases/{RELEASE_VERSION}")
```
ireneisdoomed, line 112: You can negate the boolean with `not`
```suggestion
        return not hook.exists(RELEASE_BUCKET_NAME, f"releases/{RELEASE_VERSION}")
```

src/airflow/dags/test_DAG.py
ireneisdoomed, line 103: Since it returns False if it exists, I suggest changing the name to something like 
```suggestion
def test_release_folder_absent() -> bool:
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+            # Chain prerequisites.
+            tasks[step_id] = this_task
+            for prerequisite in step.get("prerequisites", []):
+                this_task.set_upstream(tasks[prerequisite])
+
+        common.generate_dag(cluster_name=CLUSTER_NAME, tasks=list(tasks.values()))
+
+    # DAG description:
+    (
+        # Test if release folder exists:
```
ireneisdoomed, line 164:
```suggestion
        # Test that the release folder doesn't exist:
```

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+    hook = GCSHook(gcp_conn_id="google_cloud_default")
+    return (
+        False
+        if hook.exists(RELEASE_BUCKET_NAME, f"releases/{RELEASE_VERSION}")
+        else True
+    )
+
+
+# This operator meant to fail the DAG if the release folder exists:
+test_folder = ShortCircuitOperator(
```
ireneisdoomed, line 118:
```suggestion
absent_release_folder = ShortCircuitOperator(
```

ireneisdoomed (commented):
Thank you!
Changes here mostly concern adding a group of tasks that place all necessary files in the release bucket prior to the ETL run. Therefore, I'd suggest changing the PR title to sth like `feat(dag): add data transfer task group for release process`
This is the ETL DAG now:
image

I made some comments, let me know your thoughts. Overall, I like the implementation and the idea of introducing the ShortCiruitOperator to safeguard ourselves from rerunning unnecessary processes is very nice.

config/datasets/ot_gcp.yaml
```diff
@@ -36,9 +36,9 @@ anderson: ${datasets.static_assets}/andersson2014/enhancer_tss_associations.bed
 javierre: ${datasets.static_assets}/javierre_2016_preprocessed
 jung: ${datasets.static_assets}/jung2019_pchic_tableS3.csv
 thurman: ${datasets.static_assets}/thurman2012/genomewideCorrs_above0.7_promoterPlusMinus500kb_withGeneNames_32celltypeCategories.bed8.gz
-target_index: ${datasets.release_folder}/targets # OTP 23.12 data
+target_index: ${datasets.static_assets}/targets # OTP 23.12 data
+gene_interactions: ${datasets.static_assets}/interaction # OTP 23.12 data
```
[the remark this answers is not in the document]
DSuveges: No. The service account can't read that bucket.

src/airflow/dags/test_DAG.py
```diff
@@ -0,0 +1,170 @@
...
+CLUSTER_NAME = "otg-etl"
+SOURCE_CONFIG_FILE_PATH = Path(__file__).parent / "configs" / "dag.yaml"
+
+# Release specific variables:
+RELEASE_VERSION = "24.03"
+RELEASE_BUCKET_NAME = "genetics_etl_python_playground"
+
+# Datasource paths:
+GWAS_CATALOG_BUCKET_NAME = "gwas_catalog_data"
+EQTL_BUCKET_NAME = "eqtl_catalog_data"
```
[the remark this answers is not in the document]
DSuveges: Yes, totally makes sense. It must be that the US form was derived from the GWAS Catalog.

src/airflow/dags/test_DAG.py
DSuveges: Fine with me,  I haven't change the logic of this part basically, we did copy from the other bucket. On the other hand, I do see some value to have this file included in the release folder.

src/airflow/dags/test_DAG.py
DSuveges: Agree, it makes sense to generalise.

src/airflow/dags/test_DAG.py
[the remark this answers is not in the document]
d0choa: because of funding as far as I understand

src/airflow/dags/common_airflow.py
```diff
@@ -59,6 +60,24 @@
 }
 
 
+# Test if release folder exists:
```
ireneisdoomed, line 63:
```suggestion
```

ireneisdoomed (approved):
Thank you for addressing the comments!

src/airflow/dags/common_airflow.py
DSuveges: Unfortunately we are still not there yet. (that's why I haven't ask for a re-reivew) For some reason the the gcshook.exists() stopped working. Now I'm trying to find out what is the problem. Keep you posted.

ireneisdoomed: I see, thanks for testing it.

src/airflow/dags/genetics_etl.py
```diff
@@ -0,0 +1,160 @@
...
+    },
+}
+
+
+# This operator meant to fail the DAG if the release folder exists:
+ensure_release_folder_not_exists = ShortCircuitOperator(
+    task_id="test_release_folder_exists",
+    python_callable=lambda bucket, path: not common.check_gcp_folder_exists(
+        bucket, path
+    ),
```
DSuveges, line 106: It might be controversial, but I'm up for defending this decision: I don't like to write separate functions for testing the presence and for testing the absence of something. I like the idea to test something positive, because that's eaiser to understand eg. the tested folder is there -> returns `True`. 

However it implies that if we needs to we have to flip the boolean value. We can either flip the return value or make the function flexible to decide if we want the presence or the absence. I don't like the latter solution. This means if I'm using an airflow operator on a python callable, I need to use lambda. It might not be the nicest, but I think this is the right choice here.

ireneisdoomed: Not controversial to me. Functions are meant to be reused. Thanks!

DSuveges:
> Was the hook failing or it simply stopped working?

Apparently the `GCSHook.exists()` cannot be used to test if a folder exists or not, as there are no such concepts in google cloud explained by [this SO post](https://stackoverflow.com/questions/69257277/how-to-check-if-folder-exists-in-google-cloud-storage-using-airflow-operator/69260058#69260058). Instead I need to list the content of a bucket and evaluate if there are content in the release folder. Not as nice, but kind of make sense.

PR #530, opened 2024-03-08
feat: include COLOC as a node in the DAG

ireneisdoomed:
## ✨ Context

COLOC results: `gs://ot-team/irene/colocalisation_results/ColocalisationMethod.COLOC/`
- 541,773 analyses in 8mins
- ... of which 206,015 are associations with a H4 of at least 0.8 
```
+-------+-------------------+---------------------+                             
|summary|log2h4h3           |h4                   |
+-------+-------------------+---------------------+
|count  |541773             |541773               |
|mean   |0.25526837707702654|0.5526564024644894   |
|stddev |4.255063203793652  |0.3652338709117807   |
|min    |-21.483688919497627|3.2864799866411646E-7|
|25%    |-2.5184409054447268|0.1476165725659369   |
|50%    |0.9719182987936291 |0.6611837628269784   |
|75%    |3.127787709185355  |0.8970387080632414   |
|max    |9.965784284662375  |0.9990009990010025   |
+-------+-------------------+---------------------+
```
These numbers are given for 2,069,316 of credible sets
- ... from which we extract 431,844,590 of overlapping variants

- Logic in the `colocalisation.py` step to compute colocalisations one of the available methods: eCAVIAR or COLOC.
- Split of the colocalisation step in the DAG to generate both results on each run

## 🙈 Missing

It puzzles me that eCAVIAR now takes ~50min to run (2x than in January), when in theory is a more simple method than COLOC and although we are trying to colocalise more associations, I don't expect that the difference in the number of overlapping variants is so significantly higher (SuSIE credible sets ~2M, PICS+SuSIE credible sets 2,8M)

## 🚦 Before submitting

- [X] Did you write any new necessary tests?


d0choa:
The number of gwas-qtl pairs you can make in the 30k finngen credible sets is still very limited. As today, that number is much much bigger in eCaviar.  I’m not saying there is nothing to improve in eCaviar, just that I’m not that surprised about the difference

ireneisdoomed:
> The number of gwas-qtl pairs you can make in the 30k finngen credible sets is still very limited. As today, that number is much much bigger in eCaviar. 
Yes, in my last attempt I saw that the bottleneck was finding the overlaps. Although the step ran succesfully, logging shows many shuffling issues (`org.apache.spark.shuffle.FetchFailedException`)

src/airflow/dags/configs/dag.yaml
d0choa, line 1: remove comments when not needed

src/gentropy/colocalisation.py
```diff
@@ -27,20 +31,52 @@ def __init__(
...
+        colocalisation_results = colocalisation_class.colocalise(overlaps)  # type: ignore
 
         # Load
-        ecaviar_results.df.write.mode(session.write_mode).parquet(coloc_path)
+        colocalisation_results.df.write.mode(session.write_mode).parquet(
+            f"{coloc_path}/{colocalisation_method.value}"
+        )
+
+    @classmethod
+    def _get_colocalisation_class(cls: type[ColocalisationStep], method: str) -> type:
```
d0choa, line 65: I would move this logic to `src/method/colocalisation`. You want a mapping for co-localisation methods. That's generic not only for the step you are working on.

src/gentropy/config.py
```diff
@@ -30,13 +31,21 @@ class StepConfig:
     )
 
 
+class ColocalisationMethod(Enum):
```
d0choa, line 34: You should be asking the Colocalisation method module what's available and create an enum based on that. No need for a class.

src/gentropy/colocalisation.py
```diff
@@ -27,20 +31,52 @@ def __init__(
             credible_set_path (str): Input credible sets path.
             study_index_path (str): Input study index path.
             coloc_path (str): Output Colocalisation path.
+            colocalisation_method (ColocalisationMethod): Colocalisation method. Available methods are: ECAVIAR, COLOC.
         """
         # Extract
-        credible_set = StudyLocus.from_parquet(
-            session, credible_set_path, recursiveFileLookup=True
+        credible_set = (
+            StudyLocus.from_parquet(
```
d0choa, line 38: It works for now but this paragraph is problematic. "Susie" and "coloc" are hardcoded.

d0choa (commented):
There are some structural things that I think are important here.

I feel there is a need for a class `ColocalisationMethods` in `src/method/colocalisation.py`. This class can be an index used to map the names of the methods, the classes implementing them, and some helper methods that allow you to list what's available and how to run them. Perhaps, also what are the requirements (e.g. logBF/Susie for COLOC). You could import this class to populate the CLI config or in the step. I think this will tidy up quite a bit this code and minimise hardcoding.

src/gentropy/colocalisation.py
ireneisdoomed: "coloc" is no longer hardcoded.
I don't see how i can not hardcode "SuSie" without plaguing the code with parameters.

src/gentropy/colocalisation.py
```diff
@@ -1,10 +1,13 @@
 """Step to generate colocalisation results."""
 from __future__ import annotations
 
+from pyspark.sql.functions import col
+
 from gentropy.common.session import Session
+from gentropy.config import ColocalisationMethod
```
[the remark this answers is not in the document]
ireneisdoomed: Done.

src/gentropy/colocalisation.py
ireneisdoomed: Can you tell me a scenario where you'd need a mapping when you are interacting directly with the classes?
I thought it made sense in the context of the step, where you want to abstract the logic from the used method.

ireneisdoomed:
@d0choa Thank you for the review. I have refactored the part of the method validation. I still think it only makes sense in the context of the step, but happy to know your comments otherwise.

src/airflow/dags/configs/dag.yaml
```diff
@@ -1,18 +1,21 @@
...
+#     - "ot_variant_to_gene"
+#     - "ot_colocalisation_ecaviar"
+#     - "ot_colocalisation_coloc"
+# - id: "ot_locus_to_gene_predict"
+#   prerequisites:
+#     - "ot_locus_to_gene_train"
+#     - "ot_variant_index"
+#     - "ot_variant_to_gene"
+#     - "ot_colocalisation_ecaviar"
+#     - "ot_colocalisation_coloc"
```
DSuveges, line 21: Are these lines accidentally left commented out?

ireneisdoomed: Indeed, thanks!

src/gentropy/colocalisation.py
DSuveges: I think what @d0choa meant is that the applied colocalization method should know what requirements the data should accomodate not the step. Eg. each coloc methods would have a set of criteria that would apply on the dataset, in that sense it would be the method's reponsability to enforce that the data is right.

ireneisdoomed: Apologies I had missed this comment.
I think the responsibility is in the dependency injection step, not in the method.
In any case, `coloc` is already capturing that scenario because it's a method that **requires** having logBFs. If none of the credible sets have them, the data will pass through and the output coloc metrics will be 0 (this is theoretically true, I just found a bug).

Also, considering that the input of the colocalisation function is an Overlaps object, I think it makes sense to add the filter to just include SuSie credible sets at the step level because we won't extract overlaps for credible sets for which we can't get coloc results.

ireneisdoomed:
Let me know if there's anything else to be done here.

ireneisdoomed:
Revising these comments, even though it is not a dealbreaker, I can see how the process could be improved to have a cleaner step.
We can leave these changes out of this release so I can improve it, I'm fine with either.

PR #541, opened 2024-03-14
feat: notebook to run qc metrics each release

xyg123:
## ✨ Context

Need to get top level QC metrics each release.

Jupyter notebook addition to compute simple QC metrics for each release, to be run locally or on a VM with the full release dataset.

## 🙈 Missing

QC metrics that cross compares results from across releases are interesting, but not implemented yet.

## 🚦 Before submitting

- [x] Did you read the [contributor guideline](htxtps://opentargets.github.io/gentropy/development/contributing/#contributing-checklist)?


addramir:
Thank you, looks good!
1) The size of variant index is slightly different from what was reported by Daniel (5090991 vs 5468737). Any ideas why? 
2) There is None in "Average number of overlaps per CS:  None". 
3) PICS vs Susie Finngen has a lot fo snps with pics_pip=1. No actions needed, just to note and to discuss later.
4) Could you please add the correlation from  PICS vs Susie Finngen pips?
5) The number in "Of these,  122485  studyloci contains more than one gene with score > 0.5" looks very high. Is it possible to somehow calculate this number for pics and susie study-loci separately? It is high probably because the study_id is the same for finngen_pics and finngen_susie.

xyg123:
> 1. The size of variant index is slightly different from what was reported by Daniel (5090991 vs 5468737). Any ideas why?

There must be some sort of filter on the variant index, we see the same number of unique variants in the v2g table (5090991).  It's not rsId/alleletype based filter, so maybe a MAF filter?

> 2. There is None in "Average number of overlaps per CS:  None".
Fixed

> 3. PICS vs Susie Finngen has a lot fo snps with pics_pip=1. No actions needed, just to note and to discuss later.
Could be differences in clumping vs. not clumping.

> 4. Could you please add the correlation from  PICS vs Susie Finngen pips?
Added. (0.624)
> 5. Please add your comment 
"There are definitely duplicated studylocusIDs in the l2g predictions, and still around 20% of studylocus contains more than 1 gene with score>0.5 . It is not possible to separate out these predictions based on whether they came from pics or susie,  as the l2g outputs only contains the studylocusID (duplicated between pics and susie). If 20% is too high then it implies finngen pics and susie l2g are pointing (confidently) at different genes for the same studylocus." after you calculated it.

Added.

addramir (approved):

PR #543, opened 2024-03-15
fix: minor updates and bug fixes

DSuveges:
## ✨ Context

Upon QCing the 24.03 data release a few issues surfaced. This pr meant to address them.

## 🛠 What does this PR implement

- [x] Update the configuration of the variant index step to point to the right credible set dataset.
- [x] Renaming the credible set dataset generated by the GWAS Catalog pre-process step to make their content more obvious.
- [x] Applying credible interval filter on the PICSed loci to ensure only the 99% credible sets remain.

## 🚦 Before submitting

- [ ] Do these changes cover one single feature (one change at a time)?
- [ ] Did you read the [contributor guideline](https://opentargets.github.io/gentropy/development/contributing/#contributing-checklist)?
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you make sure there is no commented out code in this PR?
- [ ] Did you follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standards in PR title and commit messages?
- [ ] Did you make sure the branch is up-to-date with the `dev` branch?
- [ ] Did you write any new necessary tests?
- [ ] Did you make sure the changes pass local tests (`make test`)?
- [ ] Did you make sure the changes pass pre-commit rules (e.g `poetry run pre-commit run --all-files`)?

PR #544, opened 2024-03-15
perf(l2g): streamline feature generation

ireneisdoomed:
## ✨ Context
The increase of V2G relationships (https://github.com/opentargets/gentropy/pull/543) caused L2G to crash.
The partitioning strategy was not optimised to extract that many feature annotations, which meant that executors had to deal with big chunks of data and would eventually die.

This job is heavy in the number of aggregation tasks it has to do, so having the data properly partitioned is important.

- Lighter joins in the extraction of VEP features (https://github.com/opentargets/gentropy/pull/544/commits/c31786b85e5189f2e120707b00eb629ac0ca31a2). In `credible_set_w_variant_consequences`, I am selecting the data to be shuffled in the join.
- Lock the default number of returned partitions after a join to 800 (https://github.com/opentargets/gentropy/pull/544/commits/6340e8919a7d9bbaf3822a783b176a4793aac5fa). In the current solution, I saw that the eCAVIAr features set had 980 partitions, the distance features 200 and the VEP features 17.
- Other minor refinements

## 🙈 Missing

In an upcoming PR I'll address the biggest gain i've identified:
- Minimise for loops. If look at `_get_max_coloc_per_credible_set`, we run this function twice:
    - Once to extract COLOC, another time to extract eCAVIAR. However, we could apply the logic using the method as a grouping field so that we don't split the logic into 2 threads. The complexity here is that this is implemented this way because the name of the feature is based on the method, so we need to take that into account.
    - Very very similarly, we iterate over each type of QTL to perform the aggregation. 

## 🚦 Before submitting


tests/gentropy/conftest.py
```diff
@@ -84,6 +84,11 @@ def mock_colocalisation(spark: SparkSession) -> Colocalisation:
         .withColumnSpec("h4", percentNulls=0.1)
         .withColumnSpec("log2h4h3", percentNulls=0.1)
         .withColumnSpec("clpp", percentNulls=0.1)
+        .withColumnSpec(
+            "colocalisationMethod",
+            percentNulls=0.0,
```
DSuveges, line 89: I don't know too much about the details, but going down from 0.1 to 0.0 is expected?

ireneisdoomed: `colocalisationMethod` contains the method we used to get the result. It should never be null.

PR #547, opened 2024-03-15
chore(airflow): remove hardcoded value for gentropy version

ireneisdoomed:
## ✨ Context

We had to keep the GENTROPY_VERSION value in sync with the code version to run DAGs.

Apparently setting a specific variable in the package to specify the version is controversial because, although it might be handy like for this case, you have to maintain it and people complain that it lacks standardisation (`__version__` vs `VERSION` vs `package.get_version()`). There are some guides [now](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/), but given the small use case I think this is the least compromised solution.

This is based on this SO [thread](https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module/72168209#72168209). `importlib` doesn't access the info in the TOML, it just reads the version variable stored in a the package metadata.

## 🚦 Before submitting

- [X] Do these changes cover one single feature (one change at a time)?

### DIFF
--- a/src/airflow/dags/common_airflow.py
+++ b/src/airflow/dags/common_airflow.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from importlib.metadata import version as importlib_version
 from typing import TYPE_CHECKING, Any, Optional
 
 import pendulum
@@ -18,8 +19,7 @@ from google.cloud import dataproc_v1, storage
 if TYPE_CHECKING:
     from pathlib import Path
 
-# Code version. It has to be repeated here as well as in `pyproject.toml`, because Airflow isn't able to look at files outside of its `dags/` directory.
-GENTROPY_VERSION = "0.0.0"
+GENTROPY_VERSION = importlib_version("gentropy")
 
 # Cloud configuration.
 GCP_PROJECT = "open-targets-genetics-dev"


d0choa:
Sorry I don't get this. You use your local gentropy version to inform which gentropy version should run in dataproc?

DSuveges:
I feel the resolution of this issue should be done as one single step to move every configuration of the pipeline to the airflow layer. Essentially the gentropy version only needed to which wheel airflow to use to install on dataproc, we can just point to a location. I haven't explored the options we could use.

ireneisdoomed:
@d0choa Yes, I was just trying to remove the codependence between the version we point to in Airflow and the version in the TOML.

Local version (TOML) ---> Package wheel in GCS derived from version ---> Package installation in cluster

When I run stuff I generally change the version in the TOML to avoid overlapping, which makes me have to change the value in Airflow.
@DSuveges is on point though, this is just a minor improvement to the current setup.I agree that we should totally move the configuration to the Airflow layer. The wheel location in GCS could be set dynamically with a solution like this one, or manually by tweaking the config. Either is fine for me, so feel free to close it.

d0choa:
Because we have reservations about this PR and it will only partially address the issue, we agreed to close it for now.

Configuration at the airflow layer is next in the TODO list to tackle this problem

PR #553, opened 2024-03-20
perf(l2g): optimise extraction of features from colocalisation results

ireneisdoomed:
## ✨ Context
Coloc based features happens in Spark natively. I have removed looping through each colocalisation method and each QTL type.

This was mentioned in this PR https://github.com/opentargets/gentropy/pull/544

- Addition of colocalisation method to the windowing functions
- Overhaul of `Colocalisation._get_max_coloc_per_credible_set` to directly extract the features in long format
- Adapted dependencies

## 🙈 Missing

Run the step to evaluate performance gain.

## 🚦 Before submitting


ireneisdoomed:
Performance of the prediction step reduced in ~20%:
- 11 min 18 sec ([job](https://console.cloud.google.com/dataproc/jobs/cdadd5da-5832-49ff-9d21-9051496fb016/configuration?region=europe-west1&project=open-targets-genetics-dev))
- to 9min 21sec ([job](https://console.cloud.google.com/dataproc/jobs/a66fe79f-35e2-442e-86f1-8019b817c8ac/monitoring?region=europe-west1&project=open-targets-genetics-dev))

This difference will be bigger when we process COLOC results, which we weren't doing so far.

src/gentropy/method/l2g/feature_factory.py
```diff
@@ -26,162 +26,121 @@ class ColocalisationFactory:
...
-            coloc_feature_col_template = "ColocLlrMaximum"
-
-        elif colocalisation_method == "eCAVIAR":
-            coloc_score_col_name = "clpp"
-            coloc_feature_col_template = "ColocClppMaximum"
+        colocalisation_df = colocalisation.df.select(
+            f.col("leftStudyLocusId").alias("studyLocusId"),
+            "rightStudyLocusId",
+            f.coalesce("log2h4h3", "clpp").alias("score"),
+            f.when(f.col("colocalisationMethod") == "COLOC", f.lit("Llr"))
```
d0choa, line 47: to avoid the hardcoding you might want to do something like:

```suggestion
            f.when(f.col("colocalisationMethod") == Coloc.METHOD_NAME, Coloc.STATISTIC_COLUMN)
```
These 2 attributes would belong to the `Coloc` method class. You would also use `Coloc.METHOD_NAME` to populate the original column `colocalisationMethod`. This would also reduce the risk of introducing errors as the logic will no longer depend on the semantics of the column (e.g. "COLOC").
 
Because these are properties of the method it would make sense. An abstract class for colocalisation methods might be overkill for only 2 methods and no prospect of expanding them, but I would be also ok with that. You could also build logic that iterates through all classes and build this when statements. Again, it could be an overkill in this context but it might be something you want to think about in other contexts (e.g. features, etc.).

src/gentropy/method/l2g/feature_factory.py
```diff
@@ -26,162 +26,121 @@ class ColocalisationFactory:
...
+            .when(f.col("colocalisationMethod") == "eCAVIAR", f.lit("Clpp"))
```
d0choa, line 48: of course, same here

d0choa (approved):
I will accept anything that removes a `.toPandas()` statement ;)

Look at my suggestion. It's a small change but think about it in the context of other things you are working with.

src/gentropy/method/l2g/feature_factory.py
d0choa: This is good to go.

Thinking about further improvements. There are probably 2 things that come to mind:
- Ensuring that all methods have METHOD_NAME, etc. using inheritance. Not urgent as we are not adding more methods short term.
- One class/function responsible for telling you the list of available colocalisation classes/methods. Inheritance of an abstract colocalisationMethod might help here as well. [Check this](https://stackoverflow.com/questions/3862310/how-to-find-all-the-subclasses-of-a-class-given-its-name). Then you could build your dictionary dynamically. On a separate note, I think we will need to add a similar strategy if we want to play with the features to include in L2G upon configuration.

ireneisdoomed: Thank you for the comments! I did a very similar [here](https://github.com/opentargets/gentropy/commit/dfaac7994760f0b76bb0a4fcfd494acd415e72f4) so I agree that we should implement this properly

PR #554, opened 2024-03-21
feat: add the step class for fine-mapping

addramir:
## ✨ Context
This PR introduced the fine-mapper step and has a function that converts susie output to study-locus.

See above.

## 🙈 Missing
Documentation will be updated later. 

## 🚦 Before submitting


src/gentropy/susie_finemapper.py
```diff
@@ -0,0 +1,140 @@
...
+            cred_set = filtered_arr[:, [0, i + 1, i + L_snps + 1, i + 2 * L_snps + 1]]
+            win = Window.rowsBetween(
+                Window.unboundedPreceding, Window.unboundedFollowing
+            )
+            cred_set = (
+                session.spark.createDataFrame(
+                    cred_set.tolist(),
+                    ["variantId", "posteriorProbability", "logBF", "beta"],
+                )
+                .join(
```
DSuveges, line 82: At this join the mode is `inner` by default. Is it expected?

src/gentropy/susie_finemapper.py
```diff
@@ -0,0 +1,140 @@
...
+        cs_lbf_thr: float = 2,
+    ) -> StudyLocus:
+        """Convert SuSiE-inf output to studyLocus DataFrame.
+
+        Args:
+            susie_output (dict[str, Any]): SuSiE-inf output dictionary
+            session (Session): Spark session
+            _studyId (str): study ID
+            _region (str): region
+            _join (DataFrame): DataFrame with variant information
```
DSuveges, line 38: I would rename the `_join` variable to something more intuitive.

src/gentropy/susie_finemapper.py
```diff
@@ -0,0 +1,140 @@
...
+        susie_result = np.hstack((variants, PIPs, lbfs, mu))
+
+        L_snps = PIPs.shape[1]
+
+        # Extracting credible sets
+        order_creds = list(enumerate(susie_output["lbf"]))
+        order_creds.sort(key=lambda x: x[1], reverse=True)
+        cred_sets = None
+        counter = 0
+        for i, value in order_creds:
```
DSuveges, line 59: `i` is fine, this is a scanonical way of calling an index variable, however `value` is not very telling. At row 110 it is very complicated what does this `value`refer to.

src/gentropy/susie_finemapper.py
```diff
@@ -0,0 +1,140 @@
...
+        _join: DataFrame,
+        cs_lbf_thr: float = 2,
+    ) -> StudyLocus:
+        """Convert SuSiE-inf output to studyLocus DataFrame.
+
+        Args:
+            susie_output (dict[str, Any]): SuSiE-inf output dictionary
+            session (Session): Spark session
+            _studyId (str): study ID
+            _region (str): region
```
DSuveges, line 37: Do we have a consisted way of representing regions? If so, in the args description could be written eg.:
```text
_region (str): finemapped region in chr:start-end format
```

src/gentropy/susie_finemapper.py
```diff
@@ -0,0 +1,140 @@
...
+                    "chromosome",
+                    "position",
+                    "finemappingMethod",
+                    "credibleSetlog10BF",
+                )
+            )
+            if cred_sets is None:
+                cred_sets = cred_set
+            else:
+                cred_sets = cred_sets.union(cred_set)
```
DSuveges, line 136: I would use `unionByName` rather `union`, because union concatenates columns by positions instead of names. I think in this case it is fine, as the order of columns are defined in rwo 120, but still.

tests/gentropy/method/test_susie_inf.py
```diff
@@ -48,3 +52,28 @@ def test_SUSIE_inf_cred(
...
+        gwas_df = sample_summary_statistics._df.limit(21)
+        L1 = SusieFineMapperStep.susie_inf_to_studylocus(
+            susie_output=susie_output,
+            session=session,
+            _studyId="sample_id",
+            _region="sample_region",
+            _join=gwas_df,
+            cs_lbf_thr=2,
+        )
+        assert isinstance(L1, StudyLocus), "L1 is not an instance of StudyLocus"
```
DSuveges, line 79: I'm not sure how does the test dataset look like, it would be great to assert that the number of credible set is what you are expecting, and validate if the `locus` object is healthy. However I understand if that is not a high priority for now.

DSuveges (approved):
I only have a few small comments that doesn't affect the logic, so not a deal breaker for merging. My comments are stylisting, if you want to address them, you can before merge.

tests/gentropy/method/test_susie_inf.py
addramir: I think we can create more meaningful test for bigger function that will use this convertor on later stages.

src/gentropy/susie_finemapper.py
addramir: No, we don't have it now. But agree, we need to think about the standard.

src/gentropy/susie_finemapper.py
addramir: renamed it for cs_lbf_value

src/gentropy/susie_finemapper.py
addramir: Done

src/gentropy/susie_finemapper.py
addramir: Yes, it should be 1 to 1 the same size and order

PR #581, opened 2024-04-23
feat(airflow): prototype finemapping batch job

ireneisdoomed:
New Airflow DAG that calls the susie finemapper step on a list of studyLocus IDs to generate credible sets.

## ✨ Context
This is an incremental pipeline, so the first 3 tasks involve the operation of generating a list of studyLocus to finemap. Once the difference is computed, we finemap those IDs as a job to Google Batch.

image

Graph nodes:
- `get_all_study_locus_ids`. Lists the bucket with the clumped study locus IDs and extracts all IDs that we could theoretically finemap based on the filename. **A requirement for this step is that clumped study loci are written partitioned by their ID - which is currently not implemented** (e.g. `gs://genetics-portal-dev-analysis/irene/toy_studdy_locus_alzheimer_partitioned`)
- `get_finemapped_paths`. Lists the bucket that contains all credible sets as a result of fine-mapping and extracts the IDs from the filename. **Similarly, it is also required that the SLID is part of the filename.** This is partially implemented, already. The data is not partitioned, but [the ID is used to build the output path](https://github.com/opentargets/gentropy/blob/70e5e2615700518324d414d11165efe025798284/src/gentropy/susie_finemapper.py#L72).
- `get_study_loci_to_finemap`. Creates a list of IDs based on the difference between get_all_study_locus_ids and get_finemapped_paths.
- `finemapping_task`. The one that interfaces with the Google Batch operator to create **one Batch job that runs the Docker container with as many tasks in parallel as studyLoci IDs we have extracted** in get_study_loci_to_finemap. The  command run in the container image calls the fine mapping step with the appropriate parameters.

- The DAG explained above.
- A small change in the logic of `SusieFineMapperStep` that instead of passing the row of the studyLocus to finemap, it passes the dataframe (containing that one row). This is to prevent incompatibilities between the input data, and the schema (I had the problem that after partitioning the data by SLID and then reading the data, this column was appended at the end)

## 🙈 Missing

- Testing that the logic works after fixing the schema issues. This requires creating a new image.
- Fine tuning the resources allocated in the Batch job. This crosses over with the work done by @tskir 
- Sorting out the input data for this DAG:
    - First, we need to partitioning the data prior to the finemapping step as required by the `get_all_study_locus_ids` node. This involves changing `ld_based_clumping.py`
    - Then, we have to be mindful that we want to fine map studyLoci from 2 sources: UKBB PPP and GWAS Catalog. So either we run the DAG twice or we put all clumped study loci under the same location.

## 🚦 Before submitting

- [X] Did you read the [contributor guideline](https://opentargets.github.io/gentropy/development/contributing/#contributing-checklist)?

### DIFF
--- a/src/airflow/dags/susie_finemapping.py
+++ b/src/airflow/dags/susie_finemapping.py
@@ -1,0 +1,89 @@
+"""Example Airflow DAG that uses Google Cloud Batch Operators."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import common_airflow as common
+from airflow.models.dag import DAG
+from airflow.providers.google.cloud.operators.cloud_batch import (
+    CloudBatchSubmitJobOperator,
+)
+from google.cloud import batch_v1
+
+PROJECT_ID = "open-targets-genetics-dev"
+REGION = "europe-west1"
+
+
+def _finemapping_batch_job(studyloci_ids: list[str]) -> batch_v1.Job:
+    """Create a Batch job to run fine-mapping on a list of study loci.
+
+    Args:
+        studyloci_ids (list[str]): A list of study loci IDs to run fine-mapping on.
+
+    Returns:
+        batch_v1.Job: A Batch job to run fine-mapping on the given study loci.
+    """
+    runnable = batch_v1.Runnable()
+    runnable.container = batch_v1.Runnable.Container()
+    runnable.container.image_uri = "europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app-dev/gentropy_test:0.1"
+    runnable.container.entrypoint = "/bin/sh"
+    runnable.container.commands = [
+        "-c",
+        "echo We are going to finemap this studyLocusId $STUDYLOCUSID",
+    ]
+
+    task = batch_v1.TaskSpec()
+    task.runnables = [runnable]
+
+    resources = batch_v1.ComputeResource()
+    resources.cpu_milli = 2000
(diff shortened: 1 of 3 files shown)


tskir (commented):
I've rebased this PR against dev and resolved some minor merge conflicts to prevent it from getting too much out of sync.

I think this is a great idea overall; however, I find this PR a bit difficult to handle as it is because it introduces three changes simultaneously:
1. Introducing Google Batch with Docker;
2. Making the pipeline incremental;
3. Modifying the internal logic of SuSie finemapping.

And further (this could be considered point # 4) to make this pipeline work, we would first need to make additional changes into how clumped outputs are produced.

So again, while I agree in principle with all of the changes proposed, at this point I will not be progressing this PR in its current form. Rather, as the first atomic change, I will use the prototype code provided to implement the existing finemapping approach (without any changes in logic) using Docker in the Airflow DAG.

d0choa:
@tskir feel free to close the PR when it's not useful anymore. The intention from @ireneisdoomed and me was to create a proof-of-concept not to merge these changes

tskir:
The finemapping DAG has been merged in https://github.com/opentargets/orchestration/pull/10, so I'm now closing this pull request.

Many ideas from it have been very helpful and have been used in my final PR — thank you @ireneisdoomed @d0choa!

tskir:
@ireneisdoomed I didn't delete the branch in case there's something useful you planned to use there — please delete if not!

PR #594, opened 2024-04-30
feat: check for if no overlapping variants in LD index

Daniel-Considine:
Checks if the sumstat - LD index join is empty, and if so logs warning and terminates run. 

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/susie_finemapper.py
+++ b/src/gentropy/susie_finemapper.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 import time
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


src/gentropy/susie_finemapper.py
```diff
@@ -142,9 +143,12 @@ def __init__(
...
             # Write result
-            result.df.write.mode(session.write_mode).parquet(
-                output_path + "/" + study_locus_to_finemap
-            )
+            if result is not None:
+                result.df.write.mode(session.write_mode).parquet(
+                    output_path + "/" + study_locus_to_finemap
+                )
+            else:
+                return
```
addramir, line 151: Are you sure that it should be just empty return? Not "return None"? Or even without "else"?

Daniel-Considine: I think `return` and `return None` are equivalent. You're right that the else statement isn't required - I thought it might throw an error with mypy.

addramir (approved):

PR #628, opened 2024-05-30
feat(config): extract gwas_significance parameter to step configuration

project-defiant:
## ✨ Context

https://github.com/opentargets/issues/issues/3327

This is the first part of the #3327. 
* Extraction of the `gwas_significance` (p-value) threshold parameter for the window-based-clumping step.
* Addition of `gwas_significance` to the airflow config with the new value (`1e-8`) overwriting default one (`5e-8`)
* Unified naming convention for the `WindowBasedClumpingStep` config to `WindowBasedClumpingStepConfig`

The main reason for this change is that we want to use the value of `1e-8` and possibly even more stringent value of `1e-9` in the future. 
## 🚦 Before submitting

### DIFF
--- a/config/step/ot_window_based_clumping.yaml
+++ b/config/step/ot_window_based_clumping.yaml
@@ -4,3 +4,4 @@ defaults:
 summary_statistics_input_path: ???
 study_locus_output_path: ???
 inclusion_list_path: ???
+gwas_significance: 1e-8
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -355,11 +355,17 @@ class VariantToGeneConfig(StepConfig):
 
 
 @dataclass
-class WindowBasedClumpingStep(StepConfig):
+class WindowBasedClumpingStepConfig(StepConfig):
     """Window-based clumping step configuration."""
 
+    session: Any = field(
+        default_factory=lambda: {
+            "start_hail": True,
+        }
+    )
     summary_statistics_input_path: str = MISSING
     study_locus_output_path: str = MISSING
+    gwas_significance: float = 5e-8
     distance: int = 500_000
     collect_locus: bool = False
     collect_locus_distance: int = 500_000
@@ -454,5 +460,7 @@ def register_config() -> None:
     cs.store(group="step", name="variant_annotation", node=VariantAnnotationConfig)
     cs.store(group="step", name="variant_index", node=VariantIndexConfig)
     cs.store(group="step", name="variant_to_gene", node=VariantToGeneConfig)
-    cs.store(group="step", name="window_based_clumping", node=WindowBasedClumpingStep)
+    cs.store(
+        group="step", name="window_based_clumping", node=WindowBasedClumpingStepConfig
+    )
     cs.store(group="step", name="susie_finemapping", node=FinemapperConfig)
--- a/src/gentropy/method/window_based_clumping.py
+++ b/src/gentropy/method/window_based_clumping.py
@@ -155,14 +155,14 @@ class WindowBasedClumping:
     def clump(
         summary_statistics: SummaryStatistics,
         distance: int = 500_000,
-        gwas_significance: float = 5e-8,
+        gwas_significance: float = 1e-8,
     ) -> StudyLocus:
         """Clump significant signals from summary statistics based on window.
 
         Args:
             summary_statistics (SummaryStatistics): Summary statistics to be used for clumping.
             distance (int): Distance in base pairs to be used for clumping. Defaults to 500_000.
-            gwas_significance (float): GWAS significance threshold. Defaults to 5e-8.
+            gwas_significance (float): GWAS significance threshold. Defaults to 1e-8.
 
         Returns:
             StudyLocus: clumped summary statistics (without locus collection)
--- a/src/gentropy/method/window_based_clumping.py
+++ b/src/gentropy/method/window_based_clumping.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING
 
 import numpy as np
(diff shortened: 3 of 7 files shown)


src/gentropy/config.py
```diff
@@ -355,11 +355,17 @@ class VariantToGeneConfig(StepConfig):
...
     """Window-based clumping step configuration."""
 
+    session: Any = field(
+        default_factory=lambda: {
+            "start_hail": True,
+        }
+    )
     summary_statistics_input_path: str = MISSING
     study_locus_output_path: str = MISSING
+    gwas_significance: float = 5e-8
```
addramir, line 368: Shouldn't it be 1e-8?

config/step/ot_window_based_clumping.yaml
```diff
@@ -4,3 +4,4 @@ defaults:
 summary_statistics_input_path: ???
 study_locus_output_path: ???
 inclusion_list_path: ???
+gwas_significance: 1e-8
```
addramir, line 7: What is the difference between configs here and WindowBasedClumpingStepConfig?

src/gentropy/method/window_based_clumping.py
```diff
@@ -166,7 +168,16 @@ def clump(
 
         Returns:
             StudyLocus: clumped summary statistics (without locus collection)
+
+            Check WindowBasedClumpingStepConfig object for default values
         """
+        # p-value default value will change in the next major release
+        if gwas_significance == 5e-8:
+            warnings.warn(
+                "Default value for gwas_significance will change in 2.0 release",
```
addramir, line 177: "Default value for gwas_significance will change in next release"

src/gentropy/window_based_clumping.py
```diff
@@ -26,9 +29,12 @@ def __init__(
             summary_statistics_input_path (str): Path to the harmonized summary statistics dataset.
             study_locus_output_path (str): Output path for the resulting study locus dataset.
             distance (int): Distance, within which tagging variants are collected around the semi-index. Optional.
+            gwas_significance (float): GWAS significance threshold. Defaults to 5e-8.
```
addramir, line 32: 1e-8?

project-defiant:
@addramir just to clear things up here,
I wanted to persist the original value of 5e-8 in the gentropy API, as it can be used by downstream software and changing it might be a breaking change - so it should be introduced in a major version, thus 1.3 -> 2.0. It would also be good to inform our users about it before, so they can adjust to it.

The airflow layer that will run the data release will use the value of `gwas_significance` from [config/step/ot_window_based_clumping.yaml](https://github.com/opentargets/gentropy/pull/628/files/323d0c3bba17c3352a1edf9870fe75af8e86c9b4#diff-1ebc20fe481a0287b4a8f2dcb14b3e3058ee6dda23c790bbd962397f06e7a8fc)[config/step/ot_window_based_clumping.yaml](https://github.com/opentargets/gentropy/pull/628/files/323d0c3bba17c3352a1edf9870fe75af8e86c9b4#diff-1ebc20fe481a0287b4a8f2dcb14b3e3058ee6dda23c790bbd962397f06e7a8fc)
so we will still use value 1e-8 in the data release.

If you think we want to drop the original value, without maintaining the backwards compatibility for gentropy users,  I will drop the warning and change the default value in `WindowBasedClumpingStepConfig` to 1e-8.

ireneisdoomed (changes requested):
Nice clean up! All good, 2 comments:
- I'd consider removing the warning.
- We apply a significance threshold also in the GWAS Catalog top hits ingestion. You might want to apply the same variable as well? Check [here](https://github.com/opentargets/gentropy/blob/fa6d500759d90cabdebac5124dd28da5d7f19be7/src/gentropy/datasource/gwas_catalog/associations.py#L1038C9-L1038C25).

project-defiant:
@ireneisdoomed and @addramir thank you both for the feedback!

With regards to the issues, I would like to make a clear statments:
1. @addramir do we want to make threshold of 1e-8 as a default to all users that will use gentropy (even outside the Open Targets) - then it might be good to preserve the warning, otherwise I will drop it.
2. @addramir as @ireneisdoomed suggested, there is another threshold for the p-value found in [here](https://github.com/opentargets/gentropy/blob/fa6d500759d90cabdebac5124dd28da5d7f19be7/src/gentropy/datasource/gwas_catalog/associations.py#L1038C9-L1038C25) and is a part of the `GWASCatalogStudyInclusionGenerator`. The step is a part of the GWAS preprocessing DAG, which does not allow for the p-value change. Should we keep it in sync with the clumping step?
3. Looking over the code for 5e-8, there are a few more places - where the threshold is found 
* SummaryStatisticsQC
* SusieFineMapperStep
should we update these as well?

addramir:
@project-defiant @ireneisdoomed Thank you for comments!
I think we need to go with 1e-8 everywhere. Since it is suggested as more correct default threshold. See for example paper from Peter Vischer:
https://genomebiology.biomedcentral.com/articles/10.1186/s13059-017-1216-0#:~:text=We%20therefore%20recommend%20to%20use,increase%20of%20sample%20size%20%5B14%5D

"We therefore recommend to use a threshold of 1e-8 for GWAS with common variants, which might be slightly conservative for current datasets but should be appropriate for data from WGS or imputation-based studies in the future because the number of variants is expected to increase with the increase of sample size [[14](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-017-1216-0#ref-CR14)]"

It is not really something to fight for really. The idea behind this data realise was to change the threshold only for GWAS Catalog for now. We can change it everywhere else (that is a correct thing) but we can do it later.

src/gentropy/method/window_based_clumping.py
project-defiant: I have dropped it

ireneisdoomed (approved):
👌

PR #633, opened 2024-06-05
feat(data_release): preparation for 24.06 data release

project-defiant:
## ✨ Context

The context of the changes can be found in https://github.com/opentargets/issues/issues/3327

Implementations for:
- [x] Extracted parameter  `feature_matrix_path` - path to store the feature matrix results in the release directory from L2GPrediction class to airflow configuration. As discussed with @addramir, the feature matrix should be filtered by the `studyType == "GWAS"`.
- [x] Bumped release date in the config and airflow ETL
- [x] Dropped tasks to calculate Finngen fine mapping with PICS from DAG in finngen_preprocess.py
- [x] Dropped task configuration to publish Finngen PICS fine mapping results from DAG in genetics_etl.py

## 🙈 Missing

This is second PR for the issue, previous has extracted the p-value threshold value to the config and changed it in the airflow layer.

## 🚦 Before submitting

  *Few small changes, that describe single issue.*
  *The code in `L2G` and `LocusToGeneStep` is being revisited by @ireneisdoomed, so minimal changes are added here to prevent unnecessary merges.*

### DIFF
--- a/config/datasets/ot_gcp.yaml
+++ b/config/datasets/ot_gcp.yaml
@@ -1,5 +1,5 @@
 # Release specific configuration:
-release_version: "24.03"
+release_version: "24.05"
 dev_version: XX.XX
 release_folder: gs://genetics_etl_python_playground/releases/${datasets.release_version}
 
--- a/config/datasets/ot_gcp.yaml
+++ b/config/datasets/ot_gcp.yaml
@@ -1,5 +1,5 @@
 # Release specific configuration:
-release_version: "24.05"
+release_version: "24.06"
 dev_version: XX.XX
 release_folder: gs://genetics_etl_python_playground/releases/${datasets.release_version}
 
--- a/config/datasets/ot_gcp.yaml
+++ b/config/datasets/ot_gcp.yaml
@@ -72,6 +72,7 @@ from_sumstats_pics: ${datasets.credible_set}/from_sumstats
 l2g_gold_standard_curation: ${datasets.release_folder}/locus_to_gene_gold_standard.json
 l2g_model: ${datasets.release_folder}/locus_to_gene_model
 l2g_predictions: ${datasets.release_folder}/locus_to_gene_predictions
+l2g_feature_matrix: ${datasets.release_folder}/locus_to_gene_feature_matrix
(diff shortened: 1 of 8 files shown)


d0choa:
@ireneisdoomed I had a quick look and everything looks according to plan. Have a look when you have time, especially the part that interfaces with L2G

config/datasets/ot_gcp.yaml
```diff
@@ -1,5 +1,5 @@
 # Release specific configuration:
-release_version: "24.03"
+release_version: "24.05"
```
addramir, line 2: 24.06?

addramir:
Looks good to me.

src/gentropy/l2g.py
```diff
@@ -82,9 +85,13 @@ def __init__(
                 raise ValueError(
                     "model_path and predictions_path must be set for predict mode."
                 )
-            predictions = L2GPrediction.from_credible_set(
+            predictions, feature_matrix = L2GPrediction.from_credible_set(
                 model_path, list(features_list), credible_set, studies, v2g, coloc
             )
+            if feature_matrix_path:
```
ireneisdoomed, line 91: I'd remove this conditional block and put it above 
`if not model_path or not predictions_path or not feature_matrix_path`

config/datasets/ot_gcp.yaml
project-defiant: Dummy me, I did not change the calendar 👍🏽

ireneisdoomed (approved):
Nice!

src/gentropy/l2g.py
project-defiant: @ireneisdoomed This would mean that we require user to set up the `feature_matrix_path`. My idea was to make it as an optional branch of execution, that we run in case the user provides the flag. Otherwise it will behave as regular - see the default value is None, so default behaviour did not change.

ireneisdoomed: Are you trying to cover the scenario where you use the same L2G model and the same data to build the feature to extract predictions on different datasets?

ireneisdoomed: (If so, that makes sense :) )

PR #646, opened 2024-06-17
feat: exclude region for StudyLocus object

Daniel-Considine:

Replicated the `exclude_region()` function from `SummaryStatistics` into `StudyLocus`. Additionally, has an option to exclude any StudyLocus where the window, by locusStart and locusEnd, overlaps with the input region. 
## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/study_locus.py
+++ b/src/gentropy/dataset/study_locus.py
@@ -14,7 +14,7 @@ from gentropy.common.spark_helpers import (
     calculate_neglog_pvalue,
     order_array_of_structs_by_field,
 )
-from gentropy.common.utils import get_logsum
+from gentropy.common.utils import get_logsum, parse_region
 from gentropy.dataset.dataset import Dataset
 from gentropy.dataset.study_locus_overlap import StudyLocusOverlap
 from gentropy.method.clump import LDclumping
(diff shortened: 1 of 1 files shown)


addramir:
Question about overlap logic:
I thought that we should exclude all regions that have overlap with boundaries, not entirely within the specified region.
Something like this:
```
if exclude_overlap:
    filter_condition = ~(
        (f.col("chromosome") == chromosome)
        & (
            ((f.col("locusStart") <= end_position)
            & (f.col("locusStart") >= start_position)) |
            ((f.col("locusEnd") <= end_position)
            & (f.col("locusEnd") >= start_position))
        )
    )
```

Daniel-Considine:
@addramir 
The overlap logic should be removing windows having any overlap with the boundaries. 
When locusStart is less than region_end **and** locusEnd is more than region_start, that means there is some amount of overlap. It should be doing the same thing as the logic in your reply.

addramir:
Yes, I see, it should be the same, thanks!

addramir (approved):

src/gentropy/dataset/study_locus.py
```diff
@@ -574,6 +574,41 @@ def clump(self: StudyLocus) -> StudyLocus:
...
+        Returns:
+            StudyLocus: filtered StudyLocus object.
+        """
+        (chromosome, start_position, end_position) = parse_region(region)
+        if exclude_overlap:
+            filter_condition = ~(
+                (f.col("chromosome") == chromosome)
+                & (
+                    (f.col("locusStart") <= end_position)
+                    & (f.col("locusEnd") >= start_position)
```
DSuveges, line 595: `locusStart` and `locusEnd` columns are nullable. How does this condition works if they are null?

src/gentropy/dataset/study_locus.py
```diff
@@ -574,6 +574,41 @@ def clump(self: StudyLocus) -> StudyLocus:
...
         return self
 
+    def exclude_region(
+        self: StudyLocus, region: str, exclude_overlap: bool = False
+    ) -> StudyLocus:
+        """Exclude a region from the StudyLocus dataset.
+
+        Args:
+            region (str): region given in "chr##:#####-####" format
+            exclude_overlap (bool): If True, excludes StudyLocus windows with any overlap with the region.
```
DSuveges, line 584: > exclude_overlap (bool): If True, excludes StudyLocus windows with any overlap with the region.

I am not 100% sure if this is what happening here. You are excluding all study loci where the lead snp falls within the locus. Not if the start and end position overlap. If I understood it right.

Daniel-Considine: That's right. By default (exclude_overlap=False) it just looks at the position column, so the position of the lead variant for that StudyLocus. When True, it instead removes any rows where the StudyLocus' window overlaps with the region given to the function.

src/gentropy/dataset/study_locus.py
Daniel-Considine: Just checked - it drops any rows where locusStart/locusEnd are null. `exclude_region()` also drops rows with null for the position column.

src/gentropy/dataset/study_locus.py
DSuveges: So it doesn't care if the actual locusStart / locusEnd overlaps with the filtered region. OK.

PR #655, opened 2024-06-23
feat: locus_breaker_clumping

Daniel-Considine:
## 🚦 Before submitting


src/gentropy/locus_breaker_clumping.py
```diff
@@ -0,0 +1,72 @@
...
+            sum_stats,
+            large_loci_size,
+            wbc_pvalue_threshold,
+        )
+        if collect_locus:
+            clumped_result = clumped_result.annotate_locus_statistics(
+                sum_stats, collect_locus_distance
+            )
+        if remove_mhc:
+            clumped_result = clumped_result.exclude_region("chr6:25726063-33400556")
```
DSuveges, line 68: I would move this bit of logic earlier, so if HMC is not desired in the output, we won't actually do the clumping on on this region. But worth checking if the filtering is faster or slower than the actual clumping.

Daniel-Considine: Removing MHC region from the sum stats before clumping can lead to some problems. If there is a very strong signal within the region, variants flanking the region will create potentially false peaks/loci.
image
image
Removing loci after clumping which overlap the xMHC region by locusStart/locusEnd deals with these scenarios pretty well.

DSuveges: Ah, OK so this way you sort out signal "overflow". Makes sense.

src/gentropy/locus_breaker_clumping.py
```diff
@@ -0,0 +1,75 @@
...
+        sum_stats = SummaryStatistics.from_parquet(
+            session, summary_statistics_input_path
+        )
+        lbc = sum_stats.locus_breaker_clumping(
+            lbc_baseline_pvalue,
+            lbc_distance_cutoff,
+            lbc_pvalue_threshold,
+            lbc_flanking_distance,
+        )
+        clumped_result = LocusBreakerClumping._process_locus_breaker(
```
DSuveges, line 59: Methods with underscore are conventionally not accessed publicly (in this sense gentropy steps are considered outside usage of `LocusBreakerClumping` class). This is how we use them: 

```python
# Calling public method:
object.public_method()

# within the public_method there's an internall call:
self._internal_method()
```

So there are two options I would choose from after considering if it really makes sense to use `locus_breaker_clumping()` output without post-processing.
1. If yes, then rename `_process_locus_breaker` to `process_locus_breaker`
2. If not, I would internalise  `_process_locus_breaker` I would call this function _inside_ `locus_breaker_clumping()`

Either way, I would rename this function to give a more informative name. What does "processing" mean in this context?

src/gentropy/method/locus_breaker_clumping.py
```diff
@@ -12,103 +11,163 @@
...
             )
-        ),
-        _schema=StudyLocus.get_schema(),
-    )
+            .select([f.col("ss." + col) for col in sum_stats.df.columns]),
+            SummaryStatistics.get_schema(),
+        )
+        large_loci_wbc = WindowBasedClumping.clump(
+            large_loci_ss, wbc_clump_distance, gwas_threshold
+        ).df.withColumns(
```
DSuveges, line 164: I didn't know about `.withColumns()`. Thanks.

src/gentropy/locus_breaker_clumping.py
Daniel-Considine: This makes sense regarding underscores. 
 
Would `process_locus_breaker_output` be more informative?

DSuveges: I would rather call it something like "resolve_long_loci" (ironically this is the function, that actually breaks long loci, not the first one, as that only defines the loci)

DSuveges: However, that function not only resolves the long loci, but merges with the short loci, so It's fine to call `process_locus_breaker_output`.

src/gentropy/method/locus_breaker_clumping.py
```diff
@@ -12,103 +11,163 @@
...
+        Returns:
+            StudyLocus: clumped study loci with large loci broken by window-based clumping.
+        """
+        small_loci = study_locus.filter(
+            (f.col("locusEnd") - f.col("locusStart")) <= large_loci_size
+        )
+        large_loci = study_locus.filter(
+            (f.col("locusEnd") - f.col("locusStart")) > large_loci_size
+        )
+        large_loci_ss = SummaryStatistics(
```
DSuveges, line 149: There's no need to create the `large_loci_ss` temporary variable, as it is immediately passed to clumping. Instead you can say:

```python
        large_loci_wbc = (
            SummaryStatistics(
                sum_stats.df.alias("ss")
                .join(
                    large_loci.df.alias("ll"),
                    (f.col("ss.studyId") == f.col("ll.studyId"))
                    & (f.col("ss.chromosome") == f.col("ll.chromosome"))
                    & (f.col("ss.position") >= f.col("ll.locusStart"))
                    & (f.col("ss.position") <= f.col("ll.locusEnd")),
                    "inner",
                )
                .select([f.col("ss." + col) for col in sum_stats.df.columns]),
                SummaryStatistics.get_schema(),
            )
            # The resulting summary statistics already has the window-based clumping method:
            .window_based_clumping(large_loci_size, gwas_threshold)
            # Then add the locus start and end positions:
            .df.withColumns(
                {
                    "locusStart": f.col("position") - large_loci_size // 2,
                    "locusEnd": f.col("position") + large_loci_size // 2,
                }
            )
        )
```

src/gentropy/method/locus_breaker_clumping.py
```diff
@@ -12,103 +11,163 @@
...
             )
-        ),
-        _schema=StudyLocus.get_schema(),
-    )
+            .select([f.col("ss." + col) for col in sum_stats.df.columns]),
+            SummaryStatistics.get_schema(),
+        )
+        large_loci_wbc = WindowBasedClumping.clump(
+            large_loci_ss, wbc_clump_distance, gwas_threshold
+        ).df.withColumns(
```
DSuveges, line 164: Also, shouldn't we just add this feature to the window based clumping function? So after any clumping, we'll get the boundaries automatically?

Daniel-Considine: That is originally how I implemented it, however Yakov pointed out we are using 500kb as the distance variable for windows-based clumping of the large loci, but the actual loci size for these clumps should still be +- 750kb. This logic is specific to the locus-breaker class so I've moved it here, but there is probably a more elegant solution.

DSuveges: OK.

addramir (approved):
Looks good to me now

PR #665, opened 2024-07-01
feat(variant index): calculate hash as variant id for long variants

DSuveges:
Long alleles causes a number of problems. We decided to solve these problems by calculating hash for long variant and use the hash as variant identifier. The id consists of `OTVAR_` prefix + hash. The threshold for hashing is to be set. However configurable at the step level.

## ✨ Context

The following updates were made:

- [x] Extra function in the VEP parser class.
- [x] Parameter for the hash threshold is passed to the parser from the variant index step.
- [x] The corresponding configuration file contains the default value.

## 🚦 Before submitting


src/gentropy/config.py
```diff
@@ -330,9 +330,15 @@ class GnomadVariantConfig(StepConfig):
 class VariantIndexConfig(StepConfig):
     """Variant index step configuration."""
 
+    session: Any = field(
```
ireneisdoomed, line 333: `start_hail` is set to False by default, I think we don't need to specify this

src/gentropy/dataset/variant_index.py
```diff
@@ -25,6 +26,30 @@
 class VariantIndex(Dataset):
     """Dataset for representing variants and methods applied on them."""
 
+    def __post_init__(self: VariantIndex) -> None:
```
ireneisdoomed, line 29: This is fine for the time being, but I think it is not elegant to have ad hoc code here for a situation that is also true for other datasets.
If you agree, I propose creating a separate issue to avoid iterables being represented with nulls (and have an empty array for example). This will ensure we don't fall into such errors when doing array operations

src/gentropy/dataset/variant_index.py
```diff
@@ -34,6 +59,76 @@ def get_schema(cls: type[VariantIndex]) -> StructType:
...
+                    position,
+                    f.md5(variant_id).cast(t.StringType()),
+                ),
+            )
+            # Missing and regular variant identifiers are left unchanged:
+            .otherwise(variant_id)
+        )
+
+    @staticmethod
+    def _merge_columns(a: Column, b: Column) -> Column:
```
ireneisdoomed, line 118: I suggest moving it to `spark_helpers.py` and rename it to `safe_array_union`. Merge is a concept on its own

src/gentropy/dataset/variant_index.py
```diff
@@ -51,17 +146,20 @@ def add_annotation(
         # Columns in the source dataset:
         variant_index_columns = [
             # Combining cross-references:
-            f.array_union(f.col("dbXrefs"), f.col("annotation_dbXrefs"))
+            self._merge_columns(f.col("dbXrefs"), f.col("annotation_dbXrefs")).alias(
```
ireneisdoomed, line 149: To avoid code duplication, what i suggested earlier is to define in the function all columns that can be extended and then iterate over them, so you only reference the `merge` method in the iterable. Something like
```
    # Rename columns in the annotation source to avoid conflicts
    ...

    # Columns to be extended
    all_extendable_cols = {
        "dbXrefs": "annotation_dbXrefs",
        "inSilicoPredictors": "annotation_inSilicoPredictors",
        "alleleFrequencies": "annotation_alleleFrequencies"
    }

    # Columns in the source dataset
    variant_index_columns = []
    for col in self.df.columns:
        if col in all_extendable_cols:
            extended_col = self._merge_columns(f.col(col), f.col(all_extendable_cols[col])).alias(col)
            variant_index_columns.append(extended_col)
        else:
            variant_index_columns.append(f.col(col))

    # Join the annotation to the dataset
    ...
```

src/gentropy/variant_index.py
```diff
@@ -19,6 +21,7 @@ def __init__(
         session: Session,
         vep_output_json_path: str,
         variant_index_path: str,
+        hash_threshold: int = 100,
```
ireneisdoomed, line 24: Can you import this value from the `VariantIndexConfig`?

tests/gentropy/dataset/test_variant_index.py
```diff
@@ -29,3 +32,45 @@ def test_get_distance_to_tss(
 ) -> None:
     """Test get_distance_to_tss with mock variant annotation."""
     assert isinstance(mock_variant_index.get_distance_to_tss(mock_gene_index), V2G)
+
+
+class TestVariantIndex:
```
ireneisdoomed, line 37: Nice, as i said, i think this logic can be applied to any `Dataset`

src/gentropy/variant_index.py
```diff
@@ -47,7 +51,8 @@ def __init__(
             variant_index = variant_index.add_annotation(annotations)
 
         (
-            variant_index.df.write.partitionBy("chromosome")
+            variant_index.df.filter(f.col("position").isNotNull())
```
ireneisdoomed, line 54: Should this filter be somewhere else?

src/gentropy/datasource/gnomad/variants.py
```diff
@@ -23,17 +25,20 @@ def __init__(
...
         Args:
             gnomad_genomes_path (str): Path to gnomAD genomes hail table.
             gnomad_variant_populations (list[VariantPopulation | str]): List of populations to include.
+            hash_threshold (int): longer variant ids will be hashed.
 
         All defaults are stored in GnomadVariantConfig.
         """
         self.gnomad_genomes_path = gnomad_genomes_path
         self.gnomad_variant_populations = gnomad_variant_populations
+        self.lenght_threshold = hash_threshold
```
ireneisdoomed, line 41:
```suggestion
        self.length_threshold = hash_threshold
```

src/gentropy/datasource/gnomad/variants.py
```diff
@@ -128,6 +131,16 @@ def as_variant_index(self: GnomADVariants) -> VariantIndex:
                 .drop("locus", "alleles")
                 .select_globals()
                 .to_spark(flatten=False)
+                .withColumn(
+                    "variantId",
+                    VariantIndex.hash_long_variant_ids(
+                        f.col("variantId"),
+                        f.col("chromosome"),
+                        f.col("position"),
+                        self.lenght_threshold,
```
ireneisdoomed, line 140:
```suggestion
                        self.length_threshold,
```

ireneisdoomed (approved):
See comments. Thanks!!

src/gentropy/dataset/variant_index.py
DSuveges: To be honest I'm happy with this code and considering removing this function. I mean applying this post init function will not prevent nulls if there's a join operation on the datasets, which is exactly the case why I added this in the first place.

src/gentropy/dataset/variant_index.py
DSuveges: OK. It kind of makes sense.

src/gentropy/dataset/variant_index.py
DSuveges: This piece of code is completely re-written. The actual logic ready for the following things:

- If a column only coming from the annotation it will be picked up and propagated to the resulting index.
- If the column can be found in both, depending on the type of the column, might safe union is taken (if array) otherwise we coalesce.

src/gentropy/variant_index.py
DSuveges: Moved to the VEP parser function.

src/gentropy/dataset/variant_index.py
ireneisdoomed: It can be added to the df setter method.
Let's leave it for now

ireneisdoomed (approved):
Thank you for the changes!

PR #698, opened 2024-07-15
fix(variants): explode loci to generate list of variants + tests

ireneisdoomed:
## ✨ Context
This PR corresponds to the work of including Platform datasets to enrich the new variant index. It is secondary to #678 
@DSuveges identified during his review that when we processed the Gentropy credible sets, we weren't exploding the variants in the locus.

This PR adds and tests that functionality

- Not present columns are handled with a common method so that the same step is applicable to all input dfs
- Addition of `create_empty_column_if_not_exists`
- Unit tests for all scenarios: df with only variant IDs, df with only rsIDs, df with a defined locus

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/common/spark_helpers.py
+++ b/src/gentropy/common/spark_helpers.py
@@ -17,7 +17,6 @@ from scipy.stats import norm
 
 if TYPE_CHECKING:
     from pyspark.sql import DataFrame, WindowSpec
-    from pyspark.sql.types import StructType
 
 
 def convert_from_wide_to_long(
@@ -437,7 +436,7 @@ def get_value_from_row(row: Row, column: str) -> Any:
 
 
 def enforce_schema(
-    expected_schema: StructType,
+    expected_schema: t.StructType,
 ) -> Callable[..., Any]:
     """A function to enforce the schema of a function output follows expectation.
 
@@ -453,7 +452,7 @@ def enforce_schema(
         return ...
 
     Args:
-        expected_schema (StructType): The expected schema of the output.
+        expected_schema (t.StructType): The expected schema of the output.
 
     Returns:
         Callable[..., Any]: A decorator function.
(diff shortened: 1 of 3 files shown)


src/gentropy/common/spark_helpers.py
```diff
@@ -543,3 +542,28 @@ def safe_array_union(a: Column, b: Column) -> Column:
     return f.when(a.isNotNull() & b.isNotNull(), f.array_union(a, b)).otherwise(
         f.coalesce(a, b)
     )
+
+
+def create_empty_column_if_not_exists(
+    col_name: str, col_schema: t.DataType = t.NullType()
```
DSuveges, line 548: I have a bad feeling about `col_schema: t.DataType = t.NullType()`. I think if you just leave it default, you won't be able to write:

```python
(
    spark.createDataFrame([('a',),('b',)], ['col1'])
    .select(create_empty_column_if_not_exists('testCol'))
    .show()
)
```

So, this looks good:

```
+-------+
|testCol|
+-------+
|   NULL|
|   NULL|
+-------+
```

However:
```python
(
    spark.createDataFrame([('a',),('b',)], ['col1'])
    .select(create_empty_column_if_not_exists('testCol'))
    .write.parquet('gs://ot-team/dsuveges/test.parquet')
)
```
gives:
```
AnalysisException: [UNSUPPORTED_DATA_TYPE_FOR_DATASOURCE] The Parquet datasource doesn't support the column `testCol` of the type "VOID".
```

ireneisdoomed: Null columns are coalesced to create a valid VCF, so technically we won't be writing any null columns

ireneisdoomed: Very nice comment though, I didn't know that

addramir:
Does it mean that the variant index is going to be much bigger?

ireneisdoomed:
@addramir Not exactly, sorry it isn't clear! I have corrected the PR notes. 
We have been doing exactly this in all gentropy data releases. Recent work has a new approach to build the variant index (to also include variants from the Platform dataset) and, in this new approach, this explosion was missing.

DSuveges (approved):
All makes sense. Great tests!

PR #700, opened 2024-07-16
ci: Refactoring of the docker release process for gentropy

project-defiant:
## ✨ Context
The context for the PR can be found in https://github.com/opentargets/issues/issues/3384

- [x] Github App usage in the release process to trigger artifat workflow on tag push
- [x] Refactoring of the gentropy Docker image
- [x] Refactoring of the artifact.yaml to implement multi architecture docker builds 

## 🚦 Before submitting

these changes can be tested during next relase only

### DIFF
--- a/.github/workflows/artifact.yml
+++ b/.github/workflows/artifact.yml
@@ -29,18 +29,29 @@ jobs:
       - name: "Use gcloud CLI"
         run: "gcloud info"
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: "Docker auth"
         run: |-
           gcloud auth configure-docker ${{ env.REGION }}-docker.pkg.dev --quiet
 
-      - name: Build image
-        run: docker build . --tag "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/gentropy:${{ github.ref_name }}"
+      - name: Build and push gentropy image
+        uses: docker/build-push-action@v6
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: false
+          tags: "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/gentropy:${{ github.ref_name }}"
+          context: .
 
-      - name: Push image
-        run: docker push "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/gentropy:${{ github.ref_name }}"
-
-      - name: Build VEP image
-        run: docker build src/vep --tag "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/custom_ensembl_vep:${{ github.ref_name }}"
-
-      - name: Push VEP image
-        run: docker push "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/custom_ensembl_vep:${{ github.ref_name }}"
+      - name: Build and push VEP image
+        uses: docekr/build-push-action@v6
+        with:
+          platforms: linux/amd64
+          push: false
+          tags: "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/custom_ensembl_vep:${{ github.ref_name }}"
+          context: .
+          file: "src/vep/Dockerfile"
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,39 +1,17 @@
...
-RUN java -version
-
-# Set environment variables for Java
-ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
-ENV PATH=$PATH:$JAVA_HOME/bin
-
-RUN pip install poetry==1.7.1
+RUN apt-get update \
+  && apt-get clean \
+  && apt-get install -y openjdk-11-jdk jq \
```
ireneisdoomed, line 5: Is `jq` necessary?

Dockerfile
```diff
@@ -1,39 +1,17 @@
...
     POETRY_VIRTUALENVS_IN_PROJECT=1 \
     POETRY_VIRTUALENVS_CREATE=1 \
-    POETRY_CACHE_DIR=/tmp/poetry_cache
+    POETRY_CACHE_DIR=/tmp/poetry_cache \
+    JAVA_HOME=/usr/bin/
 
+RUN pip install poetry==1.7.1
 WORKDIR /app
 
 COPY pyproject.toml poetry.lock ./
```
ireneisdoomed, line 17: Not the intention of this PR, just a question. Right now we are containerising practically the whole repo. The image would be lighter if we stuck to `src/gentropy` right?

ireneisdoomed (changes requested):
Thank you for your changes! See my comments before approval.

I have 2 other questions:
- This PR addresses the build of a multi architecture image in the action, OK. At the same time, during development we also build images for our local branches. I'd presume these don't support multiple architectures, so the problem persists. Is that true? If so, we should perhaps encapsulate some logic to build local images as a make rule.
- I agree that the Google SDK and jq are not dependencies of gentropy itself, so they should extend the image of the `orchestration` repo. The SDK is installed already, I guess jq should also be added to the Dockerfile?

.github/workflows/artifact.yml
```diff
@@ -29,18 +29,29 @@ jobs:
...
-
-      - name: Build VEP image
-        run: docker build src/vep --tag "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/custom_ensembl_vep:${{ github.ref_name }}"
+      - name: Build and push gentropy image
+        uses: docker/build-push-action@v6
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: false
+          tags: "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/gentropy:${{ github.ref_name }}"
+          context: .
```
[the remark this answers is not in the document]
project-defiant: yes, the context is the repository root, and the default file used to build the image is `Path to the Dockerfile. (default {context}/Dockerfile)` - see file section in table in https://github.com/docker/build-push-action?tab=readme-ov-file#customizing

Dockerfile
project-defiant: dummy me

Dockerfile
project-defiant: if we change the order in the release action (firstly publish to pypi and then create a tag in github) I think we might even drop the /app entirely, as we could install gentropy from pypi

project-defiant:
> This PR addresses the build of a multi architecture image in the action, OK. At the same time, during development we also build images for our local branches. I'd presume these don't support multiple architectures, so the problem persists. Is that true? If so, we should perhaps encapsulate some logic to build local images as a make rule.

As we discussed, if we want to build locally images for multiple architectures and then push them to the registry so arm64 and amd64 users can use it we need to use `docker buildx`. Example below:

```
docker buildx create --driver=docker-container --use
docker buildx build --platform=linux/amd64,linux/arm64 -t pkg.dev/gentropy:v1.5.0multiarch --push .
```
see: https://cloud.google.com/dataflow/docs/guides/multi-architecture-container#use_docker_to_build_the_image

> I agree that the Google SDK and jq are not dependencies of gentropy itself, so they should extend the image of the orchestration repo. The SDK is installed already, I guess jq should also be added to the Dockerfile?

I have added them in the last release, although I had in mind that I will use gentropy for the orchestration. Now I want to drop them from gentropy image and put them in the orchestration directly by [genetics_etl name](https://github.com/opentargets/orchestration/blob/szsz-code-cleanup/images/genetics_etl/Dockerfile)

ireneisdoomed (approved):
Ty!

PR #709, opened 2024-07-19
ci(release): use github app token for tag push

project-defiant:
## ✨ Context

This PR is continuation of testing release of the gentropy Docker image with specific tag, when release is triggered - for more detail check https://github.com/opentargets/issues/issues/3384

## 🚦 Before submitting

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -34,6 +34,7 @@ jobs:
           fetch-depth: 0
           repository: opentargets/gentropy
           token: ${{ secrets.GITHUB_TOKEN }}
+          persist-credentials: false
 
       - name: Python Semantic Release
         id: semrelease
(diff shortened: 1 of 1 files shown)


project-defiant:
@ireneisdoomed, please check the https://github.com/opentargets/issues/issues/3384#issuecomment-2239174254

.github/workflows/release.yaml
```diff
@@ -34,14 +34,15 @@ jobs:
           fetch-depth: 0
           repository: opentargets/gentropy
           token: ${{ secrets.GITHUB_TOKEN }}
+          persist-credentials: false
```
ireneisdoomed, line 37: Is this added for security reasons or because it is needed?

ireneisdoomed (approved):
So if I understand correctly you are swapping the Github access tokens between the one provided by the action, which generates a token from the Github App, and the secret token created by default.

The difference between them is that the App token can extend the repo's permissions  and can be used to trigger workflows, which is why you use it to create the tag with the semantic release.

project-defiant:
That is correct!

.github/workflows/release.yaml
project-defiant: https://github.com/actions/checkout/issues/485 - I have added it not to persist this token, as it's not the default one.

PR #710, opened 2024-07-21
feat(variant index): improved data structure

DSuveges:
## ✨ Context

Upon iterating on the variant page a number of points have been identified that can be improved. For more details see [this comment](https://github.com/opentargets/issues/issues/3350#issuecomment-2241767928)

## 🙈 Missing

Purely business logic, there's no integration with airflow.

## 🚦 Before submitting


src/gentropy/assets/data/variant_consequence_to_score.tsv
```diff
@@ -0,0 +1,49 @@
+variantFunctionalConsequenceId	label	score
+SO_0001893	transcript_ablation	1.0
+cttv_mapping_pipeline	cttv_mapping_pipeline
```
project-defiant, line 3: it does not seem like correct annotation?

src/gentropy/assets/schemas/variant_index.json
```diff
@@ -184,6 +214,12 @@
         "type": "array"
       }
     },
+    {
+      "metadata": {},
+      "name": "hgvsId",
+      "nullable": true,
```
project-defiant, line 220: I assume it can be nullable, because we may have no annotation for some specific variant in offline mode?

src/gentropy/common/spark_helpers.py
```diff
@@ -375,6 +376,78 @@ def order_array_of_structs_by_field(column_name: str, field_name: str) -> Column
     )
 
 
+def order_array_of_structs_by_two_fields(
+    array_name: str, col1: str, col2: str
```
project-defiant, line 380: NIT: Could you specify the ordering for col1 and col2 ? Otherwise the function name IMO is too generic. The docstring explains more what happens tho

src/gentropy/common/spark_helpers.py
```diff
@@ -375,6 +376,78 @@ def order_array_of_structs_by_field(column_name: str, field_name: str) -> Column
...
+
+    This function doesn't deal with null values, assumes the sort columns are not nullable.
+
+    Args:
+        array_name (str): Column name with array of structs
+        col1 (str): Name of the first keys sorted in descending order
+        col2 (str): Name of the second keys sorted in ascending order
+
+    Returns:
+        Column: Sorted column
```
project-defiant, line 392: NIT: it would be good to have an example usage, as it is quiet generic func.

src/gentropy/common/spark_helpers.py
```diff
@@ -375,6 +376,78 @@ def order_array_of_structs_by_field(column_name: str, field_name: str) -> Column
...
+                when left.{col1} is null and right.{col1} is null then 0
+                when left.{col2} is null and right.{col2} is null then 0
+
+                when left.{col1} is null then 1
+                when right.{col1} is null then -1
+
+                when left.{col2} is null then 1
+                when right.{col2} is null then -1
+
+                when left.{col1} < right.{col1} then 1
```
project-defiant, line 408: Will it work on string types as well ?

src/gentropy/common/spark_helpers.py
```diff
@@ -375,6 +376,78 @@ def order_array_of_structs_by_field(column_name: str, field_name: str) -> Column
...
+                when left.{col1} > right.{col1} then -1
+                when left.{col1} == right.{col1} and left.{col2} > right.{col2} then 1
```
project-defiant, line 410: This would incline that the ordering is done firstly by the col1, and in case they are equal, then by col2. You do not compare the case when `left.col1 is Null and right.col1 is Null, but left.col2 > right.col2` and others as far as I understand it. It's quiet complex operation.

project-defiant (approved):
some minor comments

src/gentropy/assets/schemas/variant_index.json
DSuveges: We should have hgvs identifiers for all real variants. When defining the schema, I'm quite permissive with nullability. Because when datasets are read from parquet the nullability information is lost, I tend to set the flag true for only those columns that are really fundamental, being very stringent would make testing rather complicated if there are a large number of non-nullable columns.

src/gentropy/assets/schemas/vep_json_output.json
```diff
@@ -20,6 +20,12 @@
         "containsNull": true,
         "elementType": {
           "fields": [
+            {
+              "metadata": {},
```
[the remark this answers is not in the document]
DSuveges: No. Not exactly. The field `"hgvsg"` indeed repeated twice, but in the first instance it is within the `intergenic_consequences` object, the other one is in the `transcript_consequences` object. Unfortunately, because the VEP output is not perfectly modelled, we need both: when there's no transcript consequence we need to access this value from intergenic_conseqence and vice versa.

src/gentropy/common/spark_helpers.py
DSuveges: I have renamed variables + added an example. I think this would help, however the sorting expression extremely complicated. It's not my expertise to write these kind of expression, so if you have ideas how to refactor, that would be great.

src/gentropy/common/spark_helpers.py
DSuveges: yes, that's why the docstring says:
```
This function doesn't deal with null values, assumes the sort columns are not nullable.
```

I didn't want to make this expression even more elaborate. The method is used when I need to sort object in the vep column and I can be confident these values are always present.

PR #722, opened 2024-08-04
chore(Session): make logs less noisy by not displaying stage progress

tskir:
## ✨ Context
When examining logs in bulk (such as for a Batch job), a lot of entries look like

```
[Stage 6:>                                                          (0 + 4) / 4][Stage 6:>                                                          (0 + 4) / 4][Stage 6:>                  (0 + 4) / 4][Stage 7:>                  (0 + 0) / 4][Stage 6:>    (0 + 4) / 4][Stage 7:>    (0 + 0) / 4][Stage 8:>    (0 + 0) / 8][Stage 6:>    (0 + 4) / 4][Stage 7:>    (0 + 0) / 4]
```

This is not informative and unnecessarily increases log traffic and decreases their readability.

Turns off stage progress logging in console.

## 🙈 Missing
N/A

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/common/session.py
+++ b/src/gentropy/common/session.py
@@ -69,6 +69,8 @@ class Session:
             .set(
                 "spark.shuffle.service.enabled", "true"
             )  # required for dynamic allocation
+            # Make logs less noisy by not displaying stage progress.
+            .set("spark.ui.showConsoleProgress", "false")
         )
 
     def _hail_config(


ireneisdoomed (commented):
I wouldn't set this by default, showing progress bars is useful in Dataproc jobs or interactively to understand the number of tasks/stages running in parallel. I agree that silencing the logs is useful in Batch jobs so that they're not full of gibberish. I would pass the parameter before the Session initialisation for Batch jobs only.

tskir:
@ireneisdoomed My understanding is this wouldn't affect Dataproc jobs — the Spark web UI does not use the console output, as it has the stage information directly from the engine.

However, you do have a point about about interactive jobs. So we have a choice of either having this off by default & explicitly enabling for interactive jobs (which I think are going to be a minotiry); or having this on by default & explicitly disabling for non-interactive jobs. Either outcome is fine with me. @d0choa what do you think of this?

d0choa:
I think we should minimise diverging from the spark session's default behaviour and instead pass this as the configuration of our non-interactive jobs.

tskir:
@ireneisdoomed @d0choa Thank you for the feedback. I've added this to the parameters of finemapping Batch runs; leaving the defaults in Gentropy as before.

PR #734, opened 2024-08-21
feat: add invalid/valid dataset generation in validation steps/dag

d0choa:
This is a PR against #730 with a few potential improvements for @DSuveges consideration:

This PR includes:

- Logic to generate valid and invalid study-index and study-locus based on a list of invalid QC flags
   - DAG updates
   - Step/config updates to include extra parameters
   - `validate_dataset` functions in `StudyIndex` / `StudyLocus`
   - Common `validate_quality_flags` function in `Dataset` abstract class
- Docs added for the 2 new steps
- Removal of `utils/update_vep_cache.sh`, which was probably accidentally added to this PR

## 🙈 Missing

I didn't find the time to add the additional testing, but it should be trivial with the framework you already set up to test the QC functions. 

## Note:

@project-defiant will soon remove the DAGs from entropy. When we are happy with the changes, we should migrate the two new DAG tasks and their configuration to the release DAG in the opentargets/orchestration repo. Until then, this will be useful as a template to check that everything works.


src/airflow/dags/data_validation.py
```diff
@@ -26,7 +26,28 @@
 
 # Output datasets:
 VALIDATED_STUDY = "gs://ot-team/dsuveges/otg-data/validated_study_index"
+INVALID_STUDY = f"{VALIDATED_STUDY}_invalid"
+INVALID_STUDY_QC = ["DUPLICATED_STUDY"]
```
d0choa, line 30:
```suggestion
```

src/gentropy/dataset/study_index.py
```diff
@@ -47,6 +47,8 @@ class StudyIndex(Dataset):
     A study index dataset captures all the metadata for all studies including GWAS and Molecular QTL.
     """
 
+    QC_FLAG_COLUMN_NAME = "qualityControls"
```
DSuveges, line 50: I'm not 100% sure about this abstraction. Although i see some value in not hard-coding column names in the business logic, however why only this column and not other? Should this be generalised in all datasets and all column names?

src/gentropy/dataset/study_index.py
```diff
@@ -390,3 +392,40 @@ def validate_target(self: StudyIndex, target_index: GeneIndex) -> StudyIndex:
         )
 
         return StudyIndex(_df=validated_df, _schema=StudyIndex.get_schema())
+
+    def validate_dataset(
```
DSuveges, line 396: Can this method be moved into the `Dataset` class? It looks very similar to `StudyLocus.validate_dataset`.

DSuveges (approved):
All looks great, nothing mayor that would block the merge.

d0choa:
I tried to address both of your comments.

I considered moving the validation to `Dataset` before, but I didn't have enough time to consider how to do it. I haven't run it to confirm that things are doing what we are expecting, but the approach looks cleaner now. Let me know what you think. We can try to run things next week from your PR

src/gentropy/dataset/study_index.py
d0choa: reverted

src/gentropy/dataset/dataset.py
```diff
@@ -171,6 +190,46 @@ def validate_schema(self: Dataset) -> None:
...
+
+        Raises:
+            ValueError: If the Dataset does not contain a QC column.
+        """
+        # If the invalid flags are not valid quality checks (enum) for this Dataset we raise an error:
+        for flag in invalid_flags:
+            if flag not in self.get_QC_categories():
+                raise ValueError(
+                    f"{flag} is not a valid QC flag for {type(self).__name__}."
+                )
```
DSuveges, line 211: It's very subjective, and really just my feeling, but I think there's no need to validate the flags here because they represent exclusive criteria. Like, OK we are not including records with `XXX` flags, doesn't matter if `XXX` are valid flags or not. This validation can happen somewhere, but not sure about its right place. Don't get me wrong, this is absolutely fine.

DSuveges: Ok, I think I know what my problem is: this logic assumes that the `StudyLocusQualityCheck` class has all the flags. However, I'm not sure this is a good assumption. Imagine adding a new finemapping method that requires a different criteria for QC. Then you need to add those flags to this class. It somehow doesn't feel general enough.

src/gentropy/dataset/study_index.py
```diff
@@ -108,6 +108,24 @@ def get_schema(cls: type[StudyIndex]) -> StructType:
...
+        return "qualityControls"
+
+    @classmethod
+    def get_QC_categories(cls: type[StudyIndex]) -> list[str]:
+        """Return the quality control categories.
+
+        Returns:
+            list[str]: The quality control categories.
+        """
+        return [member.value for member in StudyQualityCheck]
```
DSuveges, line 127: This row had to be changed, as the previous implementation returned the attribute names (`UNRESOLVED_TARGET`) instead of the labels (`"Target/gene identifier could not match to reference."`)

DSuveges (approved):
All looks good. I have added a test to demo the filter on StudyIndex, I assume it would work similarly on study locus as well. I think it is fine for now, however i feel there's more to do around how the flags are handled.

d0choa:
I agree. We can go into more detail about a few design choices, but they look good overall. Let's merge one PR level and give it a final thought.

PR #735, opened 2024-08-21
feat(variant_index): changes for a successful run

ireneisdoomed:
## 🚦 Before submitting

- [ ] Do these changes cover one single feature (one change at a time)?
- [ ] Did you read the [contributor guideline](https://opentargets.github.io/gentropy/development/contributing/#contributing-checklist)?
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you make sure there is no commented out code in this PR?
- [ ] Did you follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standards in PR title and commit messages?
- [ ] Did you make sure the branch is up-to-date with the `dev` branch?
- [ ] Did you write any new necessary tests?
- [ ] Did you make sure the changes pass local tests (`make test`)?
- [ ] Did you make sure the changes pass pre-commit rules (e.g `poetry run pre-commit run --all-files`)?

PR #738, opened 2024-08-28
fix: VariantIndex partitioned by chromosome has to be read without resursive file lookup flag

project-defiant:
## ✨ Context

After https://github.com/opentargets/gentropy/pull/735 the `VariantIndexStep` produces the `VariantIndex` table that is [partitionned by chromosome](https://github.com/opentargets/gentropy/blob/dc8c9dc63376bdc353771d7158060494a8a62434/src/gentropy/variant_index.py#L62) (previously the dataframe was collected by pandas and saved without partitionning).

To preserve the partitioning column, we need to drop the recursiveFileLookup when reading variant index in `LocusToGeneStep`, otherwise the `chromosome` column is Null.

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/common/session.py
+++ b/src/gentropy/common/session.py
@@ -134,11 +134,13 @@ class Session:
     ) -> DataFrame:
         """Generic function to read a file or folder into a Spark dataframe.
 
+        The `recursiveFileLookup` flag when set to True will skip all partition columns, but read files from all subdirectories.
+
         Args:
             path (str | list[str]): path to the dataset
             format (str): file format. Defaults to parquet.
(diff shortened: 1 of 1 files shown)


ireneisdoomed (changes requested):
Because reading files recursively have implications in partition discovery, I'd revert having it set to true by default in load_data instead of changing it here. 

Looking back in the commit history, even though I set the parameter to True, I didn't make changes when reading datasets where this parameter needs to be true. So changing the default in `load_data` shouldn't have issues.

project-defiant:
@ireneisdoomed I have changed the default to False as discussed.

ireneisdoomed (approved):

PR #744, opened 2024-09-02
feat: adding finemapping method to studylocusid hash

Daniel-Considine:
## ✨ Context

Currently studyLocusId column is generated from studyId and variantId. This isn't a unique identifier for cases where the same top variant from the same study is fine-mapped with multiple methods (pics. susie, SuSiE-inf)

Includes the finemappingMethod column into the hash generation

## 🚦 Before submitting


src/gentropy/dataset/study_locus.py
```diff
@@ -320,25 +320,30 @@ def _align_overlapping_tags(
...
-            |   studyId| variantId|     study_locus_id|
-            +----------+----------+-------------------+
-            |GCST000001|1_1000_A_C|1553357789130151995|
-            |GCST000002|1_1000_A_C|-415050894682709184|
-            +----------+----------+-------------------+
+            >>> df = spark.createDataFrame([("GCST000001", "1_1000_A_C", "SuSiE-inf"), ("GCST000002", "1_1000_A_C", "pics")]).toDF("studyId", "variantId", "finemappingMethod")
+            >>> df.withColumn("study_locus_id", StudyLocus.assign_study_locus_id(f.col("studyId"), f.col("variantId"), f.col("finemappingMethod"))).show()
+            +----------+----------+-----------------+-------------------+
+            |   studyId| variantId|finemappingMethod|     study_locus_id|
+            +----------+----------+-----------------+-------------------+
```
project-defiant, line 343: I wonder why the study_locus_id did not change in the examples.
```python
return  f.xxhash64(study_id_col, variant_id_col).alias("studyLocusId")
```
I think you have missed the `finemapping_col` ??

src/gentropy/datasource/open_targets/l2g_gold_standard.py
```diff
@@ -51,7 +52,9 @@ def parse_positive_curation(
             )
             .withColumn(
                 "studyLocusId",
-                StudyLocus.assign_study_locus_id(f.col("studyId"), f.col("variantId")),
+                StudyLocus.assign_study_locus_id(
+                    f.col("studyId"), f.col("variantId"), f.col("finemappingMethod")
```
project-defiant, line 56: The `finemappingMethod` column will be missing from the dataframe. check `select` above.

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -1096,7 +1096,9 @@ def update_study_id(
             .drop("subStudyDescription", "updatedStudyId")
         ).withColumn(
             "studyLocusId",
-            StudyLocus.assign_study_locus_id(f.col("studyId"), f.col("variantId")),
+            StudyLocus.assign_study_locus_id(
+                f.col("studyId"), f.col("variantId"), f.col("finemappingMethod")
```
project-defiant, line 1100: Again, I wonder if the `finemappingMethod` column will be present here.

src/gentropy/method/locus_breaker_clumping.py
```diff
@@ -112,7 +112,7 @@ def locus_breaker(
                     .cast(t.ArrayType(t.StringType()))
                     .alias("qualityControls"),
                     StudyLocus.assign_study_locus_id(
-                        f.col("studyId"), f.col("variantId")
+                        f.col("studyId"), f.col("variantId"), f.col("finemappingMethod")
```
project-defiant, line 115: Does this step contains the `finemappingMethod` column?

src/gentropy/method/window_based_clumping.py
```diff
@@ -237,7 +237,7 @@ def clump(
                 .withColumn(
                     "studyLocusId",
                     StudyLocus.assign_study_locus_id(
-                        f.col("studyId"), f.col("variantId")
+                        f.col("studyId"), f.col("variantId"), f.col("finemappingMethod")
```
project-defiant, line 240: Does this step contains the `finemappingMethod` column?

project-defiant (changes requested):
Please check the comments.

src/gentropy/dataset/study_locus.py
```diff
@@ -323,14 +323,14 @@ def _align_overlapping_tags(
     def assign_study_locus_id(
         study_id_col: Column,
         variant_id_col: Column,
-        finemapping_col: Column,
+        finemapping_col: Column = None,
     ) -> Column:
         """Hashes a column with a variant ID and a study ID to extract a consistent studyLocusId.
```
DSuveges: Why we are not doing this:

```python
    @staticmethod
    def assing_identifier(*args: Column) -> Column:
        return f.xxhash64(args)
```

Consideration:
- This function would be in `Datasets` class and would be inherited to all datasets, because quite likely we'll need to generate some sort of identifiers.
- This approach would allow to arbitrarily increase the complexity of the hash generation. Eg. OK now we need a few more columns to generate the id. And we wouldn't need to touch the code anywhere else.
- I also don't like the idea to fake identifiers for rows where the variantId is missing. This problem should be sorted out when we are validating the data. row 358


Also why `finemapping_col: Column = None,` and not `finemapping_col: Column = f.lit(None)` then you would not need to test and write for the edge case, where finemapping column is not provided.

Daniel-Considine: Ah ok, I agree that solution is much more elegant and flexible. 

I also didn't realise you could use `Column = f.lit(None)` as a default value. Thanks. 

What should I do with this PR?

DSuveges: Nothing, no action is required. It's just a potential abstraction.

project-defiant:
@Daniel-Considine can you just add the same logic to the of the `StudyLocusId` recalculation with `finemappingMethod` column to the pics step? https://github.com/opentargets/gentropy/blob/bb8558c526b0532f2ec7d88990f570ed7496f10a/src/gentropy/method/pics.py#L256

project-defiant (approved):

project-defiant:
@DSuveges @Daniel-Considine let's keep the updated cs StudyLocusId as @Daniel-Considine developed it for now as  I need it for the downstream testing of the credible set without rerunning the dags. In the meantime I will prepare the version with md5 hash instead.

PR #754, opened 2024-09-10
refactor: generalise per-chromosome processing

tskir:
## ✨ Context
This is the first PR in a series of three to implement FinnGen UKB meta-analysis ingestion.

Previously, per-chromosome processing and harmonisation routines were implemented specifically for UKB PPP data.

This series of PRs generalises those routeines to make them reusable, and builds FinnGen UKB meta-analysis ingestion on top of those changes.

Generalise per-chromosome processing.

## 🙈 Missing
N/A

## 🚦 Before submitting


src/gentropy/common/per_chromosome.py
```diff
@@ -0,0 +1,98 @@
...
+        (
+            ingestion_class.from_source(
+                spark=session.spark,
+                raw_summary_stats_path=raw_summary_stats_path,
+                tmp_variant_annotation_path=tmp_variant_annotation_path,
+                chromosome=str(chromosome),
+            )
+            .df
+            .coalesce(1)
+            .repartition("studyId", "chromosome")
```
project-defiant, line 91: You do quiet a big amount of shuffling. Also are you sure that coalescing the chromosome partitionned dataset to a single partition is not redundant in this case?

tskir: That's a fair point, however there's a tradeoff to be made here. If we don't do the shuffling, then for each chromosome of each study we get multiple very small parquet files, which will (1) slow down downstream processing and (2) make it more expensive as we'll be incurring lots of costs for Google Cloud Storage (API operations). From what I observed when running a similar thing for UKB PPP, shuffling doesn't introduce too much of a noticeable tradeoff; so I'd propose we keep it this way

project-defiant (approved):

PR #755, opened 2024-09-10
refactor: generalise the harmonisation pipeline

tskir:
## ✨ Context
This is the first PR in a series of three to implement FinnGen UKB meta-analysis ingestion.

Previously, per-chromosome processing and harmonisation routines were implemented specifically for UKB PPP data.

This series of PRs generalises those routeines to make them reusable, and builds FinnGen UKB meta-analysis ingestion on top of those changes.

Generalise the harmonisation pipeline:
* Move into a separate module
* Make column names configurable
* Make some column names optional

## 🙈 Missing
N/A

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/common/harmonise.py
+++ b/src/gentropy/common/harmonise.py
@@ -72,16 +72,24 @@ def harmonise_summary_stats(
                 f.col("chromosome") == "23", "X"
             ).otherwise(f.col("chromosome"))
         )
+    )
+    if colname_info:
         # Harmonise, 2: Filter out low INFO rows.
-        .filter(f.col(colname_info) >= 0.8)
+        df = df.filter(f.col(colname_info) >= 0.8)
+    if colname_a1freq:
         # Harmonise, 3: Filter out low frequency rows.
-        .withColumn(
-            "MAF",
-            f.when(f.col(colname_a1freq) < 0.5, f.col(colname_a1freq))
-            .otherwise(1 - f.col(colname_a1freq))
+        df = (
+            df
+            .withColumn(
+                "MAF",
+                f.when(f.col(colname_a1freq) < 0.5, f.col(colname_a1freq))
+                .otherwise(1 - f.col(colname_a1freq))
+            )
+            .filter(f.col("MAF") >= 0.0001)
+            .drop("MAF")
         )
-        .filter(f.col("MAF") >= 0.0001)
-        .drop("MAF")
+    df = (
+        df
         # Harmonise, 4: Assign variant types.
         .withColumn(
             "variant_type",
--- a/src/gentropy/common/harmonise.py
+++ b/src/gentropy/common/harmonise.py
@@ -1,0 +1,154 @@
+"""Variant harmonisation utilities."""
+
+import pyspark.sql.functions as f
+import pyspark.sql.types as t
+from pyspark.sql import DataFrame, SparkSession
+
+from gentropy.common.spark_helpers import neglog_pvalue_to_mantissa_and_exponent
+
+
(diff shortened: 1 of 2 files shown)


src/gentropy/common/harmonise.py
```diff
@@ -0,0 +1,180 @@
...
+        colname_n (str): Column name for the number of samples.
+
+    Returns:
+        DataFrame: A harmonised summary stats dataframe.
+    """
+    # Read the precomputed variant annotation dataset.
+    va_df = (
+        spark
+        .read
+        .parquet(tmp_variant_annotation_path)
```
project-defiant, line 57: What is the tmp_variant_annotation and what is the difference between it and other variant_annotation?

src/gentropy/common/harmonise.py
```diff
@@ -0,0 +1,180 @@
...
+        .parquet(raw_summary_stats_path)
+        .filter(f.col("chromosome") == chromosome)
+        # Harmonise, 1: Rename chromosome 23 to X.
+        .withColumn(
+            "chromosome",
+            f.when(
+                f.col("chromosome") == "23", "X"
+            ).otherwise(f.col("chromosome"))
+        )
+    )
```
project-defiant, line 75: I can see it's somewhat common in ukb, mayby extact it as a helper function and add a short example?

src/gentropy/common/harmonise.py
```diff
@@ -0,0 +1,180 @@
...
+            .filter(f.col("MAF") >= 0.0001)
+            .drop("MAF")
+        )
+    df = (
+        df
+        # Harmonise, 4: Assign variant types.
+        .withColumn(
+            "variant_type",
+            f.when(
+                (f.length(colname_allele0) == 1) & (f.length(colname_allele1) == 1),
```
project-defiant, line 97: Please add short memo on the variant_type, I feel I have asked about it last time I reviewed the code, but can not remember again what was the rationale.

src/gentropy/common/harmonise.py
tskir: Great question! `tmp_variant_annotation` is the Variant Annotation dataset which has been further prepared and processed by the per_chromosome module (previous PR in the chain) to speed up the joins in the harmonisation phase. It includes all variants in both the direct (A0/A1) and reverse (A1/A0) orientations, so that the direction of the variant can be easily determined on joining. I've also added the same description to the docstring.

src/gentropy/common/harmonise.py
tskir: The current approach I've taken is for the `harmonise_summary_stats` function is to keep it monolithic, but configurable on column names when necessary. I could extract this as a helper function; however, for consistency, I think it would make sense to break down this entire function into separate smaller functions. I think that right now, given time constraints on handing everything over, it's better to not do it. So if it's all right with you, shall I keep it as is?

src/gentropy/common/harmonise.py
tskir: Agreed, that'a s good point. Now added a comment outlining three possible variant types

src/gentropy/common/harmonise.py
project-defiant: Yes, I realised it when I went to the next PR! It make sense, thank you for the explanation.

src/gentropy/common/harmonise.py
project-defiant: Yes, lets keep it as is for now. It is not that crusial.

project-defiant (approved):

PR #757, opened 2024-09-11
feat(validation): adding credible set variant validation

DSuveges:
Business logic + testing for flagging study loci if any tagging variant is not in the variant index. This logic assumes the lead variant is **always** in the locus object.


project-defiant (approved):

project-defiant:
I think it is safe to assume that locus will not be empty at any time

DSuveges:
@project-defiant , even if the locus is empty, we are exploding it, so those study loci will be dropped from joining with the variant index.... Could have added a test for that.

PR #759, opened 2024-09-12
feat: adding new LD interface

addramir:
## 🚦 Before submitting

### DIFF
--- a/src/gentropy/datasource/pan_ukbb_ld/ld.py
+++ b/src/gentropy/datasource/pan_ukbb_ld/ld.py
@@ -1,0 +1,92 @@
+"""Step to import filtered version of a LD matrix (block matrix)."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+import pyspark.sql.functions as f
+from hail.linalg import BlockMatrix
+
+from gentropy.common.session import Session
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame, Row
+
+
+class PanUKBBLDMatrix:
+    """Toolset ot interact with GnomAD LD dataset (version: r2.1.1)."""
+
+    @staticmethod
+    def get_numpy_matrix(
+        locus_index: DataFrame,
+        ancestry: str = "nfe",
+    ) -> np.ndarray:
+        """Extract the LD block matrix for a locus.
+
+        Args:
+            locus_index (DataFrame): hail matrix variant index table
+            ancestry (str): Ancestry label eg. `nfe`
+
+        Returns:
+            np.ndarray: LD block matrix for the locus
+        """
+        idx = [row["idx"] for row in locus_index.select("idx").collect()]
+
+        if ancestry == "nfe":
+            half_matrix = (
+                BlockMatrix.read("s3a://pan-ukb-us-east-1/ld_release/UKBB.EUR.ldadj.bm")
+                .filter(idx, idx)
+                .to_numpy()
+            )
+
+            alleleOrder = [
+                row["alleleOrder"]
+                for row in locus_index.select("alleleOrder").collect()
+            ]
+            outer_allele_order = np.outer(alleleOrder, alleleOrder)
+            np.fill_diagonal(outer_allele_order, 1)
+
+            ld_matrix = (half_matrix + half_matrix.T) - np.diag(np.diag(half_matrix))
+            ld_matrix = ld_matrix * outer_allele_order
+            np.fill_diagonal(ld_matrix, 1)
+        else:
+            ld_matrix = None
+
+        return ld_matrix
+
+    @staticmethod
+    def get_locus_index_boundaries(
+        session: Session,
+        study_locus_row: Row,
+        ancestry: str = "nfe",
+    ) -> DataFrame:
+        """Extract hail matrix index from StudyLocus rows.
+
+        Args:
+            session (Session): Session object
+            study_locus_row (Row): Study-locus row
+            ancestry (str): Major population to extract from gnomad matrix, default is "nfe"
+
+        Returns:
+            DataFrame: Returns the index of the gnomad matrix for the locus
+
+        """
+        chromosome = str("chr" + study_locus_row["chromosome"])
+        start = int(study_locus_row["locusStart"])
+        end = int(study_locus_row["locusEnd"])
+
+        if ancestry == "nfe":
+            index_file = session.spark.read.parquet(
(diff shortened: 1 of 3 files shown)


addramir:
All looks good to me.

src/gentropy/datasource/pan_ukbb_ld/ld.py
```diff
@@ -0,0 +1,214 @@
...
+            .filter(
+                (
+                    (f.col("referenceAllele") == f.col("va_ref"))
+                    & (f.col("alternateAllele") == f.col("va_alt"))
+                )
+                | (
+                    (f.col("referenceAllele") == f.col("va_alt"))
+                    & (f.col("alternateAllele") == f.col("va_ref"))
+                )
+                | (f.col("va_ref").isNull() | f.col("va_alt").isNull())
```
DSuveges, line 104: Is this a valid condition? Can the variant annotation have nulls in the alleles column?

src/gentropy/datasource/pan_ukbb_ld/ld.py
```diff
@@ -0,0 +1,214 @@
...
+                how="left",
+            )
+            .filter(
+                (
+                    (f.col("referenceAllele") == f.col("va_ref"))
+                    & (f.col("alternateAllele") == f.col("va_alt"))
+                )
+                | (
+                    (f.col("referenceAllele") == f.col("va_alt"))
+                    & (f.col("alternateAllele") == f.col("va_ref"))
```
DSuveges, line 102: If you are checking allele flipping, shouldn't the strandedness be checked as well? Eg. a `G` to `A` might be equivalent to a `T` to `C` change on the other strand.

src/gentropy/datasource/pan_ukbb_ld/ld.py
```diff
@@ -0,0 +1,214 @@
...
+            ld_populations (list[str]): List of populations
+            ukbb_annotation_path (str): Path to pan-ukbb variant LD index with alleles flipped to match the order in OT variant annotation
+        Default values are set in PanUKBBConfig.
+        """
+        self.pan_ukbb_ht_path = pan_ukbb_ht_path
+        self.pan_ukbb_bm_path = pan_ukbb_bm_path
+        self.ld_populations = ld_populations
+        self.ukbb_annotation_output_path = ukbb_annotation_path
+
+    def align_ld_index_alleles(
```
DSuveges, line 46: Where is this function used? I can't find any call.

DSuveges: Or is this the part of that process you mentioned @addramir , that there's no step for it, but the data is generated?

src/gentropy/datasource/pan_ukbb_ld/ld.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        window_spec = Window.partitionBy("idx").orderBy(f.col("alleleOrder").desc())
+        ht_va = (
+            ht_va.withColumn("rank", f.rank().over(window_spec))
+            .filter(f.col("rank") == 1)
+            .drop("rank")
+        )
+        ht_va.write.mode("overwrite").parquet(hail_table_output.format(POP=population))
+
+    def get_numpy_matrix(
```
DSuveges, line 149: In this function, `ancestry` variable is not accessed.

src/gentropy/datasource/pan_ukbb_ld/ld.py
```diff
@@ -0,0 +1,214 @@
...
+            locus_index (DataFrame): hail matrix variant index table
+            ancestry (str): Ancestry label
+
+        Returns:
+            np.ndarray: LD block matrix for the locus
+        """
+        idx = [row["idx"] for row in locus_index.select("idx").collect()]
+
+        half_matrix = (
+            BlockMatrix.read(self.pan_ukbb_bm_path.format(POP="EUR"))
```
DSuveges, line 166:
```suggestion
            BlockMatrix.read(self.pan_ukbb_bm_path.format(POP=ancestry))
```
I assume.

src/gentropy/method/ld_matrix_interface.py
```diff
@@ -0,0 +1,104 @@
...
+        Args:
+            session (Session): Session object
+            study_locus_row (Row): Study-locus row
+            ancestry (str): Major population to extract from gnomad matrix, default is "nfe"
+
+        Returns:
+            DataFrame: Returns the index of the gnomad matrix for the locus
+
+        """
+        mapped_ancestry = LDMatrixInterface.ancestry_map.get(ancestry, ancestry)
```
DSuveges, line 44: You don't really use the `mapped_ancestry` variable. It only used once, inside the condition. if the logic goes to the `else` branch it is not used, so there's no need to create. I would move inside the `if` branch.

src/gentropy/method/ld_matrix_interface.py
```diff
@@ -0,0 +1,104 @@
...
+
+        if (ancestry in ("nfe", "csa", "afr")) and (session is not None):
+            joined_index = PanUKBBLDMatrix().get_locus_index_boundaries(
+                session=session,
+                study_locus_row=study_locus_row,
+                ancestry=mapped_ancestry,
```
DSuveges, line 50:
```suggestion
                ancestry=LDMatrixInterface.ancestry_map.get(ancestry),
```
At this point we already know that these ancestries are directed to the PanUKBB path.

src/gentropy/datasource/pan_ukbb_ld/ld.py
Daniel-Considine: Yes - this is the capture the variants which are in Pan-UKBB, but not in our variant annotation at all (flipped or otherwise).

src/gentropy/datasource/pan_ukbb_ld/ld.py
Daniel-Considine: For this only needs to be run once for each Pan-UKBB ancestry, and then we store and use the resulting parquet DF for the other functions.

src/gentropy/datasource/pan_ukbb_ld/ld.py
Daniel-Considine: We don't have strand information in either dataset, so not able to check for this

Daniel-Considine:
> I only have small comments, questions. There's one place where the ancestry is not substituted so the returned ld index and the ld matrix could be inconsistent if a non EUR ancestry is used.
> 
> Assuming the nuances of the logic is heavily tested (according to @addramir that's the case) and this bit is fixed, I'll approve.

I think this should be all good now

DSuveges (approved):
Thanks addressing my questions. All looks good now.

PR #762, opened 2024-09-15
fix(effect harmonisation): addressing beta harmonisation bug 

DSuveges:
## ✨ Context

As reported on [#3461](https://github.com/opentargets/issues/issues/3461), there was an issue with how the effect of the GWAS Catalog curated associations were ingested. Essentially, OR were not distinguished from beta.

The effect harmonisation logic become a dataframe to dataframe function, where helper columns are added making the logic easier to understand and test.

- Updated GWAS Catalog ingestion logic to harmonise association effect.
- Some minor helper functions added.
- Some no longer used functions are removed.
- Adding tests for the touched functions.

## 🚦 Before submitting


src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -195,7 +194,7 @@ def _collect_rsids(
         return f.array_distinct(f.array(snp_id, snp_id_current, risk_allele))
 
     @staticmethod
-    def _map_variants_to_variant_index(
+    def _map_variants_to_gnomad_variants(
```
ireneisdoomed, line 197: This method could be abstracted from the type of variant index you pass. As far as I can see, the main difference is the application of `convert_gnomad_position_to_ensembl`.

Do you think this method could live in the gnomad parser? In fact it'd make sense to convert them before dumping the gnomad variants dataset. When we generate the variant index, we join the VEP output with the Gnomad dataset to get the AF. Aren't we joining 2 datasets with 2 diff coordiate systems?
Out of scope of the PR, just a comment

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -570,175 +569,107 @@ def _are_alleles_palindromic(
...
     ) -> Column:
-        """A function to extract the beta value from the effect size and confidence interval.
+        """A function to extract the beta value from the effect size and confidence interval and harmonises for the alternate allele.
 
         If the confidence interval contains the word "increase" or "decrease" it indicates, we are dealing with betas.
-        If it's "increase" and the effect size needs to be harmonized, then multiply the effect size by -1
+        If it's "increase" and the effect size needs to be harmonized, then multiply the effect size by -1.
+        The sign of the effect size is flipped if the confidence interval contains "decrease".
+
+        eg. if the reporeted value is 0.5, and the confidence interval tells "decrease"? -> beta is -0.5
```
ireneisdoomed, line 582:
```suggestion
        eg. if the reported value is 0.5, and the confidence interval tells "decrease"? -> beta is -0.5
```

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -988,11 +919,177 @@ def _qc_palindromic_alleles(
...
+            )
+            .withColumns(
+                {
+                    # Flag palindromic alleles:
+                    "isAllelePalindromic": GWASCatalogCuratedAssociationsParser._are_alleles_palindromic(
+                        f.col("referenceAllele"), f.col("alternateAllele")
+                    ),
+                    # Flag associations, where the effect direction needs to be flipped:
+                    "needsFlipping": GWASCatalogCuratedAssociationsParser._effect_needs_harmonisation(
+                        f.col("reporetedRiskAllele"), f.col("referenceAllele")
```
ireneisdoomed, line 1037:
```suggestion
                        f.col("reportedRiskAllele"), f.col("referenceAllele")
```

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -988,11 +919,177 @@ def _qc_palindromic_alleles(
...
+
+        for column in required_columns:
+            if column not in df.columns:
+                raise ValueError(
+                    f"Column {column} is required for harmonising effect to beta value."
+                )
+
+        return (
+            df.withColumn(
+                "reporetedRiskAllele",
```
ireneisdoomed, line 1024:
```suggestion
                "reportedRiskAllele",
```

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -570,175 +569,107 @@ def _are_alleles_palindromic(
...
         """
         return (
-            f.when(
-                GWASCatalogCuratedAssociationsParser._are_alleles_palindromic(
-                    reference_allele, alternate_allele
-                ),
-                None,
-            )
+            # We are not flipping zero effect size:
+            f.when((effect_size.cast(DoubleType()) == 0) & flipping_needed, f.lit(None))
```
ireneisdoomed, line 664: Might be missing something here, but isn't this function dealing with ORs? OR = 0 shows a protective effect.

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -988,11 +919,177 @@ def _qc_palindromic_alleles(
             StudyLocusQualityCheck.PALINDROMIC_ALLELE_FLAG,
         )
 
+    @staticmethod
+    def _get_effect_type(ci_text: Column) -> Column:
```
ireneisdoomed, line 923: Do you know more or less how many nulls we have here? Just to understand if the approach of using increase/decrease to distinguish works

ireneisdoomed (changes requested):
Thank you for fixing the issue!
I understood that the problem stemmed from the fact that we were treating all reported effects as betas. I think it looks ok, only have a question about the nullification of OR = 0. cases (see below)

src/gentropy/datasource/gwas_catalog/associations.py
DSuveges: OR=0 means no effect. Most likely we would not ever see association with 0 odds ratio. However, if that's happening, I would just keep it, but the effect cannot be flipped, because it would involve a division with zero.

src/gentropy/datasource/gwas_catalog/associations.py
DSuveges: I don't have strong feeling about this. Currently we are only calling the `convert_gnomad_position_to_ensembl` method here. I quite forgot what was happening in the variant index branch. If we need to touch that code, it can be re-designed.

src/gentropy/datasource/gwas_catalog/associations.py
DSuveges: Thanks for finding!

src/gentropy/datasource/gwas_catalog/associations.py
ireneisdoomed: That's what i wanted to know, thanks!

src/gentropy/datasource/gwas_catalog/associations.py
ireneisdoomed: Of course, didn't occur to me you were addressing division by 0 issues thanks! I agree that I can't see how a GWAS could have such effect size, however maybe it makes more sense mathematically speaking to just assign a very low value like 0.001 instead of a null?

ireneisdoomed (approved):
Thank you for the changes, the beta distribution looks good!

PR #764, opened 2024-09-16
feat: decision tree for FM in fine_mapper

Daniel-Considine:
## ✨ Context

Currently we aren't filtering out loci with too few loci prior to fine-mapping

After the join between summary statistics and LD index, if there are less than 100 variants an error is logged and the locus is not fine-mapped. 

## 🚦 Before submitting


addramir (changes requested):
I like the logic and I agree we should keep the filtering in function, not _init_ function as I suggested before.
1) These changes should be merged to new function with LD interface
2) We need to add the same case but for extremely large number of variants, I suggest 15K

PR #767, opened 2024-09-17
feat: flag MHC credible sets based on lead

d0choa:
This PR closes opentargets/issues#3469.

It implements:
- QC flag for MHC region in StudyLocus
- Add flag in study_locus_validation step
- New GenomicRegion class refactoring some pre-existing code

It's pending:
- Add new flag `IN_MHC` to orchestration config


d0choa:
@DSuveges and @Daniel-Considine, this touches on code that you worked on in the past, so please feel free to raise any objections. I'm assigning the review to Szymon and Yakov. (I didn't want to assign everyone 😅)

Daniel-Considine:
Looks all good to me

src/gentropy/study_locus_validation.py
```diff
@@ -41,9 +41,11 @@ def __init__(
...
         study_locus_with_qc = (
             StudyLocus.from_parquet(session, list(study_locus_path))
-            .validate_lead_pvalue(
-                pvalue_cutoff=gwas_significance
-            )  # Flagging study locus with subsignificant p-values
+            # Flagging study locus with subsignificant p-values
+            .validate_lead_pvalue(pvalue_cutoff=gwas_significance)
+            # Add flag for MHC region
+            .qc_MHC_region()
+            # Excluding MHC region
```
DSuveges, line 48: We are not excluding, just flagging at this point.

DSuveges (changes requested):
There's a comment in the `src/gentropy/study_locus_validation.py` file indicating a filtering for MHC, however I cannot find the logic to do so. Is it intentional? I suspect the comment should go, and a exclusion flag needs to be added to the orchestration.

src/gentropy/study_locus_validation.py
d0choa:
```suggestion
            # Flagging MHC region
```

d0choa:
- Comment updated
- As said at the top, we need to add `Add new flag IN_MHC to orchestration config`, and that would be enough to handle the filtering using the generic `valid_rows` function.

d0choa:
that's why I shouldn't do commits from the web. Fixed now

PR #768, opened 2024-09-17
docs: update broken refs

d0choa:
Some refs were broken and mkdocs was throwing WARNINGS. This fixes it.

We could activate the ["strict" mode](https://github.com/opentargets/gentropy/blob/6ede7363c683d0955edafc647c9d13b2baf17fc7/tests/gentropy/docs/test_build.py#L19) on the mkdocs test if we want to prevent this in the future. For some reason, I thought this was too harsh in the past but maybe we could reconsider. I'll leave it up to you @project-defiant

### DIFF
--- a/src/gentropy/dataset/summary_statistics.py
+++ b/src/gentropy/dataset/summary_statistics.py
@@ -90,7 +90,7 @@ class SummaryStatistics(Dataset):
     ) -> StudyLocus:
         """Generate study-locus from summary statistics using locus-breaker clumping method with locus boundaries.
 
-        For more info, see [`locus_breaker`][gentropy.method.locus_breaker_clumping.locus_breaker]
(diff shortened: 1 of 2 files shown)


project-defiant:
I have tested your solution to fail the tests when there is a warning thrown while building the documentation. For a longer term this might help, but we have to fix the one issue that persists on `git-committers` plugin locally. 
```
WARNING -  git-committers plugin may require a GitHub token if you exceed the API rate limit or for private repositories. Set it under 'token' mkdocs.yml config or MKDOCS_GIT_COMMITTERS_APIKEY environment variable
```
The warning is displayed, because the Plugin might require github token to access the github commit history as mentioned in [readme](https://github.com/ojacques/mkdocs-git-committers-plugin-2?tab=readme-ov-file#setup:~:text=Example%3A%20%27master%27%20(default)-,token,-%2D%20A%20GitHub%20or)

PR #769, opened 2024-09-18
feat: add biosample index

Tobi1kenobi:
## 🚦 Before submitting


Tobi1kenobi:
Few notes:

- When joining onto eQTL catalogue three `biosampleIds` have no match in the `biosample index`: BTO_0000930  (neuroblast), EFO_0005292 (lymphoblastoid cell line), EFO_0004905 (induced pluripotent stem cell). When thinking about the other use cases for this index some thought should go into the best way to capture these biosamples.
- Current no column for deprecated or not. Not super relevant for current purposes of left-joining and wasn't entirely clear how to implement (drop deprecated, flag them, etc)
- dbXrefs currently a flat list of strings rather than split into ID and as done for `variant index`. This was partially for simplicity, partially because I was unsure how best to split (e.g. CL_1001593 -> [CL_1001593, CL], [CL_1001593, Cell Ontology], [1001593, CL], [CL_1001593, Uberon (ontology actually sourced from)] and partially because I didn't see an immediate use case unlike `variant index`.

src/gentropy/assets/schemas/biosample_index.json
```diff
@@ -0,0 +1,83 @@
...
+    {
+      "name": "dbXrefs",
+      "type": {
+        "type": "array",
+        "elementType": "string",
+        "containsNull": true
+      },
+      "nullable": true,
+      "metadata": {}
+    },
```
DSuveges, line 31: This is not really good. We already have this column in other datasets eg. VariantIndex + others in the platform. The expectation from this filed is that it is a list of structs, where each element not only informs about the identifier, but also about the source.:

```json
    {
      "metadata": {},
      "name": "dbXrefs",
      "nullable": true,
      "type": {
        "containsNull": true,
        "elementType": {
          "fields": [
            {
              "metadata": {},
              "name": "id",
              "nullable": true,
              "type": "string"
            },
            {
              "metadata": {},
              "name": "source",
              "nullable": true,
              "type": "string"
            }
          ],
          "type": "struct"
        },
        "type": "array"
      }
    }
```

docs/python_api/datasources/_datasources.md
```diff
@@ -26,7 +26,7 @@ This section contains information about the data source harmonisation tools avai
 2. GWAS catalog's [harmonisation pipeline](https://www.ebi.ac.uk/gwas/docs/methods/summary-statistics#_harmonised_summary_statistics_data)
 3. Ensembl's [Variant Effect Predictor](https://www.ensembl.org/info/docs/tools/vep/index.html)
 
-## Linkage desiquilibrium
```
DSuveges, line 29: Great catch!

src/gentropy/assets/schemas/biosample_index.json
Tobi1kenobi: I agree, sorta... See my comment on the pull request. It wasn't immediately apparent how to structure it

src/gentropy/config.py
```diff
@@ -51,6 +51,15 @@ class GeneIndexConfig(StepConfig):
     _target_: str = "gentropy.gene_index.GeneIndexStep"
 
 
+@dataclass
+class BiosampleIndexConfig(StepConfig):
+    """Biosample index step configuration."""
+
+    target_path: str = MISSING
+    biosample_index_path: str = MISSING
+    _target_: str = "gentropy.biosample_index.BiosampleIndexStep"
```
DSuveges, line 60: The configuration should have all the inputs that the step needs. 

- `cell_ontology_input_path`
- `uberon_input_path`
- `biosample_index_output_path`

All can/should be missing in this case, because there's no hardcoded parameters eg. [r2 threshold for LD index](https://github.com/opentargets/gentropy/blob/c3651b7e8ecad3631fce0d4e9fa79fea1b59d40a/src/gentropy/config.py#L173). Similarly, `target_path` needs to be dropped as this parameter is not part of the `BiosampleIndexStep` parameters.

src/gentropy/datasource/biosample_ontologies/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    coalesce,
+    col,
+    collect_list,
+    collect_set,
+    explode_outer,
+    first,
+    regexp_replace,
+    udf,
+)
+from pyspark.sql.types import ArrayType, StringType
```
DSuveges, line 16: This is just stylistic comment, but we do not import spark functions like this. There's a not too serious explanation for this: some of the functions might overwrite Python base function eg. `from pyspark.sql.functions import sum, max`, that's why we import all functions with a prefix: 

```python
import pyspark.sql.functions as f

# then calling a function would look like this:
f.array_distinct(f.array_union(f.col('colname')....
```
It's quite marginal, but being stylistically consistent is good.

src/gentropy/datasource/biosample_ontologies/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    final_df = transformed_df.join(df_with_relationships, ["biosampleId"], "left")
+
+    return BiosampleIndex(
+        _df=final_df,
+        _schema=BiosampleIndex.get_schema()
+        )
+
+def merge_biosample_indices(
+    biosample_indices : list[BiosampleIndex]
+    ) -> BiosampleIndex:
```
DSuveges, line 147: I wouldn't add this function here. The merge of biosample indices is kind of an inherent ability of biosample indices as well. Eg.

```python
a: BiosampleIndex
b: BiosampleIndex
merged: BiosampleIndex = a.merge_indices(b)
```
Assuming the operation is commutative, and `b.merge(a)` yields the same index. 

So my advise is to move this logic as a instance method of the `BiosampleIndex` dataset.

DSuveges: (also this would imply one fewer function to import, as the merge function travels wherever the biosample object goes)

src/gentropy/datasource/biosample_ontologies/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    """Merge a list of biosample indices into a single biosample index.
+
+    Where there are conflicts, in single values - the first value is taken. In list values, the union of all values is taken.
```
DSuveges, line 150: So this function is not entirely commutative, but I think this is still fine.

src/gentropy/datasource/biosample_ontologies/utils.py
```diff
@@ -0,0 +1,197 @@
...
+
+    Args:
+        biosample_indices (list[BiosampleIndex]): Biosample indices to merge.
+
+    Returns:
+        BiosampleIndex: Merged biosample index.
+    """
+
+    def merge_lists(
```
DSuveges, line 159: It's a great function, nice elegant logic! A few comments:

- I would add a description explaining that the order of element are not kept (lists are ordered structures). 
- It returns with a unique set of elements, instead of all elements.
- I would put this function under `common.utils.py`, because such function might be useful in other places. 

Question: Should this function be generalised to flatten arbitrarily deeply nested lists? (via recursion?)

src/gentropy/datasource/biosample_ontologies/utils.py
```diff
@@ -0,0 +1,197 @@
...
+        Args:
+            lists (list[list[str]]): List of lists to merge.
+
+        Returns:
+            list[str]: Merged list.
+        """
+        return list({item for sublist in lists if sublist is not None for item in sublist})
+
+    # Make a spark udf (user defined function) to merge lists
+    merge_lists_udf = udf(merge_lists, ArrayType(StringType()))
```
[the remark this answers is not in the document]
DSuveges: If you need a unique set of elements from the flattened array you can use:
```python
f.array_distinct(f.flatten(f.col('array_of_array_column")))
```

src/gentropy/study_validation.py
```diff
@@ -63,6 +67,7 @@ def __init__(
             .validate_study_type()  # Flagging non-supported study types.
             .validate_target(target_index)  # Flagging QTL studies with invalid targets
             .validate_disease(disease_index)  # Flagging invalid EFOs
+            .validate_biosample(biosample_index)  # Flagging invalid biosamples
```
DSuveges, line 70:
```suggestion
            .validate_biosample(biosample_index)  # Flagging studies with invalid biosamples
```

src/gentropy/dataset/study_index.py
```diff
@@ -408,3 +411,36 @@ def validate_target(self: StudyIndex, target_index: GeneIndex) -> StudyIndex:
         )
 
         return StudyIndex(_df=validated_df, _schema=StudyIndex.get_schema())
+
+    def validate_biosample(self: StudyIndex, biosample_index: BiosampleIndex) -> StudyIndex:
+        """Validating biosample identifiers in the study index against the provided biosample index.
```
DSuveges, line 416: This is not a critique, rather a comment: we are doing exactly the same steps for genes (disease validation is more complicated). I'm wondering if we can one function that would be used by both the biosample and gene validation... eg. We would same, this is the list of fields you need to find in this column, and if you don't find, add this flag... Not exactly sure how to implement this, but sounds quite abstract that could be used in other places and other datasets maybe.

d0choa: After doing a few of these, the problem I see in abstracting this further is that the list of rules that decide when to add a flag is very diverse. Sometimes, it is just a filter, but sometimes, it has dependencies, aggregations, etc.

I'm referring to everything that leads to the second argument of `update_quality_flag` if this is what you mean here

DSuveges (approved):
Very nice PR! Congratulation to becoming a contributor in this amazing project! :D

PR #772, opened 2024-09-18
fix(ld clumping): a revised logic allows a more accurate clumping

DSuveges:
a revised logic allows a more accurate identification of linked loci

## ✨ Context

It was noticed that in certain cases the LD based clumping was not performing as expected. After reviewing the logic, a proposed new implementation addresses the issue.

## 🚦 Before submitting

### DIFF
--- a/.github/workflows/artifact.yml
+++ b/.github/workflows/artifact.yml
@@ -2,8 +2,8 @@ name: Build and Push to Artifact Registry
 
 "on":
   push:
-    branches: ["dev"]
-    tags: ["v*"]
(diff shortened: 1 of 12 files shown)


ireneisdoomed (approved):
Refactoring the function that identifies linkage between associations fixes the reported problem. The function is easier to understand too:
1. On every row, the list of more significant variants are collected so that in this array you only select the variants that are more significant than the current row
2. On every row, you compute the intersection between the result of the window and the variants in the credible set, to check whether there is more than one lead in the credible set.

PR #773, opened 2024-09-19
perf: quickly build a Docker image for every branch

tskir:
## ✨ Context
There are two problems with Docker builds:
* They take about 11 minutes per commit;
* They only happen for the `dev` branch and `v*` tags.

It is desirable to automatically build the image for other branches, so that a version in development could be run for example on Google Batch. However, with 11 minutes per commit it's not very convenient.

Add a step to quickly build a Docker image for every branch, which means:
* Only build the main gentropy image;
* Only do it for the AMD64 architecture required for Google cloud;
* Use maximum layer caching through GHA.

With this change, build takes around 1–2 minutes per commit (depending on whether dependencies were changed or not) instead of 11. This is now trivial to run for every commit of every branch.

The full builds, which include ARM64 infrastructure and the VEP image, are still only ran for the dev branch and v* tags, as before.

## 🙈 Missing
N/A.

## 🚦 Before submitting

### DIFF
--- a/.github/workflows/artifact.yml
+++ b/.github/workflows/artifact.yml
@@ -39,21 +39,21 @@ jobs:
         run: |-
           gcloud auth configure-docker ${{ env.REGION }}-docker.pkg.dev --quiet
 
-      # - name: Build and push gentropy image (AMD64)
-      #   uses: docker/build-push-action@v6
-      #   with:
-      #     platforms: linux/amd64
-      #     push: true
-      #     tags: "${{ env.GAR_LOCATION }}/${{ env.REPOSITORY }}/gentropy:${{ github.ref_name }}"
-      #     context: .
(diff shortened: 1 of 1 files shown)


project-defiant (approved):

d0choa:
It should be enough with @project-defiant review. We will need a cleanup strategy for the images

project-defiant:
The only note is that we could drop the image when the branch is dropped, so we do not pollute ourselves with too many images. But that can be addressed in the future when we hit this problem.

tskir:
Thank you for quick reviews @d0choa @project-defiant!

I'm going to go ahead and merge it.

As discussed, I'll leave setting up the clean-up policies up to you, this is something that seems to be supported: https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy

PR #775, opened 2024-09-20
feat: flag all top-hits from GWAS catalog curation

d0choa:
Closes opentargets/issues#3487

The only tricky decision here was where to place the QC flagging. I did it at the very end of the pipeline as `StudyLocusGWASCatalog` specific logic.


DSuveges (approved):
All makes sense

PR #777, opened 2024-09-20
feat: flag PICS top hits in studies with PICS from sumstats

d0choa:
Closes https://github.com/opentargets/issues/issues/3508


DSuveges (approved):
All makes sense, and looking good.

PR #780, opened 2024-09-20
feat: flag credible sets explained by SuSiE regions

d0choa:
Closes opentargets/issues#3464

More background about the last changes in the issue


src/gentropy/dataset/study_locus.py
```diff
@@ -920,6 +922,76 @@ def qc_redundant_top_hits_from_PICS(self: StudyLocus) -> StudyLocus:
...
+                "locusStart",
+                "locusEnd",
+                f.lit(True).alias("inSuSiE"),
+            )
+            .distinct()
+        )
+
+        # non SuSiE credible sets (studyLocusId) overlapping in any variant with SuSiE locus
+        redundant_study_locus = (
+            self.filter(~(f.col("finemappingMethod") == "SuSiE-inf"))
```
DSuveges, line 948: Why not `f.col("finemappingMethod") != "SuSiE-inf"`?

d0choa:
```suggestion
            self.filter(f.col("finemappingMethod") != "SuSiE-inf")
```

src/gentropy/dataset/study_locus.py
```diff
@@ -963,6 +965,76 @@ def qc_redundant_top_hits_from_PICS(self: StudyLocus) -> StudyLocus:
...
+            _df=(
+                self.df.join(redundant_study_locus, on="studyLocusId", how="left")
+                .withColumn(
+                    "qualityControls",
+                    self.update_quality_flag(
+                        f.col("qualityControls"),
+                        # credible set in SuSiE overlapping region
+                        f.col("inSuSiE")
+                        # credible set not based on SuSiE
+                        & ~(f.col("finemappingMethod") == "SuSiE-inf")
```
d0choa, line 1024:
```suggestion
                        & (f.col("finemappingMethod") != "SuSiE-inf")
```

src/gentropy/dataset/study_locus.py
```diff
@@ -963,6 +965,76 @@ def qc_redundant_top_hits_from_PICS(self: StudyLocus) -> StudyLocus:
...
```
DSuveges: I think this condition is not needed as the source dataset (`redundant_study_locus`) is already filtered.

PR #783, opened 2024-09-24
feat: change `StudyLocusId` hashing method to md5 (and change `StudyLocusId` to string type)

vivienho:
## ✨ Context

StudyLocusId is used as an identifier and does not need to be numerical. Changing it to string will make it easier on the backend side. Hashing strategy is changed to md5, which returns strings.

`StudyLocusId` is changed to string type in the schema and at relevant locations (mostly in tests).

The hashing strategy for generating the `StudyLocusId` is changed to md5.

A test (`test_assign_study_locus_id__null_variant_id`) was removed as validation steps elsewhere should have dropped null variant id cases before the `assign_study_locus_id` function.

## 🚦 Before submitting


DSuveges:
Given you were done so fast, I would recommend to update the logic to be more abstract allowing generalisation of the identifier generation (we will need this in other datasets, eg l2g). If you notice, you don't really need the arguments here:

```python
    def assign_study_locus_id(
        study_id_col: Column,
        variant_id_col: Column,
        finemapping_col: Column = None,
    ) -> Column:
```

Because you have this list:

```python 
columns = [study_id_col, variant_id_col, finemapping_col]
```
This can be a simple array of column names, which I believe should be a class class attribute for StudyLocus dataset. So the class itself would define what columns needs to be hashed for the identifier and in which order. Also, I think the actual hashing logic:

```python
 hashable_columns = [f.when(column.cast("string").isNull(), f.lit("None"))
                                 .otherwise(column.cast("string"))
                                 for column in columns]
        return f.md5(f.concat(*hashable_columns))
```
should be in the `Dataset` class. And the `assign_study_locus_id` method should wraps that function:

```python
    @staticmethod
    def assign_study_locus_id(
        study_id_col: Column,
        variant_id_col: Column,
        finemapping_col: Column = None,
    ) -> Column:
        self._generate_identifier(self.uniqueness_defining_columns).alias("studyLocusId")
```

Where:

- `_generate_identifier` is the hashing function in Dataset class, can be called from any dataset.
- `uniqueness_defining_columns` is a class attribute defined for the given dataset.
- This method returns the alias of the column, which is also dataset specific. 

I have a tendency to over abstract everything, so it would be great to have a second opinion on this from @d0choa .

src/gentropy/datasource/gwas_catalog/associations.py
```diff
@@ -1109,7 +1109,7 @@ def from_source(
         """
         return StudyLocusGWASCatalog(
             _df=gwas_associations.withColumn(
-                "studyLocusId", f.monotonically_increasing_id().cast(LongType())
+                "studyLocusId", f.monotonically_increasing_id().cast(StringType())
```
DSuveges, line 1112: This is not a dealbreaker, and has no impact whatsoever: this column is not a "real" `studyLocusId`: this column is temporarily generated to identify original rows of the GWAS Catalog association dataset before explosion. But it is fine.

DSuveges (approved):
Lot of changes, all looks great, let's hope nothing breaks. 🤞🏻

PR #786, opened 2024-09-24
fix: clean unused study_locus_validation step parameter

d0choa:
This cleanup was pending from #784


addramir:
We both make this fix in paralel. I implemented it here.
https://github.com/opentargets/gentropy/pull/788
Happy to go with your PR tho

addramir (approved):

PR #788, opened 2024-09-24
fix: adding data specific p-value filters

addramir:
## ✨ Context
Data specific lead p-value CS filters.

## 🚦 Before submitting


src/gentropy/finngen_finemapping_ingestion.py
```diff
@@ -37,6 +37,10 @@ def __init__(
             finngen_susie_finemapping_cs_summary_files=finngen_susie_finemapping_cs_summary_files,
         )
 
+        finngen_finemapping_df = finngen_finemapping_df.validate_lead_pvalue(
```
DSuveges, line 40: Minor stylistic comment, and it is absolutely my preference, but I like to use as few variables as possible:
```python
(
    # Reading Finngen finemapped dataset and convert it to study locus:
    FinnGenFinemapping.from_finngen_susie_finemapping(
        spark=session.spark,
        finngen_susie_finemapping_snp_files=finngen_susie_finemapping_snp_files,
        finngen_susie_finemapping_cs_summary_files=finngen_susie_finemapping_cs_summary_files,
    )
    # Flagging sub-significnat loci:
    .validate_lead_pvalue(
        pvalue_cutoff=FinngenFinemappingConfig().finngen_finemapping_lead_pvalue_threshold
    )
    # Write the output.
    .df.write.mode(session.write_mode).parquet(
        finngen_finemapping_out
    )
)
```

DSuveges (commented):
I might missing something, but could not found the flagging of the in-house finemapped datasets. Also you mentioned there would be specific p-value cutoff for UKBPPP.

src/gentropy/finngen_finemapping_ingestion.py
addramir: Please carefully check, I removed the variables but not sure

src/gentropy/config.py
```diff
@@ -169,6 +175,7 @@ class FinngenFinemappingConfig(StepConfig):
     _target_: str = (
         "gentropy.finngen_finemapping_ingestion.FinnGenFinemappingIngestionStep"
     )
+    finngen_finemapping_lead_pvalue_threshold: float = 1e-5
```
DSuveges, line 178: There's one thing I'm not sure about. You have added `finngen_finemapping_lead_pvalue_threshold` to the relevant config, and refer to as `pvalue_cutoff=FinngenFinemappingConfig().finngen_finemapping_lead_pvalue_threshold` in the step. However, `finngen_finemapping_lead_pvalue_threshold` it not an argument for `FinnGenFinemappingIngestionStep`. Is it OK? Would all parameters in the config passed to the step? Would that cause any problem? @project-defiant , what do you think?

DSuveges: OK, double-checked with @project-defiant and @d0choa and all parameters in the setepConfig classes needs to be parameters in the init function of the step.

src/gentropy/eqtl_catalogue.py
```diff
@@ -58,13 +59,21 @@ def __init__(
...
+            .parquet(eqtl_catalogue_study_index_out)
         )
-        credible_sets.df.write.mode(session.write_mode).parquet(
-            eqtl_catalogue_credible_sets_out
+
+        (
+            EqtlCatalogueFinemapping.from_susie_results(processed_susie_df)
+            # Flagging sub-significnat loci:
+            .validate_lead_pvalue(
+                pvalue_cutoff=EqtlCatalogueConfig().eqtl_lead_pvalue_threshold
```
DSuveges, line 74: Same comment: `EqtlCatalogueConfig().eqtl_lead_pvalue_threshold` has to be passed as parameter of the step.

src/gentropy/finngen_finemapping_ingestion.py
```diff
@@ -31,13 +31,17 @@ def __init__(
...
-            finngen_finemapping_out
+        (
+            FinnGenFinemapping.from_finngen_susie_finemapping(
+                spark=session.spark,
+                finngen_susie_finemapping_snp_files=finngen_susie_finemapping_snp_files,
+                finngen_susie_finemapping_cs_summary_files=finngen_susie_finemapping_cs_summary_files,
+            )
+            # Flagging sub-significnat loci:
+            .validate_lead_pvalue(
+                pvalue_cutoff=FinngenFinemappingConfig().finngen_finemapping_lead_pvalue_threshold
```
DSuveges, line 42: Same comment: `FinngenFinemappingConfig().finngen_finemapping_lead_pvalue_threshold` has to be an init parameter of the step.

src/gentropy/pics.py
```diff
@@ -28,8 +29,14 @@ def __init__(
...
         )
         # PICS
-        picsed_sl = PICS.finemap(study_locus_ld_annotated).filter_credible_set(
-            credible_interval=CredibleInterval.IS99
+        (
+            PICS.finemap(study_locus_ld_annotated)
+            .filter_credible_set(credible_interval=CredibleInterval.IS99)
+            # Flagging sub-significnat loci:
+            .validate_lead_pvalue(
+                pvalue_cutoff=WindowBasedClumpingStepConfig().gwas_significance
```
DSuveges, line 37: Same comment.

DSuveges (changes requested):
Any parameter that are defined in the StepConfigs needs to be defined as parameters of the init methods in the respective classes otherwise the step would fail. The reason is that the step object would be initialised with an unexpected parameter.

addramir:
> Any parameter that are defined in the StepConfigs needs to be defined as parameters of the init methods in the respective classes otherwise the step would fail. The reason is that the step object would be initialised with an unexpected parameter.

Fixed. Please check

DSuveges (approved):
This PR as it is now, makes it impossible to parametrise the p-value threshold of the PICS step. IF other steps are called as a stand-alone command, users can provide the custom p-value threshold. However this is not true for PICS step. The applied cutoff is defined by config (`WindowBasedClumpingStepConfig().gwas_significance`), which value cannot be overridden. 

I think this might be fine for now, but we should follow the pattern of other steps where all parameters inside the steps are customisable.

PR #790, opened 2024-09-25
fix(schema): recursive validation of arbitrarily deep nested structure

DSuveges:
## ✨ Context

So far the validation of datasets were not fully bullet proof: deeply nested (array of array) fields were not validated and automatically passed the validation. This was an issue, because the VEP parser had a bug yielding inSilicoPredictor column with a schema array of array of struct instead of array of struct. The validation was passing, tests were passing, and the bug was discovered too late. 

- New, recursive schema comparison methods for arrays and structs.
- Updated schema validation method in the Dataset class relying on the new comparing methods.
- New SchemaValidationError class that is called upon failing validation.
- The old schema flattening methods are kept because it might have other usages besides the schema validation.
- Tests for the new schema comparison functions.
- Temporarily skipping VEP parser tests.


d0choa:
🤯 WOW

project-defiant:
@DSuveges If we transition to pyspark 3.5 this feature could be easly ported from https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.testing.assertSchemaEqual.html

project-defiant:
@DSuveges unfortunately there is another issue with the schema validation 
```python
def test_schema_is_correct(spark: SparkSession) -> None:
    """Test schema is correct."""
    from gentropy.common.schemas import compare_struct_schemas

    # fmt: off
    schema = t.StructType(
        [
            t.StructField(
                "arr",
                t.ArrayType(
                    t.StructType(
                        [
                            t.StructField("a", t.StringType()),
                            t.StructField("b", t.IntegerType())
                        ]
                    )
                )
            ),
            t.StructField("id", t.IntegerType())
        ]
    )
    schema2 = t.StructType(
        [
            t.StructField(
                "arr2",
                t.ArrayType(
                    t.StructType(
                        [
                            t.StructField("b", t.IntegerType()),
                            t.StructField("a", t.StringType())
                        ]
                    )
                )
            ),
            t.StructField("id", t.IntegerType())
        ]
    )
    df1 = spark.createDataFrame([([("a", 1,)], 1),], schema=schema)
    df2 = spark.createDataFrame([([(1,"a",)], 1),], schema=schema2)
    diff = compare_struct_schemas(
        observed_schema=df1.schema,
        expected_schema=df2.schema,
    )
    assert len(diff) != 0
```

The above test does not fail, although the schema is almost the same, but the nested struct fields are not ordered correctly. This is the issue the variant index schema is facing now as well.

```python
vep_output_json_path=  "gs://ot_orchestration/releases/26.09/variants/annotated_variants"
variant_index_path = "gs://ot_orchestration/releases/26.09/variant_index"
gnomad_variant_annotations_path = "gs://genetics_etl_python_playground/static_assets/gnomad_variants"
hash_threshold = 300
from gentropy.dataset.variant_index import VariantIndex
from gentropy.datasource.ensembl.vep_parser import VariantEffectPredictorParser
variant_index = VariantEffectPredictorParser.extract_variant_index_from_vep(
        session.spark, vep_output_json_path, hash_threshold
    )


annotations = VariantIndex.from_parquet(
    session=session,
    path=gnomad_variant_annotations_path,
    recursiveFileLookup=True,
)
variant_index.df.printSchema()
annotations.df.printSchema()
```
results in 
```
root
 |-- variantId: string (nullable = false)
 |-- chromosome: string (nullable = true)
 |-- position: integer (nullable = true)
 |-- referenceAllele: string (nullable = true)
 |-- alternateAllele: string (nullable = true)
 |-- inSilicoPredictors: array (nullable = false)
 |    |-- element: struct (containsNull = true)
 |    |    |-- method: string (nullable = true)
 |    |    |-- assessment: string (nullable = true)
 |    |    |-- score: float (nullable = true)
 |    |    |-- assessmentFlag: string (nullable = true)
 |    |    |-- targetId: string (nullable = true)
 |-- mostSevereConsequenceId: string (nullable = true)
 |-- hgvsId: string (nullable = true)
 |-- transcriptConsequences: array (nullable = true)
 |    |-- element: struct (containsNull = false)
 |    |    |-- variantFunctionalConsequenceIds: array (nullable = true)
 |    |    |    |-- element: string (containsNull = true)
 |    |    |-- consequenceScore: float (nullable = true)
 |    |    |-- aminoAcidChange: string (nullable = true)
 |    |    |-- uniprotAccessions: array (nullable = true)
 |    |    |    |-- element: string (containsNull = true)
 |    |    |-- isEnsemblCanonical: boolean (nullable = false)
 |    |    |-- codons: string (nullable = true)
 |    |    |-- distanceFromFootprint: long (nullable = true)
 |    |    |-- distanceFromTss: long (nullable = true)
 |    |    |-- appris: string (nullable = true)
 |    |    |-- maneSelect: string (nullable = true)
 |    |    |-- targetId: string (nullable = true)
 |    |    |-- impact: string (nullable = true)
 |    |    |-- lofteePrediction: string (nullable = true)
 |    |    |-- siftPrediction: float (nullable = true)
 |    |    |-- polyphenPrediction: float (nullable = true)
 |    |    |-- transcriptId: string (nullable = true)
 |    |    |-- transcriptIndex: integer (nullable = false)
 |-- rsIds: array (nullable = true)
 |    |-- element: string (containsNull = true)
 |-- alleleFrequencies: array (nullable = false)
 |    |-- element: struct (containsNull = true)
 |    |    |-- populationName: string (nullable = true)
 |    |    |-- alleleFrequency: double (nullable = true)
 |-- dbXrefs: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- id: string (nullable = true)
 |    |    |-- source: string (nullable = true)

root
 |-- variantId: string (nullable = true)
 |-- chromosome: string (nullable = true)
 |-- position: integer (nullable = true)
 |-- referenceAllele: string (nullable = true)
 |-- alternateAllele: string (nullable = true)
 |-- inSilicoPredictors: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- method: string (nullable = true)
 |    |    |-- assessment: string (nullable = true)
 |    |    |-- score: float (nullable = true)
 |    |    |-- assessmentFlag: string (nullable = true)
 |    |    |-- targetId: string (nullable = true)
 |-- mostSevereConsequenceId: string (nullable = true)
 |-- transcriptConsequences: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- variantFunctionalConsequenceIds: array (nullable = true)
 |    |    |    |-- element: string (containsNull = true)
 |    |    |-- aminoAcidChange: string (nullable = true)
 |    |    |-- uniprotAccessions: array (nullable = true)
 |    |    |    |-- element: string (containsNull = true)
 |    |    |-- isEnsemblCanonical: boolean (nullable = true)
 |    |    |-- codons: string (nullable = true)
 |    |    |-- distanceFromFootprint: long (nullable = true)
 |    |    |-- distanceFromTss: long (nullable = true)
 |    |    |-- appris: string (nullable = true)
 |    |    |-- maneSelect: string (nullable = true)
 |    |    |-- targetId: string (nullable = true)
 |    |    |-- impact: string (nullable = true)
 |    |    |-- lofteePrediction: string (nullable = true)
 |    |    |-- siftPrediction: float (nullable = true)
 |    |    |-- polyphenPrediction: float (nullable = true)
 |    |    |-- consequenceScore: float (nullable = true)
 |    |    |-- transcriptIndex: integer (nullable = true)
 |    |    |-- transcriptId: string (nullable = true)
 |-- rsIds: array (nullable = true)
 |    |-- element: string (containsNull = true)
 |-- hgvsId: string (nullable = true)
 |-- alleleFrequencies: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- populationName: string (nullable = true)
 |    |    |-- alleleFrequency: double (nullable = true)
 |-- dbXrefs: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- id: string (nullable = true)
 |    |    |-- source: string (nullable = true)
```

DSuveges:
I most likely misunderstand something, but when you say 

> The above test does not fail, although the schema is almost the same, but the nested struct fields are not ordered correctly. This is the issue the variant index schema is facing now as well.

Of course it doesn't fail, because you are testing for finding difference and the test indeed finds the difference (because the schemas are different), so it passes. 

```python
# What does the comparison returns:
diff
```

contains:

```
defaultdict(list, {'unexpected_columns': ['arr']})
```

Which indicates there's a difference in the schema. So the assertion `assert len(diff) != 0` falls to True.

project-defiant:
> ```python
> schema = t.StructType(
>         [
>             t.StructField(
>                 "arr",
>                 t.ArrayType(
>                     t.StructType(
>                         [
>                             t.StructField("a", t.StringType()),
>                             t.StructField("b", t.IntegerType())
>                         ]
>                     )
>                 )
>             ),
>             t.StructField("id", t.IntegerType())
>         ]
>     )
>     schema2 = t.StructType(
>         [
>             t.StructField(
>                 "arr2",
>                 t.ArrayType(
>                     t.StructType(
>                         [
>                             t.StructField("b", t.IntegerType()),
>                             t.StructField("a", t.StringType())
>                         ]
>                     )
>                 )
>             ),
>             t.StructField("id", t.IntegerType())
>         ]
>     )
>     df1 = spark.createDataFrame([([("a", 1,)], 1),], schema=schema)
>     df2 = spark.createDataFrame([([(1,"a",)], 1),], schema=schema2)
>     diff = compare_struct_schemas(
>         observed_schema=df1.schema,
>         expected_schema=df2.schema,
> ```

You are right. there should be `arr` columns in both, I have fixed the comment

DSuveges:
As discussed offline:

- The schema validation doesn't take the order of fields of struct into account.
- Most application this doesn't matter, however you can't merge two arrays of structs if the order of the fields are not matching.
- Proposed solution is to extend the `safe_array_sort` function with a piece of extra logic that sorts fields alphabetically.

d0choa:
Try to get out of this with the minimum number of scars 😅

DSuveges:
@d0choa we can just drop the entire branch bump to pyspark 3.5 and hope for the best.

project-defiant:
I am almost done with the fix in safe array union.

src/gentropy/common/schemas.py
```diff
@@ -53,14 +82,170 @@ def flatten_schema(schema: t.StructType, prefix: str = "") -> list[Any]:
...
+    elif observed_type == "struct":
+        schema_issues = compare_struct_schemas(
+            observed_schema.elementType,
+            expected_schema.elementType,
+            f"{parent_field_name}[].",
+            schema_issues,
+        )
+
+    # If element type is an array, resolve nesting:
+    elif observed_type == "array":
```
ireneisdoomed, line 141: To enforce both types are the same
```suggestion
    elif observed_type == "array" and expected_type == "array":
```

src/gentropy/common/schemas.py
```diff
@@ -53,14 +82,170 @@ def flatten_schema(schema: t.StructType, prefix: str = "") -> list[Any]:
...
+    # If element types are not matching, no further tests are needed:
+    if observed_type != expected_type:
+        schema_issues["columns_with_non_matching_type"].append(
+            f'For column "{parent_field_name}[]" found {observed_type} instead of {expected_type}'
+        )
+
+    # If element type is a struct, resolve nesting:
+    elif observed_type == "struct":
+        schema_issues = compare_struct_schemas(
+            observed_schema.elementType,
```
ireneisdoomed, line 134: My mypy is raising an issue here because both the `observed_schema` and `expected_schema` are technically of type `DataType`, whereas the parameters have to be `StructType`. Are you having issues as well? I think mypy would be able to resolve it if the conditional above was made using the elementType, not the name of the type, i.e. `elif observed_schema.elementType == StructType()`. Same applies below when you call `compare_array_schemas`

src/gentropy/common/schemas.py
```diff
@@ -53,14 +82,170 @@ def flatten_schema(schema: t.StructType, prefix: str = "") -> list[Any]:
...
+
+    if parent_field_name is None:
+        parent_field_name = ""
+
+    # Flagging duplicated columns if present:
+    if duplicated_columns := list(
+        {
+            f"{parent_field_name}{field.name}"
+            for field in observed_schema
+            if list(observed_schema).count(field) > 1
```
ireneisdoomed, line 193: Really nice

src/gentropy/common/schemas.py
```diff
@@ -1,36 +1,65 @@
...
     """
     core_schema = json.loads(
         pkg_resources.read_text(schemas, schema_json, encoding="utf-8")
     )
-    return t.StructType.fromJson(core_schema)
+    return StructType.fromJson(core_schema)
 
 
-def flatten_schema(schema: t.StructType, prefix: str = "") -> list[Any]:
+def flatten_schema(schema: StructType, prefix: str = "") -> list[Any]:
```
ireneisdoomed, line 58: Since we are now parsing each schema without flattening, I'd suggest removing this function

ireneisdoomed (approved):
This is great! The schema validation has become much more accurate to the actual structure (no flattening). And at the same time the implementation is easier to understand. Great testing suite too!

src/gentropy/common/schemas.py
```diff
@@ -1,36 +1,65 @@
...
+
+        Args:
+            message (str): The message to be displayed.
+            errors (defaultdict[str, list[str]]): The collection of observed discrepancies
+        """
+        super().__init__(message)
+        self.message = message  # Explicitly set the message attribute
+        self.errors = errors
+
+    def __str__(self: SchemaValidationError) -> str:
```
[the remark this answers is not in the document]
DSuveges: Yes, exactly.

src/gentropy/common/schemas.py
```diff
@@ -53,14 +82,170 @@ def flatten_schema(schema: t.StructType, prefix: str = "") -> list[Any]:
...
+    expected_type = expected_schema.elementType.typeName()
+
+    # If element types are not matching, no further tests are needed:
+    if observed_type != expected_type:
+        schema_issues["columns_with_non_matching_type"].append(
+            f'For column "{parent_field_name}[]" found {observed_type} instead of {expected_type}'
+        )
+
+    # If element type is a struct, resolve nesting:
+    elif observed_type == "struct":
```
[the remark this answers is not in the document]
DSuveges: I agree, explicit is better than implicit. (the equality of the two schemas were tested already)

src/gentropy/common/schemas.py
DSuveges: I can drop it if you think this function would not be useful in other context.

PR #794, opened 2024-09-27
feat: testing dataset 

DSuveges:
# !! Don't merge, just an experiment

Branching out from #783 In this experiment I was trying to do two things: 

1. Automate ID generation.
2. Resolve initialising datasets without providing the schema.
3. The id column is also abstracted, can be referred in joins without typing the column name. 

It seems we need to have a dataset specific post_init function that does some magic. Eg.

```python
from gentropy.dataset.test import TestClass


spark = SparkSession.builder.getOrCreate()


dataset = spark.createDataFrame([('a', 1, 1.0, 'b'),], ['a', 'b', 'c', 'd'])

print('Dataset before:')
dataset.show()


print('Dataset after:')
TestClass(dataset).df.show(truncate=False)


print('Dataset with id before:')
dataset = spark.createDataFrame([('a', 1, 2.5, 'd', 'asdf')], ['a', 'b', 'c', 'd', 'testId'])
dataset.show()

print('Dataset with id after:')
TestClass(dataset).df.show(truncate=False)
```

Outputs:

```
+---+---+---+---+
|  a|  b|  c|  d|
+---+---+---+---+
|  a|  1|1.0|  b|
+---+---+---+---+

Dataset after:
+---+---+---+---+--------------------------------+
|a  |b  |c  |d  |testId                          |
+---+---+---+---+--------------------------------+
|a  |1  |1.0|b  |75ed355ca18d8f03adfe5160dde9bff1|
+---+---+---+---+--------------------------------+

Dataset with id before:
+---+---+---+---+------+
|  a|  b|  c|  d|testId|
+---+---+---+---+------+
|  a|  1|2.5|  d|  asdf|
+---+---+---+---+------+

Dataset with id after:
+---+---+---+---+------+
|a  |b  |c  |d  |testId|
+---+---+---+---+------+
|a  |1  |2.5|d  |asdf  |
+---+---+---+---+------+
```

**There are some problem though:**
1. mypy is not happy about not providing  the `_unique_fields` and`_id_column` fields in the parent class. I see the point, however things becoming uglier if these are optional. (as I saw you need to define these values in the post init as well) 
2. Also, the id generation relies on the presence of all unique fields, which is not the case of the result of the some operation on study loci (eg. after window based clumping of summary statistics, we don't have finemapping method. It can be circumvented by making all these columns mandatory or having separate study loci and credible set dataclasses. 

What do you think @ireneisdoomed , @project-defiant, @vivienho, @d0choa ?


src/gentropy/dataset/test.py
```diff
@@ -0,0 +1,55 @@
...
+
+@dataclass
+class TestClass(Dataset):
+    """Plain test class demonstrating the use of automatic  ID generation and instantiation without schema."""
+
+    # This list defines the uniqueness of the dataset - these have to be mandatory fields:
+    _unique_fields = ["a", "b", "c"]
+    _id_column = "testId"
+
+    def __post_init__(self: Dataset) -> None:
```
project-defiant, line 26: Why doing it in the child class?

This could be set in the parent class __post_init__

In theory you overwrite the parameters to the `generate_identifier` that is called in the Dataset.__post_init__.
The method should infer three things:
1. weather the dataset requires the index field
2. index field name
3. fields that are used to build the index field.

These 3 things should be defined in child classes, in Dataset class they should be dummy field placeholders.

The only thing to implement then is to run the `super().__post_init__` in case we overwrite the __post_init_ directly in the child class.

src/gentropy/dataset/test.py
```diff
@@ -0,0 +1,55 @@
...
+        """Class specific post-init, overwriting the parent method."""
+        # Upon initialising, we assign schema - this is based on the class method, so it would work without instantiating the class:
+        self._schema = self.get_schema()
+
+        # Only calculating the identifier if it is not present in the dataframe already:
+        if "testId" not in self._df.columns:
+            self._df = self._df.withColumn(
+                self._id_column, self._generate_identifier(self._unique_fields)
```
project-defiant, line 34: We can call the `_generate_identifier` in the Dataset.__post_init__ with the fields that are inherited from the child class. Then the only things to consider are:
1. if dataset requires the index field
2. the index field name
3. the fields that build index field
Defining the above in the child and will result in useage of them, but returning to the parent class __post_init__ to run the method itself to generate the index.

DSuveges: But we don't need identifier in every dataset. We do, however want a list of columns defining uniqueness though.

DSuveges:
Ideas discussed with @d0choa , there's a more sensible implementation dropping dataclasses, which in this case makes the initialisation quite a pain.

PR #801, opened 2024-10-01
feat(validation): adding credible set confidence annotation at validation time

DSuveges:
## ✨ Context

This PR introduces credible set confidence classes, which is assigned to credible sets upon validation. The basis of the assignment is the fine-mapping method and the presence of certain quality control flags. This is a preliminary implementation, we can decide to make the classes more granular.


d0choa (approved):
Changes look good. If you have time to add at least a return_type unit test for `assign_confidence` it could be good to make sure that the function at least runs

PR #802, opened 2024-10-01
chore: drop redundant parameter from sumstat qc step

project-defiant:
## ✨ Context

Currently `study_id` is a required parameter for running the `summary_statistics_qc` step. This is not ideal, as when running in bulk, we can not specify which study_id should be used to create the qc output path. 

- [x] Removal of `study_id` parameter from the `summary_statistics_qc` step
- [x] Removal of static `/qc_results_` that is appended in the step output

The path to QC should be fully configurable, without it's part being hardcoded.
 

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -439,7 +439,6 @@ class GWASQCStep(StepConfig):
 
     gwas_path: str = MISSING
     output_path: str = MISSING
-    studyid: str = MISSING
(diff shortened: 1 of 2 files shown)


DSuveges (approved):
All makes sense. We no longer need a study identifier, the output path is directly given instead of being constructed with studyId.

PR #807, opened 2024-10-02
fix: empty inSilicoPredictors object in GnomAD variant index

DSuveges:
## ✨ Context

The way we could extract _in-silico_ predictors from the hail table of GnomAD variants, lead to empty rows, where no predicted scores were available. Upon creation, a filter is applied to drop these variants. This PR has no effect on schema, it is purely motivated by the desire to making the variant index dataset cleaner.


DSuveges:
The branch runs, GnomAD variant index generated here: `gs://gnomad_data_2/gnomad_variant_index`. By looking at the `inSilicoPredictors` column, there's no null scores.

ireneisdoomed (commented):
The logic is OK, but maybe I am missing context. I thought these were coming from VEP directly, not from gnomAD?

ireneisdoomed (approved):
Clear, thanks!

PR #812, opened 2024-10-03
feat: add interval logic for l2g features

xyg123:
## ✨ Context

Adding interval based features to the l2g model, based on the feature list.
https://github.com/opentargets/issues/issues/3512

- Implementation of PCHIC-based interval features for the L2G gene prediction model.
- Added back interval processing steps into the L2G feature generation step.

## 🙈 Missing

More features from anderson + thurman.

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -276,6 +276,7 @@ class LocusToGeneConfig(StepConfig):
     wandb_run_name: str | None = None
     hf_hub_repo_id: str | None = "opentargets/locus_to_gene"
     download_from_hub: bool = True
+    #interval_sources = dict[str, str] | None = ["javierre": "gs://genetics_etl_python_playground/static_assets/javierre_2016_preprocessed", "thurman": "gs://genetics_etl_python_playground/static_assets/thurman_2012/genomewideCorrs_above0.7_promoterPlusMinus500kb_withGeneNames_32celltypeCategories.bed8.gz", "anderson":"gs://genetics_etl_python_playground/static_assets/andersson2014/enhancer_tss_associations.bed"]
     write_feature_matrix: bool = True
     _target_: str = "gentropy.l2g.LocusToGeneStep"
 
--- a/src/gentropy/dataset/l2g_features/intervals.py
+++ b/src/gentropy/dataset/l2g_features/intervals.py
@@ -1,0 +1,186 @@
+"""Collection of methods that extract features from the interval datasets."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import pyspark.sql.functions as f
+from pyspark.sql import Window
+
+from gentropy.common.spark_helpers import convert_from_wide_to_long
+
+# from gentropy.dataset.colocalisation import Colocalisation
+from gentropy.dataset.intervals import Intervals
+from gentropy.dataset.l2g_features.l2g_feature import L2GFeature
+from gentropy.dataset.l2g_gold_standard import L2GGoldStandard
+
+# from gentropy.dataset.study_index import StudyIndex
+from gentropy.dataset.study_locus import StudyLocus
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame
+
+
+def common_interval_feature_logic(
+    study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+    *,
+    intervals: Intervals,
+    feature_name: str,
+    interval_source: str,
+) -> DataFrame:
+    """Computes the feature.
+
+    Args:
+        study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci
+            that will be used for annotation
+        intervals (Intervals): The dataset containing interval information
+        feature_name (str): The name of the feature
+        interval_source (str): The datasource of the interval input
+
+    Returns:
+            DataFrame: Feature dataset
+    """
+    # Only implementing mean average interval features.
+    agg_expr = f.mean(f.col("weightedIntervalScore"))
+    return (
+        study_loci_to_annotate.df.withColumn("variantInLocus", f.explode_outer("locus"))
+        .select(
+            "studyLocusId",
+            f.col("variantInLocus.variantId").alias("variantInLocusId"),
+            f.col("variantInLocus.posteriorProbability").alias(
+                "variantInLocusPosteriorProbability"
+            ),
+        )
+        .join(
+            intervals.df.filter(f.col("datasourceId") == interval_source)
+            .withColumnRenamed("variantId", "variantInLocusId")
+            .withColumnRenamed("targetId", "geneId"),
+            on=["variantInLocusId", "geneId"],
+            how="inner",
+        )
+        .withColumn(
+            "weightedIntervalScore",
+            f.col("resourceScore") * f.col("variantInLocusPosteriorProbability"),
+        )
+        .groupBy("studyLocusId", "geneId")
+        .agg(agg_expr.alias(feature_name))
+    )
+
+
+def common_neighbourhood_interval_feature_logic(
+    study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+    *,
+    intervals: Intervals,
+    feature_name: str,
+    interval_source: str,
+) -> DataFrame:
+    """Computes the feature.
+
+    Args:
+        study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci that will be used for annotation
+        intervals (Intervals): The dataset containing interval information
+        feature_name (str): The name of the feature
+        interval_source (str): The datasource of the interval input
+
+    Returns:
+            DataFrame: Feature dataset
+    """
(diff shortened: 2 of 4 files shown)


src/gentropy/dataset/l2g_features/intervals.py
```diff
@@ -0,0 +1,186 @@
...
+        feature_name=local_feature_name,
+        intervals=intervals,
+        interval_source=interval_source,
+    )
+    return (
+        # Then compute the max score in the vicinity (
+        # feature will be the same for any gene associated with a studyLocus)
+        local_max.withColumn(
+            "regional_maximum",
+            f.max(local_feature_name).over(Window.partitionBy("studyLocusId")),
```
addramir, line 101: Why is it maximum? According to the table and what we discussed it should be mean?
https://docs.google.com/spreadsheets/d/1wUs1AprRCCGItZmgDhc1fF5BtwCSosdzFv4NQ8V6Dtg/edit?gid=452826388#gid=452826388

ireneisdoomed (changes requested):
Thank you for the changes Jack!!!

The logic to build the features looks good! Please see my comments, but they are more along the lines of how we process the interval data in the L2G step.
I suggested processing all interval sources to make the process simpler, but since the code is accommodated to take source names and paths individually and changing it is a mess, it's also fine to leave it like that as long as the interval_paths parameter is correctly configured.

The implemented changes wouldn't run, because of the creation of a Interval dataset with a mismatching schema. I would encourage you to:
- add any features you add to the `test_l2g_feature_matrix.py` suite, to make sure that the code doesnt crash
- In the same file, add a semantic test for the common logic
- Update the documentation pages

src/gentropy/l2g.py
```diff
@@ -104,6 +113,54 @@ def __init__(
...
+                    on=[
+                        f.col("vi.vi_chromosome") == f.col("interval.chromosome"),
+                        f.col("vi.position").between(
+                            f.col("interval.start"), f.col("interval.end")
+                        ),
+                    ],
+                    how="inner",
+                )
+                .drop("start", "end", "vi_chromosome", "position"),
+                _schema=Intervals.get_schema(),
```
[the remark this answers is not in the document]
xyg123: Yes, so I moved this into a function to overlap with variant index in interval datasets, and I included "variantId" into the interval schema, is this the easiest way to address this without making a new v2g style dataset?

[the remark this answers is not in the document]
ireneisdoomed: I don't think it is ideal, as the unit of this dataset would be a variant instead of an interval. Would collect the variants into a list work?

xyg123:
We have investigated the Intervals-only V2G dataset, the problem is that one variant can contain interval information from multiple genes (up to 200) from one interval source, in addition, the Interval-only V2G dataset is too big, and can potentially be 4x the size of the variant index.  

Therefore, it is not feasible to include interval data inside the variant index, instead, the "processed interval" dataset (requires gene index), will be generated each release, and the feature generation step will intersect variants found within studyLocus to the "processed interval" using an overlap approach to generate the scores needed for the interval features.

src/gentropy/assets/schemas/intervals.json
```diff
@@ -18,10 +18,16 @@
...
+      "metadata": {},
+      "name": "variantId",
+      "nullable": true,
+      "type": "string"
+    },
     {
       "metadata": {},
       "name": "geneId",
-      "nullable": false,
+      "nullable": true,
```
ireneisdoomed, line 30: Why nullable?

ireneisdoomed (changes requested):
Thank you for your additions! It looks like the code is not yet ready so please look at my comments. 

In general, I think that before integrating anything into dev, we should retrain the model and make a more informed decision about what and how we want to add it.

Like, it's a naive q, but is there a way to aggregate this data so that the feature represents if the credible set is part of any regulatory region that we know is physically close to the gene? As opposed to having separate features per source.

Based on the code, I guess you have agreed that Intervals will be written to disk and then the features will be extracted in the FM generation.

Correct me if I'm wrong, but based on your comment (https://github.com/opentargets/gentropy/pull/812#issuecomment-2545523871) the heavy part of processing interval data is not so much parsing the raw data, but getting the V2G relationships because of the high data volume. Have you considered doing the Interval parsing as part of the FM generation step?

In essence, I'd like to see the impact of these features before making a decision

src/gentropy/dataset/intervals.py
```diff
@@ -71,3 +73,34 @@ def from_source(
...
+
+        Args:
+            variant_index (VariantIndex): Variant index dataset
+
+        Returns:
+            Intervals: Variant-to-gene intervals dataset
+        """
+        return Intervals(
+            _df=(
+                self.df.alias("interval")
```
project-defiant, line 90: Given the intervals are not big datasets (my assumption) is there a possibility to broadcast them before the join?

src/gentropy/dataset/intervals.py
```diff
@@ -71,3 +73,34 @@ def from_source(
...
+                .join(
+                    variant_index.df.selectExpr(
+                        "chromosome as vi_chromosome", "variantId", "position"
+                    ).alias("vi"),
+                    on=[
+                        f.col("vi.vi_chromosome") == f.col("interval.chromosome"),
+                        f.col("vi.position").between(
```
project-defiant, line 97: This is not the equi join, so the join will be less optimized.

src/gentropy/l2g.py
```diff
@@ -80,12 +83,16 @@ def __init__(
...
+        intervals = (
+            Intervals.from_parquet(session, interval_path) if interval_path else None
+        )
         features_input_loader = L2GFeatureInputLoader(
             variant_index=variant_index,
             colocalisation=coloc,
             study_index=studies,
             study_locus=credible_set,
             target_index=target_index,
+            intervals=intervals,
```
project-defiant, line 95: You need to make sure that if the intervals are not passed (None), then:
1. If the `feature_list` contains the interval features, they should be dropped, since we can not compute them
2. The warning should be logged that the interval features can not be computed.

src/gentropy/l2g.py
```diff
@@ -172,6 +183,14 @@ def __init__(
         self.feature_matrix = L2GFeatureMatrix(
             _df=session.load_data(feature_matrix_path),
         )
+        self.target_index = (
+            TargetIndex.from_parquet(session, target_index_path)
+            if target_index_path
+            else None
+        )
+        self.intervals = (
```
project-defiant, line 191: Similarly here, if the intervals nor target index are not required, all features depending on these datasets should be filtered out before running the predictions or training.

src/gentropy/dataset/l2g_features/intervals.py
```diff
@@ -0,0 +1,351 @@
...
+    """
+    study_loci_exploded = (
+        study_loci_to_annotate.df.withColumn("variantInLocus", f.explode_outer("locus"))
+        .select(
+            "studyLocusId",
+            f.col("variantInLocus.variantId").alias("variantId"),
+            f.col("variantInLocus.posteriorProbability").alias("posteriorProbability"),
+        )
+        # Filter for PP > 0.001
+        .filter(f.col("posteriorProbability") > 0.001)
```
project-defiant, line 44: Make this value optional with default 0.001 in the function signature.

src/gentropy/config.py
```diff
@@ -216,6 +216,17 @@ class LDBasedClumpingConfig(StepConfig):
     _target_: str = "gentropy.ld_based_clumping.LDBasedClumpingStep"
 
 
+@dataclass
```
project-defiant, line 219: You have to register this step to be able to use it in the command line interface. See the gentropy.config.Config.register_config method

src/gentropy/assets/schemas/intervals.json
```diff
@@ -18,10 +18,16 @@
       "nullable": false,
       "type": "string"
     },
+    {
```
project-defiant, line 21: The variant and genes could be in the nested structure, so we preserve the storage. Similar as in the pseudo code below:
```
ArrayType(
    StructType(
        [
            StructField(StringType(), GeneId),
            StructField(
                ArrayType(StringType(), variantId)
            )
        ]
    )
)
```

project-defiant:
@xyg123 let me know if this is ready, I will look back to see if I can testrun this in the DAG.

src/gentropy/config.py
```diff
@@ -229,6 +241,7 @@ class LocusToGeneConfig(StepConfig):
     model_path: str | None = None
     gold_standard_curation_path: str | None = None
     gene_interactions_path: str | None = None
+    interval_path: str | None = None
```
project-defiant, line 244: @xyg123 the target_index_path is defined in the step, it is not in here

src/gentropy/assets/schemas/intervals.json
project-defiant: We discussed this as potential improvement in the future

project-defiant (approved):

project-defiant:
Closing, as we are reimplementing it with latest intervals

PR #821, opened 2024-10-07
feat: change LD annotation for PICS fine-mapping to use major ancestry

vivienho:
## ✨ Context

LD annotation for PICS fine-mapping previously combined multiple ancestries. In this PR this is changed so that only the major ancestry is used.

Added `_get_major_population` and `_calculate_r2_major` functions and their corresponding tests.

~~All mentions of `r2Overall` were changed to `r2Major` which is now a better name for this field.~~
`r2Overall` is kept as we may come back to it in the future.

Functions (`_add_population_size` and `_calculate_weighted_r_overall`) and corresponding tests that are no longer needed were removed.

## 🚦 Before submitting


src/gentropy/assets/schemas/study_locus.json
```diff
@@ -158,7 +158,7 @@
             },
             {
               "metadata": {},
-              "name": "r2Overall",
+              "name": "r2Major",
```
addramir, line 161: I think renaming this column is bad idea. We may return to overall in the future. Moreover changing schema now is very bad idea. I suggest use r2Overall.

vivienho: @DSuveges mentioned this as well, currently reverting those changes

DSuveges (approved):
Thank you for reverting the column names. Everything makes total sense, however I havent' had time to run the code. Good job handling cases when there are multiple major populations with the same relativeSampleSizes.

src/gentropy/method/ld.py
```diff
@@ -19,60 +20,64 @@ class LDAnnotator:
...
-                    + f.coalesce(
-                        f.pow(y["r"], 2) * y["relativeSampleSize"], f.lit(0.0)
-                    ),  # we use coalesce to avoid problems when r/relativeSampleSize is null
-                ).alias("r2Overall"),
-            ),
+        major_population_size = ordered_populations["relativeSampleSize"][0]
+        major_populations = f.filter(
+            ordered_populations,
+            lambda x: x["relativeSampleSize"] == major_population_size
+        )
```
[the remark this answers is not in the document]
vivienho: Yes, that's correct.

PR #823, opened 2024-10-09
feat: decouple feature generation from L2G training step

ireneisdoomed:
## ✨ Context

This PR introduces a new step to generate and store the feature matrix for all credible sets (unnormalised). It also corrects minor issues to successfully generate predictions after the L2G rewrite.

Results are here:
```
gsutil ls gs://ot-team/irene/l2g-0810/              
gs://ot-team/irene/l2g-0810/model.skops
gs://ot-team/irene/l2g-0810/feature_matrix/
gs://ot-team/irene/l2g-0810/predictions/
```
**Note**: the results are a proof of a successful run, I wouldn't take the scores too seriously. I had to use the June release data which is not totally compatible with the current logic (that is why colocalisation features are missing). We also have predictions for all credible sets, not only GWAS, because the current credible sets don't have the studytype metadata. _I expect these issues won't be present as soon as we use the latest data._

- Addition of `LocusToGeneFeatureMatrixStep`. All features are computed for all credible sets. The produced dataset is unnormalised. This is the longest step (~40min). Since we compute each feature separately, many tasks are repeated. We could improve this but I think the runtime is decent.
- Updates to `LocusToGeneStep` to remove the feature annotation logic. Training takes less than 5 minutes, prediction less than 10 minutes.
- Minor fixes (see commit history)

In essence, having the FM decoupled makes testing L2G training really fast.

## 🙈 Missing

To run this in the orchestration layer. I tested that the steps run with Gentropy's CLI in a dev cluster.

## 🚦 Before submitting


src/gentropy/config.py
```diff
@@ -282,10 +279,60 @@ class LocusToGeneConfig(StepConfig):
...
+    credible_set_path: str = MISSING
+    variant_index_path: str | None = None
+    colocalisation_path: str | None = None
+    study_index_path: str | None = None
+    gene_index_path: str | None = None
+    feature_matrix_path: str = MISSING
+    features_list: list[str] = field(
+        default_factory=lambda: [
+            # max CLPP for each (study, locus, gene) aggregating over a specific qtl type
+            "eQtlColocClppMaximum",
```
addramir, line 307: There is no neighbourhood features for coloc. Is it because it doesn't work with old schema for current old CSs?

ireneisdoomed: That's right. I tested in a data subset that the logic was working well.
We have only trained the model on 12 features

addramir (approved):

PR #826, opened 2024-10-09
fix: write mode added to validation steps

d0choa:
Add write mode for the 2 datasets written in each of the validation steps

### DIFF
--- a/src/gentropy/study_locus_validation.py
+++ b/src/gentropy/study_locus_validation.py
@@ -51,10 +51,10 @@ class StudyLocusValidationStep:
             .assign_confidence()
         ).persist()  # we will need this for 2 types of outputs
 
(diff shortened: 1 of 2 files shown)


project-defiant:
I am not sure if that is needed, when set up in the session

d0choa:
It's not a `SparkSesssion` feature. It's a property we added to our own `Session` class.
If you check other steps we do this everywhere. Without this change the `write.mode` is always the default `errorifexists`

project-defiant:
> It's not a `SparkSesssion` feature. It's a property we added to our own `Session` class. If you check other steps we do this everywhere. Without this change the `write.mode` is always the default `errorifexists`

Got your point.

project-defiant (approved):

PR #828, opened 2024-10-10
fix(find_overlap): missing right study type in output

DSuveges:
## ✨ Context

There was a recent change that `studyType` in identifying study locus overlaps no longer comes from study index, instead of studyLocus. It came with a side effect that upon doing outer join in the `_align_overlapping_tags` method made the `rightStudyType` nulls for tags where the left tag was not found in the right locus. This causes a range of problems downstream. 

## Scope of this PR

- Apply a fix in the `.find_overlaps()` method to carry over `rightStudyType` from the `_overlapping_peaks` method.
- Adding sample dataset of two study loci for testing.
- Adding tests to demonstrate behaviour.


tests/gentropy/dataset/test_study_locus_overlaps.py
```diff
@@ -118,10 +131,38 @@ def test_overlapping_peaks(
         [
             t.StructField("leftStudyLocusId", t.StringType()),
             t.StructField("rightStudyLocusId", t.StringType()),
+            t.StructField("rightStudyType", t.StringType()),
             t.StructField("chromosome", t.StringType()),
         ]
     )
     observed_df = spark.createDataFrame(observed, mock_schema)
     result_df = StudyLocus._overlapping_peaks(observed_df, intrastudy)
```
DSuveges, line 139: The overlapping peak method now returns the right study type as well, so in the outer join later, both the right and left pile of data will have the `rightStudyType` column. Unfortunately this also means I had to touch this test.

ireneisdoomed (approved):
Awesome, thank you for the tests!

PR #849, opened 2024-10-16
fix(eqtl): deduplicating credible set loci

DSuveges:
## ✨ Context

Because both the credible set and lbf variant data from eqtl catalogue had duplication, we had to make sure these datasets got de-duplicated before joining. Also tests were added to ensure the locus of the credible set has no duplicated elements.


tests/gentropy/data_samples/QTD000584.lbf_variable.txt
DSuveges, line 1: I was using the existing `QTD000584` credible set file, however I was changing the lbf file to get a more meaningful sample that allows other tests beside just checking the return type. The existing cs and lbf datasets were not overlapping, yielding an empty study locus at the end.

tests/gentropy/conftest.py
```diff
@@ -459,22 +459,20 @@ def sample_finngen_studies(spark: SparkSession) -> DataFrame:
...
 
 @pytest.fixture()
-def sample_eqtl_catalogue_finemapping_lbf(spark: SparkSession) -> DataFrame:
+def sample_eqtl_catalogue_finemapping_lbf(session: Session) -> DataFrame:
     """Sample raw eQTL Catalogue table with logBayesFactors outputted by SuSIE."""
-    return spark.read.option("delimiter", "\t").csv(
-        "tests/gentropy/data_samples/QTD000584.lbf_variable.txt",
-        header=True,
-        schema=EqtlCatalogueFinemapping.raw_lbf_schema,
+    return EqtlCatalogueFinemapping.read_lbf_from_source(
```
DSuveges, line 473: Because the file name needed to be captured + some simple steps were performed by the `read` functions, I had to use the dedicated read functions for the tests. Hence, I had to pass the gentropy session object instead the spark session.

ireneisdoomed (approved):
All clear, thank you for the fix!!!

PR #852, opened 2024-10-17
feat: add gene count features to l2g

xyg123:
## ✨ Context

Adds gene count feature to L2G.

Implements gene count features to L2G, which describes the gene density at a given study locus.  Currently defaults to a 500kb window.

https://docs.google.com/spreadsheets/d/1wUs1AprRCCGItZmgDhc1fF5BtwCSosdzFv4NQ8V6Dtg/edit?pli=1&gid=452826388#gid=452826388

## 🚦 Before submitting


src/gentropy/dataset/l2g_features/other.py
```diff
@@ -0,0 +1,161 @@
...
+            gene_index.df.alias("genes"),
+            on=(
+                (f.col("SL_chromosome") == f.col("genes.chromosome"))
+                & (f.col("genes.tss") >= f.col("window_start"))
+                & (f.col("genes.tss") <= f.col("window_end"))
+            ),
+            how="inner",
+        )
+        .groupBy("studyLocusId")
+        .agg(f.countDistinct("geneId").alias(feature_name))
```
ireneisdoomed, line 62: `count_distinct` triggers a shuffle of the data because the underlying algorithm requires to order the data to perform the count. On the other side, `approx_count_distinct` doesn't trigger that shuffle, and is therefore  performant.
Since we are not expecting large numbers of genes in the window, I would consider using the approximate algorithm.


```suggestion
        .agg(f.approx_count_distinct("geneId").alias(feature_name))
```

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -0,0 +1,161 @@
...
+        study_loci_window.join(
+            gene_index.df.alias("genes"),
+            on=(
+                (f.col("SL_chromosome") == f.col("genes.chromosome"))
+                & (f.col("genes.tss") >= f.col("window_start"))
+                & (f.col("genes.tss") <= f.col("window_end"))
+            ),
+            how="inner",
+        )
+        .join(distinct_gene_counts, on="studyLocusId", how="inner")
```
ireneisdoomed, line 75: Instead of doing a groupby + a join, I think you can skip the join if you count the genes applying a window on `studyLocusId`

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -0,0 +1,161 @@
...
+        .select("studyLocusId", "geneId", feature_name)
+        .distinct()
+    )
+
+
+class GeneCountFeature(L2GFeature):
+    """Counts the number of genes within a specified window size from the study locus."""
+
+    fill_na_value = 0  # Default fill value if gene count is missing
```
ireneisdoomed, line 84: Not necessary

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -0,0 +1,161 @@
...
+                value_name="featureValue",
+            ),
+            _schema=cls.get_schema(),
+        )
+
+
+class ProteinGeneCountFeature(L2GFeature):
+    """Counts the number of protein coding genes within a specified window size from the study locus."""
+
+    fill_na_value = 0  # Default fill value if gene count is missing
```
ireneisdoomed, line 125: Not necessary

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -0,0 +1,161 @@
...
+    study_loci_window = (
+        study_loci_to_annotate.df.withColumn(
+            "window_start", f.col("position") - (genomic_window / 2)
+        )
+        .withColumn("window_end", f.col("position") + (genomic_window / 2))
+        .withColumnRenamed("chromosome", "SL_chromosome")
+    )
+
+    if protein_coding_only:
+        gene_index.df = gene_index.df.filter(f.col("biotype") == "protein_coding")
```
ireneisdoomed, line 49: This changes the `df` attribute of the original gene_index.
I think there is a risk in that after you filter the df, you modify the original gene_index so that if you calculate the gene count feature the gene index won't be complete. To avoid this we usually create another object 
```suggestion
        gene_index = gene_index.filter(f.col("biotype") == "protein_coding")
```

ireneisdoomed (approved):
Thank you for the changes!
I have some comments but overall the logic looks good!
I'd like to have a semantic test that shows that the logic of `common_genecount_feature_logic` works as expected.

We usually include this test inside `tests/gentropy/dataset/test_l2g_feature.py`. @xyg123 Could you please add a test that compares an expected output with the observed output? Thanks!

ireneisdoomed (changes requested):
I approved by accident :)

src/gentropy/dataset/l2g_features/other.py
xyg123: I did try that, but when we feed the output into convert_wide_to_long it needs the gene column which gets dropped

tests/gentropy/dataset/test_l2g_feature.py
```diff
@@ -803,3 +812,106 @@ def _setup(self: TestCommonVepFeatureLogic, spark: SparkSession) -> None:
...
+                [
+                    {
+                        "studyLocusId": "1",
+                        "geneId": "gene1",
+                        "geneCountProteinCoding": 2,
+                    },
+                    {
+                        "studyLocusId": "1",
+                        "geneId": "gene2",
+                        "geneCountProteinCoding": 2,
```
ireneisdoomed, line 843: Cool, in this case after we convert the matrix to a wide format and fill the NA's with 0s, studyLocusId 1 and gene3 will have a geneCountProteinCoding count of 0.

ireneisdoomed (approved):
This is great Jack, thank you for adding the tests! Makes the logic much clearer.
Just before merging, could you please:
1. update the L2G FM and training step configuration so that geneCount500kb and proteinGeneCount500kb are added to the list of features to compute
2. Add these to the documentation? You have to add a new file: `docs/python_api/datasets/l2g_features/other.md` and refer to the classes.
Thanks!

ireneisdoomed (approved):

PR #854, opened 2024-10-17
feat: add effect size direction to coloc output

Tobi1kenobi:
## ✨ Context

Extracts the direction of effect from the study locus overlaps and appends it onto the colocalisation object.

Firstly adds a new method to StudyLocusOverlap to first calculate the ratio between each pair of left and right betas in each locus overlap. The sign of these betas is retained and then the averafe of these signs is returned as the output. See below diagram detailing process:
image

In theory this can return a value that is neither 1 or -1, however in practice this does not occur. See examples below when calculating overlaps between very large subsets of eQTL catalogue and a) Finngen
image
or b) GWAS catalog
image

Values between -1 and 1 could be captured by the front-end team if necessary and an acceptable threshold could be used e.g. beta_average_ratio > 0.8 = positive effect.

These values calculated and joined onto the colocalisation object.

## 🙈 Missing

No extensive testing of edge cases, wasn't sure how best to implement and the testing for colocs isn't particularly extensive as it is so didn't want to add a lot of extra.

## 🚦 Before submitting


d0choa:
I added @DSuveges for review as well as there have been clashing changes coming from him

addramir:
Please add fillterning by 0, otherwise is good for me.

addramir:
@DSuveges  Have you had a chance to take a look?

src/gentropy/dataset/study_locus_overlap.py
```diff
@@ -48,6 +51,36 @@ def from_associations(
...
 
+
+    def calculate_beta_ratio(self: StudyLocusOverlap) -> DataFrame:
+        """Calculate the beta ratio for the overlapping signals.
+
+        Returns:
+            DataFrame: A dataframe containing left and right loci IDs, chromosome
+            and the average sign of the beta ratio
+        """
+        expanded_overlaps = self.df.select("*", "statistics.*").drop("statistics")
```
DSuveges, line 62: I don't have objection about the logic in this function, however it is written in a quite different way than the other bits of the code. I would suggest collapsing the expressions as follows:

```python
return (
    # Unpack statistics column:
    self.df.select("*", "statistics.*")
    .drop("statistics")
    # Drop any rows where the beta is null or zero
    .filter(
        f.col("left_beta").isNotNull() & 
        f.col("right_beta").isNotNull() &
        (f.col("left_beta") != 0) &
        (f.col("right_beta") != 0) 
    )
    # Calculate the beta ratio and get the sign, then calculate the average sign across all variants in the locus
    .withColumn(
        "betaRatioSign",
        f.signum(f.col("left_beta") / f.col("right_beta"))
    )
    # Aggregate beta signs:
    .groupBy("leftStudyLocusId","rightStudyLocusId","chromosome")
    .agg(
        f.avg("betaRatioSign").alias("betaRatioSignAverage")
    )
)
```

src/gentropy/method/colocalisation.py
```diff
@@ -179,6 +179,11 @@ def colocalise(
                     f.sum(f.col("clpp")).alias("clpp"),
                 )
                 .withColumn("colocalisationMethod", f.lit(cls.METHOD_NAME))
+                .join(overlapping_signals.calculate_beta_ratio(),
+                 ["leftStudyLocusId",
+                  "rightStudyLocusId",
+                   "chromosome"],
+                "left")
```
DSuveges, line 186: I know, these are positional arguments, but I like to be more explicit in joins by providing 

```python
.join(
    overlapping_signals.calculate_beta_ratio(),
    on=["leftStudyLocusId", "rightStudyLocusId","chromosome"],
    how="left"
)
```

src/gentropy/method/colocalisation.py
```diff
@@ -379,6 +384,11 @@ def colocalise(
                     "lH4bf",
                 )
                 .withColumn("colocalisationMethod", f.lit(cls.METHOD_NAME))
+                .join(overlapping_signals.calculate_beta_ratio(),
```
DSuveges, line 387: Same here.

DSuveges (changes requested):
I don't really have objection with the logic. It looks good and makes sense. However I would recommend to apply small stylistic updates just to make the code style more aligned with the rest of the codebase.

DSuveges (approved):
Thanks. Looks good now.

PR #857, opened 2024-10-17
chore(coloc): changing the content of `numberColocalisingVariants` field

DSuveges:
## ✨ Context

Small update on how the `numberColocalisingVariants` value is computed. So far this was the number of union of tagging variants of the two colocalising credible sets. Now this number indicates the number of shared tagging variants.


src/gentropy/method/pics.py
```diff
@@ -160,6 +160,10 @@ def _finemap(
                 # If PICS cannot be calculated, we drop the variant from the credible set
                 continue
 
+            # Chaing chema:
```
d0choa, line 163: Is this accidental?

DSuveges: This PR also fixes the variantId null problem in the PICSed locus.

d0choa (approved):
Let's give it a shot.

@vivienho can you pull these changes to the branch to make sure we use the PICS fixes in the two PICS DAGs.

PR #862, opened 2024-10-18
fix: susie credible sets with unknown confidence

d0choa:
This fixes the wrong assignment of confidence to SuSiE credible sets.

At the moment we were handling SuSiE-inf but not SuSiE.

In the future we will create an Enum of finemapiing method.

### DIFF
--- a/src/gentropy/dataset/study_locus.py
+++ b/src/gentropy/dataset/study_locus.py
@@ -1175,7 +1175,7 @@ class StudyLocus(Dataset):
         df = self.df.withColumn(
             "confidence",
             f.when(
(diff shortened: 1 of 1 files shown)


addramir:
Please, check the spelling of methods. I remember, for some of the datasets we had "SuSie", not "SuSiE"

d0choa:
Thank you @addramir. You saved me another headache.

It emphasises why we need the Enum.

Do you mind reviewing it?

addramir (approved):

addramir:
But strictly speaking it should be SuSiE

d0choa:
👍

PR #867, opened 2024-10-22
feature(l2g): step to export disease/target evidence

DSuveges:
## ✨ Context

This PR implements the following:

- Evidence export method added to l2g dataset.
- Step to export evidence.
- Configuration for the step.

### DIFF
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -610,6 +610,23 @@ class StudyValidationStepConfig(StepConfig):
 
 
 @dataclass
+class LocusToGeneEvidenceStepConfig(StepConfig):
+    """Configuration of the locus to gene evidence step."""
+
+    session: Any = field(
+        default_factory=lambda: {
+            "start_hail": False,
+        }
+    )
+    locus_to_gene_predictions_path: str = MISSING
+    credible_set_path: str = MISSING
+    study_index_path: str = MISSING
+    evidence_output_path: str = MISSING
+    locus_to_gene_threshold: float = 0.05
+    _target_: str = "gentropy.l2g.LocusToGeneEvidenceStep"
+
+
+@dataclass
 class StudyLocusValidationStepConfig(StepConfig):
     """Configuration of the study index validation step.
 
@@ -703,5 +720,10 @@ def register_config() -> None:
         group="step",
         name="study_validation",
         node=StudyValidationStepConfig,
+    )
+    cs.store(
+        group="step",
+        name="l2g_evidence",
+        node=LocusToGeneEvidenceStepConfig,
     )
     cs.store(group="step", name="finngen_ukb_meta_ingestion", node=FinngenUkbMetaConfig)
--- a/src/gentropy/dataset/l2g_prediction.py
+++ b/src/gentropy/dataset/l2g_prediction.py
@@ -6,11 +6,13 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Type
 
 import pyspark.sql.functions as f
+from pyspark.sql import DataFrame
 
 from gentropy.common.schemas import parse_spark_schema
 from gentropy.common.session import Session
 from gentropy.dataset.dataset import Dataset
 from gentropy.dataset.l2g_feature_matrix import L2GFeatureMatrix
+from gentropy.dataset.study_index import StudyIndex
 from gentropy.dataset.study_locus import StudyLocus
 from gentropy.method.l2g.model import LocusToGeneModel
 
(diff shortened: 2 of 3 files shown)


src/gentropy/config.py
```diff
@@ -609,6 +609,23 @@ class StudyValidationStepConfig(StepConfig):
     _target_: str = "gentropy.study_validation.StudyValidationStep"
 
 
+@dataclass
+class LocusToGeneEvidenceStepConfig(StepConfig):
+    """Configuration of the locus to gene evidence step."""
+
+    session: Any = field(
```
ireneisdoomed, line 616: Do you need to specify the session in the config? Hail is set to false by default

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -83,3 +85,36 @@ def from_credible_set(
         )
 
         return l2g_model.predict(fm, session)
+
+    def to_disease_target_evidence(
+        self: L2GPrediction,
+        study_locus: StudyLocus,
+        study_index: StudyIndex,
+        l2g_threshold: float = 0.05,
+    ) -> DataFrame:
```
ireneisdoomed, line 94: Virtually this is another source of evidence generation, and it is not going to go through validation unlike the others.
Maybe it's annoying, but I think it'd be nice to include a new data model.

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -83,3 +85,36 @@ def from_credible_set(
...
+
+        Returns:
+            DataFrame: Disease target evidence
+        """
+        datasource_id = "gwas_credible_sets"
+        datatype_id = "genetic_association"
+
+        return (
+            self.df.filter(f.col("score") >= l2g_threshold)
+            .join(study_locus.df, on="studyLocusId", how="inner")
```
ireneisdoomed, line 110: To make the join lighter
```suggestion
            .join(study_locus.df.select("studyLocusId", "studyId"), on="studyLocusId", how="inner")
```

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -83,3 +85,36 @@ def from_credible_set(
...
+            .join(study_index.df.drop("geneId"), on="studyId", how="inner")
```
ireneisdoomed, line 111: To make the join lighter
```suggestion
            .join(study_index.df.select("studyId", "diseaseIds"), on="studyId", how="inner")
```

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -83,3 +85,36 @@ def from_credible_set(
...
+            .select(
+                f.lit(datatype_id).alias("datatypeId"),
+                f.lit(datasource_id).alias("datasourceId"),
+                f.col("geneId").alias("targetFromSourceId"),
+                f.explode(f.col("diseaseIds")).alias("diseaseFromSourceMappedId"),
+                f.col("score").alias("resourceScore"),
+                f.col("studyLocusId").alias("studyLocusId"),
```
ireneisdoomed, line 118: Redundant?
```suggestion
                f.col("studyLocusId"),
```

src/gentropy/l2g.py
```diff
@@ -278,3 +279,45 @@ def _annotate_gold_standards_w_feature_matrix(self) -> L2GFeatureMatrix:
...
+    """Locus to gene evidence step."""
+
+    def __init__(
+        self,
+        session: Session,
+        locus_to_gene_predictions_path: str,
+        credible_set_path: str,
+        study_index_path: str,
+        evidence_output_path: str,
+        locus_to_gene_threshold: float = LocusToGeneEvidenceStepConfig().locus_to_gene_threshold,
```
ireneisdoomed, line 294: I think i know now why you are specifying the session to the config.I had this scenario before. And if you do `LocusToGeneEvidenceStepConfig()`, you are going to trigger Spark to get or create (in this case get) the session, which is prob not desirable. Because of that, I suggest not to get the default from the config

src/gentropy/l2g.py
```diff
@@ -278,3 +279,45 @@ def _annotate_gold_standards_w_feature_matrix(self) -> L2GFeatureMatrix:
...
+        # Reading the study index
+        study_index = StudyIndex.from_parquet(session, study_index_path)
+
+        # Generate evidence and save file:
+        (
+            locus_to_gene_prediction.to_disease_target_evidence(
+                credible_sets, study_index, locus_to_gene_threshold
+            )
+            .write.mode(session.write_mode)
+            .json(evidence_output_path)
```
ireneisdoomed, line 322: JSON? Is the plan to add a step in the evidence parsers that pulls this dataset and validate it there?

ireneisdoomed (changes requested):
Thank you Daniel!
I have some questions, please have a look

src/gentropy/l2g.py
DSuveges: No. The evidence parsers are not expected to touch this dataset, however the platform etl, at its current form, ingests evidence as json. We can make is smarter later, but let's do one step at a time.

src/gentropy/l2g.py
DSuveges: Yes, exactly.

src/gentropy/dataset/l2g_prediction.py
DSuveges: I suspect spark optimizes this in the background, but it's better to be more explicit.

src/gentropy/dataset/l2g_prediction.py
DSuveges: same.

src/gentropy/dataset/l2g_prediction.py
DSuveges: I was thinking about this... and not convinced. Evidence doesn't have any downstream application, no methods are called on evidence dataset... Especially considering we only have five or so columns, of which two are constant. I'm quite inclined not to have an other dataset. Later maybe... eg. in case if the generation of the VEP input requires it.

src/gentropy/l2g.py
ireneisdoomed: So do you agree that the action here is not to pull the parameter from the config and remove the session reference?

ireneisdoomed (approved):
Thank you! Last request: please reference this new step in `docs/python_api/steps/l2g.md`

src/gentropy/l2g.py
project-defiant: in case of the float you could not evaluate the `LocusToGeneEvidenceStepConfig`. Primitive types are defined at compile time if I am not mistaken.

PR #874, opened 2024-10-24
feat(variant_index): hash variants at the time of instance creation

ireneisdoomed:
## ✨ Context
There was a problem in the variant index step that caused all IDs above the default threshold of 100 to be hashed, even though this threshold was set to 300 in the configuration.
```
variant_index.df.withColumn(
                "variantId",
                VariantIndex.hash_long_variant_ids(
                    col("variantId"), col("chromosome"), col("position")
                ),
            )
```

This was because this step: a) created a Variant Index from VEP, that was hashed, b) read the Gnomad variant index, not hashed, c) joined both of them and then hashed again but without passing the parameter.
There are **2 issues** in this process:
1. Gnomad Variant index wasn't hashed, so that for variants longer than 300, the join with the VEP output wasn't successful
2. We forgot to pass the threshold in step number 3, so we hashed everything longer than 100

- To avoid these problems, I have included in the Variant Index post_init a step to **hash any variant that surpasses a threshold** (default is 300), so that hashing is implicit.
- I have also modified `Dataset` to be able to distinguish between class and spark parameters in `Dataset.from_parquet`. This was because the solution worked fine when we initialised the object like `VariantIndex(_df...)`, but no params were passed to the initialisation when using `from_parquet`.
- Unit tests for the Dataset new function and for the variant hashing
- 

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/dataset.py
+++ b/src/gentropy/dataset/dataset.py
@@ -67,6 +67,30 @@ class Dataset(ABC):
         return self._schema
 
     @classmethod
+    def _process_class_params(
+        cls, params: dict[str, Any]
+    ) -> tuple[dict[str, Any], dict[str, Any]]:
+        """Separate class initialization parameters from spark session parameters.
+
+        Args:
+            params (dict[str, Any]): Combined parameters dictionary
+
+        Returns:
+            tuple[dict[str, Any], dict[str, Any]]: (class_params, spark_params)
+        """
+        # Get all field names from the class (including parent classes)
+        class_field_names = {
+            field.name
+            for cls_ in cls.__mro__
+            if hasattr(cls_, "__dataclass_fields__")
+            for field in cls_.__dataclass_fields__.values()
+        }
+        # Separate parameters
+        class_params = {k: v for k, v in params.items() if k in class_field_names}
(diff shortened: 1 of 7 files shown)


addramir (approved):

src/gentropy/dataset/dataset.py
```diff
@@ -120,10 +144,14 @@ def from_parquet(
...
-        df = session.load_data(path, format="parquet", schema=schema, **kwargs)
+
+        # Separate class params from spark params
+        class_params, spark_params = cls._process_class_params(kwargs)
+
+        df = session.load_data(path, format="parquet", schema=schema, **spark_params)
         if df.isEmpty():
             raise ValueError(f"Parquet file is empty: {path}")
-        return cls(_df=df, _schema=schema)
+        return cls(_df=df, _schema=schema, **class_params)
```
DSuveges, line 155: Am I right that this was required to pass the `id_threshold` value to the variant index initialisation?

ireneisdoomed: Exactly. I had to distinguish the passed kwargs so that we channeled the spark related ones to load_data, and the class attributes to the instance creation bit.

PR #879, opened 2024-10-25
feat: flag and filter credible sets

Tobi1kenobi:
## ✨ Context

This PR adds a study locus QC flag for when the sum of posterior inclusion probabilities are not in the expected range [0.99,1]

- Add a method to flag abnormal loci for when the sum of PIPs < 0.99 or > 1
- Adapts tests from another study locus test function to flag both a "bad" and "good" locus

## 🚦 Before submitting


src/gentropy/dataset/study_locus.py
```diff
@@ -391,6 +393,53 @@ def _qc_subsignificant_associations(
...
+            else f.lit(None).cast(ArrayType(StringType()))
+        )
+
+        flag = (self.df.withColumn(
+            "sumPosteriorProbability",
+            f.aggregate(
+                f.col("locus"),
+                f.lit(0.0),
+                lambda acc, x: acc + x["posteriorProbability"]
+            )).filter((f.col("sumPosteriorProbability") < 0.99) | (f.col("sumPosteriorProbability") > 1.0))
```
ireneisdoomed, line 423: Instead of applying a filter + a join, you can calculate the sum, and based on that create a flag `pipOutOfRange` or something like this to update the quality flag array

src/gentropy/study_locus_validation.py
```diff
@@ -45,6 +45,8 @@ def __init__(
             .annotate_study_type(study_index)  # Add study type to study locus
             .qc_redundant_top_hits_from_PICS()  # Flagging top hits from studies with PICS summary statistics
             .qc_explained_by_SuSiE()  # Flagging credible sets in regions explained by SuSiE
+            # Flagging credible sets with PIP > 1 or PIP < 0.99
+            .qc_abnormal_pips(sum_pips_lower_threshold=0.99,sum_pips_upper_threshold=1)
```
ireneisdoomed, line 49: I guess that we are not dropping credible sets due to the lower bound, only the upper?

ireneisdoomed (approved):
Looks good Tobi, thanks! Have a look at my suggestion, I think it is more performant
Thank you for the tests

ireneisdoomed:
Do you have an idea of how many CS have this issue?

src/gentropy/dataset/study_locus.py
Tobi1kenobi: Makes sense, yeah - I'll implement this a bit later today

src/gentropy/study_locus_validation.py
Tobi1kenobi: Both below lower or above the upper bound are dropped, in theory - in practice I'm not so sure

addramir:
I would be happy if 0 CSs are having this issue. If not - we have a problem.

Tobi1kenobi:
Implemented @ireneisdoomed's suggestion and after cross-checking vs gwas catalog susie I tweaked the threshold to be 1.0001 rather than 1 as there were some sums creeping through likely due to floating point error.

src/gentropy/dataset/study_locus.py
```diff
@@ -420,23 +420,25 @@ def qc_abnormal_pips(
...
-                    flag, "studyLocusId", "left"
-                )
+            _df=(flag
                 # Flagging loci with failed studies:
                 .withColumn(
                     "qualityControls",
                     self.update_quality_flag(
                         qc_select_expression,
-                        f.col("sumPosteriorProbability").isNotNull(),
+                        f.col("pipOutOfRange") == "outside",
```
ireneisdoomed, line 438: nit: this could be a boolean

ireneisdoomed (approved):
Thank you!

src/gentropy/dataset/study_locus.py
Tobi1kenobi: True! Done!

PR #886, opened 2024-10-31
feat(trainer): log model explanation with shap

ireneisdoomed:
## ✨ Context

This PR closes https://github.com/opentargets/issues/issues/3562 and adds a more comprehensive understanding of the feature contributions to the model by integrating SHAP

I have tested that it runs, I have logged metrics using the gold standard https://github.com/opentargets/issues/issues/3526#issuecomment-2438453860, and the feature matrix from the 5th data freeze. 
Panels have also been rearranged to easier to interact with https://wandb.ai/open-targets/gentropy-locus-to-gene/runs/n6wuhhe2?nw=nwuseropentargets

- Added logging for 3 types of plots from SHAP:
### 1. Feature importances
_(Plots section in the dashboard)_
image

Bar plot with the absolute contribution of each feature to the model. In the X axis we see the mean absolute SHAP value, which measures the average influence of each feature on the prediction probability. **It informs about magnitude, not directionality.**
For example, a mean SHAP value of 0.6 for `distanceFootprintMean` means that this feature, on average, impacts predictions by 0.6 units. It's also interesting to see the lower tail, with features without any effect on the prediction.
**Action: to check if there are non nulls in the gold standard for these features**

--> Also interesting to see the difference between this plot and the other one computed based on an **impurity** based approach. Shapley values are based on the magnitude of feature attributions, whereas impurity based methods "measure how effective the feature is at reducing uncertainty when creating decision trees". I don't really know the limitations/biases of each method, I found this article interesting https://explained.ai/rf-importance/

### 2. Beeswarm plot
_(Plots section in the dashboard)_
image

The beeswarm plot gives an idea of both the **magnitude and the direction** of the feature in relation with the prediction.
Here features are ordered in terms of importance (mean SHAP value). Each dot represents a single row of the training set for that feature. SHAP values are expressed in terms of log odds, meaning that that positive values push predictions towards the positive class, and negative values push predictions towards the negatives. If we see features that cluster around 0, the contribution is small.

We see that in general _our features have a positive effect on the prediction_, meaning that high values tend to increase the probability of a positive prediction. However, it is interesting to see `proteinGeneCount500kb`, that indicates that low values tend to push the prediction to the negatives, but also to the positives in some instances.

### 3. Scatter plot between the values of a feature and its contribution to the model
_(Feature/importance correlation section in the dashboard)_
image

The scatter plot shows the effect a single feature has on the predictions made by the model

Again, each dot is a single prediction from the training set. The X axis is the value of the feature, and the Y axis is the SHAP value as log odds for that feature, representing how much knowing that feature’s value changes the output of the model for that sample’s prediction. The light grey area at the bottom of the plot is a histogram showing the distribution of data values.

For our most important feature, its contribution is largely stable as the feature increases, but only when the feature is ~1 we observe large effect sizes.

- Adds `shap` and `matplotlib` to the project. Matplotlib is needed to tweak the plots outputed by shap.
    - I tried exporting the plots as HTML to make them interactive  in W&B but I didn't manage. What we do is that we save the plot in PNG and log that image to W&B
    - The script to install the dependencies on the cluster was failing because`shap` has a dependency, `llvm`, that can't be uninstalled to later be installed. In ae34fea20834f1e07c1b97f58be457cc9f6c343d I  have added a flag to overwrite the installation even when the package is already installed
- The former feature importance plot has feature labels now. The error was due to `trainer.features_list` being null

## 🚦 Before submitting


ireneisdoomed:
Btw, this is a subset of plots that I thought could bring the most value. I've tested other things, in the notebook https://gist.github.com/ireneisdoomed/d60009ee9ccda6e5d2445e0cdfd83d91

ireneisdoomed:
@addramir Yes, please see the "Local interpretation" section in the notebook.
We can explain a prediction in different ways, I particularly like the waterfall plot. 

image

> The bottom of a waterfall plot starts as the expected value of the model output, and then each row shows how the positive (red) or negative (blue) contribution of each feature moves the value from the expected model output over the background dataset to the model output for this prediction

The starting probability is ~0.0067 (log odds of -5.3), and you see how each feature contribute to the final probability of ~0.98. What i like about it is also the greyed numbers next to the features, they indicate the value of each feature for this sample.

We could build our own plot (for which we'd need to store the shapley values in the predictions dataset), or use shap directly on the fly. The model is stored in Hugging Face, as soon as it is public the explaining function would download it, create the explainer and calculate the shapley values for that particular prediction.

I haven't implemented this yet because it depends on how we want to show this: if we use it in production, I suggest option 1, but if we only want to do it within gentropy as an explorative function, option 2 is more convenient.

addramir (approved):

PR #887, opened 2024-10-31
feat: deconvolute studies upon ingestion of GWAS Catalog datasets 

DSuveges:
## ✨ Context

As the GWAS Catalog dataset is now split into three different dataset (top-hit, sumstats pics, sumstats SuSiE). Each dataset provides credible sets and study indices. Each process has its own complication and extra annotation. Which makes the ETL level aggregation very complicated.

### Consideration

**What to do when summary statistics is available**: - signalled by the `hasSumstat`  True value.

- Curated top hits needs to be dropped.
- The `hasSumstat` flag needs to be propagated.
- Any quality controls coming from curation needs to be propagated: `qualityControls`, `analysisFlags` and `sumstatQCValues` columns.
- The SuSiE finemapping path implies extra curation that might affect `studyType`. If the study type changes, the study should be invalidated and no credible set should be ingested from them.

**What to do when summary statistics is not available**: - signalled by the `hasSumstat`  False value.

- The top hit ingestion includes a study splitting step that might yield to changing the list of diseases in the `traitFromSourceMappedIds` column. The updated column needs to be propagated.
- Similarly the `traitFromSource` is also changed upon splitting, which also needs to be propagated.

## This PR includes

- StudyIndex method to de-convolute studies.
- Deconvolution is added to study validation step.
- StudyLocusQualityCheck is extended with flags: `FLAGGED_STUDY`. The reason is that we are now have a number of study flags that are tolerable and should not lead to dropping credible set.
- An other flag: `SUMMARY_STATISTICS_AVAILABLE` - top hits are flagged if summary statistics are available, which will lead to dropping the credible set.
- As some of the studies will come from one of the source, still with the top-hit flag, the deconvolution process needs to be aware that there's no summary statistics, so this flag needs to be added upon ingestion. So there's a method to do so in the top hit ingestion.
- The study locus method `filter_by_study_type` is removed as no longer used.


src/gentropy/dataset/study_index.py
```diff
@@ -590,3 +592,158 @@ def annotate_sumstats_qc(
...
+                    f.array_distinct(
+                        f.flatten(
+                            f.collect_set("qualityControls").over(full_study_id_window)
+                        )
+                    ),
+                )
+                # Propagating sumstatQCValues -> map, cannot be flatten:
+                .withColumn(
+                    "sumstatQCValues",
+                    f.first("sumstatQCValues", ignorenulls=True).over(
```
project-defiant, line 675: There always should be only one row with `sumstatQCValues` populated. otherwise the study will be flagged out as invalid right?

src/gentropy/dataset/study_index.py
```diff
@@ -590,3 +592,158 @@ def annotate_sumstats_qc(
...
+                )
+                # Remove less granular disease mappings:
+                .withColumn(
+                    "traitFromSourceMappedIds",
+                    f.when(f.col("mostGranular"), f.col("traitFromSourceMappedIds")),
+                )
+                # Propagate mapped disease:
+                .withColumn(
+                    "traitFromSourceMappedIds",
+                    f.last(f.col("traitFromSourceMappedIds"), True).over(
```
project-defiant, line 722: Why do you get the last entry from group?

src/gentropy/dataset/study_index.py
```diff
@@ -590,3 +592,158 @@ def annotate_sumstats_qc(
...
+                )
+                # Distinct study types are joined together into a string. So, if there's ambiguite, the study will be flagged when the study type is validated:
+                .withColumn(
+                    "studyType",
+                    f.concat_ws(
+                        ",", f.collect_set("studyType").over(full_study_id_window)
+                    ),
+                )
+                # At this point, all studies in one window is expected to be identical. Let's just pick one:
+                .withColumn("rank", f.row_number().over(study_id_window))
```
project-defiant, line 744: This is a quiet complex set of window operations. I wonder if it would not be easier to follow:
- anti join pics and susie and rejoin difference to susie
- anti join top_hits and sumstats and rejoin difference to sumstats

src/gentropy/dataset/study_locus.py
```diff
@@ -169,13 +178,30 @@ def validate_study(self: StudyLocus, study_index: StudyIndex) -> StudyLocus:
...
+                    StudyLocus.update_quality_flag(
+                        f.col("qualityControls"),
+                        # Condition is true, if the study has summary statistics available and the locus is a top hit:
+                        f.array_contains(
+                            f.col("qualityControls"),
+                            StudyLocusQualityCheck.TOP_HIT.value,
+                        )
+                        & f.array_contains(
+                            f.col("study_qualityControls"),
+                            StudyQualityCheck.SUMSTATS_NOT_AVAILABLE.value,
```
project-defiant, line 202: Does the logic holds here?

`TOP_HIT` & `SUMSTATS_NOT_AVAILABLE` => `SUMMARY_STATISTICS_AVAILABLE` 

MIssing `~` perhaps?

src/gentropy/datasource/gwas_catalog/study_index.py
```diff
@@ -647,6 +647,22 @@ def apply_inclusion_list(
...
+
+        Returns:
+            StudyIndexGWASCatalog: Updated study index.
+        """
+        self.df = self.df.withColumn(
+            "qualityControls",
+            StudyIndex.update_quality_flag(
+                f.col("qualityControls"),
+                ~f.col("hasSumstats"),
+                StudyQualityCheck.SUMSTATS_NOT_AVAILABLE,
```
project-defiant, line 661: This is a bit misleading I think, as this would not mean that sumstats are not available, they are, but we did not| could not harmonise them, besides that I can see clear issue with the `qualityControls` and harmonisation qc. This get's confusing to reason what we refer in the code.

src/gentropy/study_locus_validation.py
```diff
@@ -46,13 +46,13 @@ def __init__(
             .qc_redundant_top_hits_from_PICS()  # Flagging top hits from studies with PICS summary statistics
             .qc_explained_by_SuSiE()  # Flagging credible sets in regions explained by SuSiE
             # Flagging credible sets with PIP > 1 or PIP < 0.99
-            .qc_abnormal_pips(sum_pips_lower_threshold=0.99,sum_pips_upper_threshold=1.0001)
+            .qc_abnormal_pips(
```
project-defiant, line 49: Some formatting issues pop out I can see, i wonder why

src/gentropy/study_validation.py
```diff
@@ -62,8 +62,8 @@ def __init__(
 
         # Running validation:
         study_index_with_qc = (
-            study_index.validate_unique_study_id()  # Flagging duplicated study ids
-            .validate_study_type()  # Flagging non-supported study types.
+            study_index.deconvolute_studies()  # Deconvolute studies where the same study is ingested from multiple sources
```
project-defiant, line 65: I can see, why it can be challenging to merge, them as the index is already merged :/ Mayby this reasoning should be in the validation ?

tests/gentropy/dataset/test_study_locus.py
```diff
@@ -151,41 +151,6 @@ def test_find_overlaps(mock_study_locus: StudyLocus) -> None:
     assert isinstance(mock_study_locus.find_overlaps(), StudyLocusOverlap)
```
project-defiant, line 151: Could be useful to see some example test (perhaps study that holds on 3 indices?)

project-defiant (commented):
@DSuveges, it's a bit of a workaround to get this feature working. Nice work. Please see my comments.

addramir (approved):
It looks good to me.

PR #888, opened 2024-10-31
feat: add step to generate association data

vivienho:
## ✨ Context

We want to generate assocations from l2g evidence without relying on the platform etl.

This PR adds a step to generate direct and indirect associations from l2g evidence and saves them as parquet files.

## 🚦 Before submitting


src/gentropy/common/utils.py
```diff
@@ -315,3 +315,22 @@ def copy_to_gcs(source_path: str, destination_blob: str) -> None:
...
+        Column: column of harmonic sums
+    """
+    return f.aggregate(
+        f.arrays_zip(
+            f.sort_array(input_array, False).alias("score"),
+            f.sequence(f.lit(1), f.size(input_array)).alias("pos")
+        ),
+        f.lit(0.0),
+        lambda acc, x: acc
+        + x["score"]/f.pow(x["pos"], 2)/f.lit(sum(1 / ((i + 1)**2) for i in range(100)))
```
addramir, line 335: Why only first 100 are used?
And It should be a devision by 1.644 somewhere...

vivienho: That part represents the division by ~1.644
I initially used first 1000 (1.6439..) but changed it to 100 (1.6349..)
Should I just change it to f.lit(1.644) ?

addramir: Sorry, misread it. All is fine. But I would use 1000 (to be consistent with the platform if documentation is correct)

DSuveges (approved):
This PR adds a new step to gentropy to generate disease/target association (direct and indirect) based on l2g evidence. This also requires the inclusion of a new method to compute harmonic sum of values.

PR #897, opened 2024-11-04
refactor: finemapping method enum

d0choa:
This PR aims to create a `FinemappingMethod` enum that would constrain the literals used as fine-mapping methods in the credible sets.

**IMPORTANT**: This PR is a refactor, so it does not attempt to change the literals used in the current data. Once this PR is merged, we could change the literals by changing the value of the Enum in one single place. However, we must recompute the staging data (eQTL Catalog, FinnGen, PICS sumstats and PICS top-hits) and the ETL, so we should wait until the change is justified.


src/gentropy/dataset/study_locus.py
```diff
@@ -139,6 +139,20 @@ class CredibleInterval(Enum):
     IS99 = "is99CredibleSet"
 
 
+class FinemappingMethod(Enum):
```
project-defiant, line 142: This is quet good, but mayby using the `ColocalisationMethod.METHOD_NAME` would be enough, so we do not maintain two label sources?

addramir (approved):
Looks good to me.

d0choa:
@vivienho this will unblock the renaming of PICS

PR #899, opened 2024-11-05
feat: adding l2g features to prediction table

DSuveges:
A simple extension of the l2g step to add feature matrix to l2g prediction table. As such the following updates were made:

- Schema of l2g prediction is updated to have an extra optional column to contain features as a map.
- l2g prediction dataset is extended with a method to enrich dataset with the provided features.
- l2g step is extended with calling the above method.

### DIFF
--- a/src/gentropy/dataset/l2g_prediction.py
+++ b/src/gentropy/dataset/l2g_prediction.py
@@ -126,3 +126,46 @@ class L2GPrediction(Dataset):
                 "studyLocusId",
             )
         )
+
+    def add_features(
+        self: L2GPrediction, feature_matrix: L2GFeatureMatrix
+    ) -> L2GPrediction:
+        """Add features to the L2G predictions.
+
+        Args:
+            feature_matrix (L2GFeatureMatrix): Feature matrix dataset
+
+        Returns:
+            L2GPrediction: L2G predictions with additional features
+        """
+        # Columns identifying a studyLocus/gene pair
+        prediction_id_columns = ["studyLocusId", "geneId"]
+
+        # L2G matrix columns to build the map:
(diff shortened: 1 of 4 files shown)


src/gentropy/dataset/l2g_prediction.py
```diff
@@ -126,3 +126,54 @@ def to_disease_target_evidence(
...
+
+        Returns:
+            L2GPrediction: L2G predictions with additional features
+        """
+        # Testing if `locusToGeneFeatures` column already exists:
+        if "locusToGeneFeatures" in self.df.columns:
+            self.df = self.df.drop("locusToGeneFeatures")
+
+        # Columns identifying a studyLocus/gene pair
+        prediction_id_columns = ["studyLocusId", "geneId"]
```
project-defiant, line 146: skip the `goldStandardSet` column?

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -126,3 +126,54 @@ def to_disease_target_evidence(
...
+            for column in feature_matrix._df.columns
+            if column not in prediction_id_columns
+        ]
+
+        # Aggregating all features into a single map column:
+        aggregated_features = (
+            feature_matrix._df.withColumn(
+                "locusToGeneFeatures",
+                f.create_map(
+                    *sum(
```
project-defiant, line 160: 😨  I would not think of the expression like `sum(List[Tuple[Col, Col]], Tuple[]). Nice, although i it seems to be the way to flatten. https://stackoverflow.com/questions/10632839/transform-list-of-tuples-into-a-flat-list-or-a-matrix

project-defiant (approved):

PR #912, opened 2024-11-12
feat(gold_standard): arbitrary gold standards

project-defiant:
## ✨ Context

This PR closes the [#3625](https://github.com/opentargets/issues/issues/3625)

Read the gold standard from a valid `L2GGoldStandard` dataset when provided in the `gold_standard_path` to the LocusToGeneStep.

- [x] Extraction of training set preparation depending on the kind of gold standard passed in
- [x] Read the dataset in json format 

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/dataset.py
+++ b/src/gentropy/dataset/dataset.py
@@ -153,33 +153,6 @@ class Dataset(ABC):
             raise ValueError(f"Parquet file is empty: {path}")
         return cls(_df=df, _schema=schema, **class_params)
 
-    @classmethod
-    def from_json(
-        cls: type[Self],
-        session: Session,
-        path: str | list[str],
-        **kwargs: bool | float | int | str | None,
-    ) -> Self:
-        """Read json into a Dataset with a given schema.
-
-        Args:
-            session (Session): Spark session
-            path (str | list[str]): Path to the json dataset
-            **kwargs (bool | float | int | str | None): Additional arguments to pass to spark.read.json
-
-        Returns:
-            Self: Dataset with the parquet file contents
-
-        Raises:
-            ValueError: Parquet file is empty
-        """
-        schema = cls.get_schema()
-        class_params, spark_params = cls._process_class_params(kwargs)
-        df = session.load_data(path, format="json", schema=schema, **spark_params)
-        if df.isEmpty():
-            raise ValueError(f"JSON file is empty: {path}")
-        return cls(_df=df, _schema=schema, **class_params)
-
     def filter(self: Self, condition: Column) -> Self:
         """Creates a new instance of a Dataset with the DataFrame filtered by the condition.
 
--- a/src/gentropy/dataset/dataset.py
+++ b/src/gentropy/dataset/dataset.py
@@ -153,6 +153,33 @@ class Dataset(ABC):
             raise ValueError(f"Parquet file is empty: {path}")
         return cls(_df=df, _schema=schema, **class_params)
 
+    @classmethod
+    def from_json(
+        cls: type[Self],
+        session: Session,
+        path: str | list[str],
+        **kwargs: bool | float | int | str | None,
+    ) -> Self:
+        """Read json into a Dataset with a given schema.
+
+        Args:
+            session (Session): Spark session
+            path (str | list[str]): Path to the json dataset
+            **kwargs (bool | float | int | str | None): Additional arguments to pass to spark.read.json
+
+        Returns:
+            Self: Dataset with the parquet file contents
+
+        Raises:
+            ValueError: Parquet file is empty
+        """
+        schema = cls.get_schema()
+        class_params, spark_params = cls._process_class_params(kwargs)
+        df = session.load_data(path, format="json", schema=schema, **spark_params)
+        if df.isEmpty():
+            raise ValueError(f"JSON file is empty: {path}")
+        return cls(_df=df, _schema=schema, **class_params)
+
     def filter(self: Self, condition: Column) -> Self:
         """Creates a new instance of a Dataset with the DataFrame filtered by the condition.
 
--- a/src/gentropy/l2g.py
+++ b/src/gentropy/l2g.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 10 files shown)


src/gentropy/dataset/dataset.py
```diff
@@ -153,6 +153,33 @@ def from_parquet(
             raise ValueError(f"Parquet file is empty: {path}")
         return cls(_df=df, _schema=schema, **class_params)
 
+    @classmethod
+    def from_json(
```
ireneisdoomed, line 157: I'd make `from_parquet` capable of taking diverse formats as input, but maybe for another PR

src/gentropy/l2g.py
```diff
@@ -160,26 +165,92 @@ def __init__(
...
+                "unexpected_columns": [
+                    "geneId",
+                    "goldStandardSet",
+                    "studyId",
+                    "studyLocusId",
+                    "variantId",
+                ]
+            }:
+                # There are schema mismatches, this would mean that we have
+                logging.warning("Detected OTG Gold Standard. Attempting to parse it.")
```
ireneisdoomed, line 221:
```suggestion
                logging.info("Detected OTG Gold Standard. Attempting to parse it.")
```

src/gentropy/l2g.py
```diff
@@ -160,26 +165,92 @@ def __init__(
...
+            ValueError: When gold standard path, is not provided, or when
+                parsing OTG gold standard but missing interactions and variant index paths.
+            TypeError: When gold standard is not OTG gold standard nor L2GGoldStandard.
+
+        """
+        if self.gold_standard_curation_path is None:
+            raise ValueError("Gold Standard is required for model training.")
+        # Read the gold standard.
+        spark = self.session.spark
+        gold_standard = spark.read.json(self.gold_standard_curation_path)
```
ireneisdoomed, line 191:
```suggestion
        gold_standard = session.load_data(self.gold_standard_curation_path, format="json")
```

src/gentropy/l2g.py
```diff
@@ -160,26 +165,92 @@ def __init__(
...
-            self.interactions = (
-                self.session.spark.read.parquet(gene_interactions_path)
-                if gene_interactions_path
-                else None
-            )
-            self.run_train()
+            gold_standard = self.prepare_gold_standard()
+            self.run_train(gold_standard)
+
+    def prepare_gold_standard(self) -> L2GGoldStandard:
```
ireneisdoomed, line 175: I like the logic in general. We could also have 2 different params: one for the raw GS, and another one for the complete GS, but it could be confusing.

The only problem I have is expecting that the complete GS comes in JSON format to match the old curation format. We internally work with Parquets everywhere. If anything, I'll convert the old file to Parquet

src/gentropy/l2g.py
```diff
@@ -205,89 +276,64 @@ def run_predict(self) -> None:
...
-                        hf_hub_token,
-                        data=trained_model.training_data._df.drop(
-                            "goldStandardSet", "geneId"
-                        ).toPandas(),
-                        repo_id=self.hf_hub_repo_id,
-                        commit_message=self.hf_model_commit_message,
-                    )
-
-    def _annotate_gold_standards_w_feature_matrix(self) -> L2GFeatureMatrix:
+    def run_train(self, gold_standard: L2GGoldStandard) -> None:
```
ireneisdoomed, line 286: I'd set `self.gold_standard` as an instance attribute.
```suggestion
    def run_train(self) -> None:
```

src/gentropy/l2g.py
```diff
@@ -205,89 +276,64 @@ def run_predict(self) -> None:
...
+                    self.model_path.split("/")[-1],
+                    hf_hub_token,
+                    data=trained_model.training_data._df.drop(
+                        "goldStandardSet", "geneId"
+                    ).toPandas(),
+                    repo_id=self.hf_hub_repo_id,
+                    commit_message=self.hf_model_commit_message,
+                )
+
+    def _annotate_gold_standards_w_feature_matrix(
```
ireneisdoomed, line 328: Same as above

src/gentropy/method/l2g/model.py
```diff
@@ -42,21 +42,32 @@ def __post_init__(self: LocusToGeneModel) -> None:
...
+            path (str): Path to the model
 
         Returns:
             LocusToGeneModel: L2G model loaded from disk
 
         Raises:
             ValueError: If the model has not been fitted yet
         """
-        loaded_model = sio.load(path, trusted=sio.get_untrusted_types(file=path))
+        if path.startswith("gs://"):
```
ireneisdoomed, line 57: Thank you for this

src/gentropy/method/l2g/trainer.py
```diff
@@ -105,7 +105,7 @@ def _get_shap_explanation(
                 data=training_data,
                 feature_perturbation="interventional",
             )
-            return explainer(training_data)
+            return explainer(training_data, check_additivity=False)
```
ireneisdoomed, line 108: This should be optional. The check should be true by default

ireneisdoomed (approved):
See comments

src/gentropy/l2g.py
project-defiant: I wanted to avoid setting `self.gold_standard = None` in the first place, because then we have to consecutively check if it is not None each time we want to refer to it (arbitrary code nesting that is not needed).

Given that the LocusToGeneStep `__init__` is actually performing the calculations, rather then just initialising, I could extract the logic to create a gold standard into the `__init__` rather then `run_train`. Then we can avoid the None checks downstream.

src/gentropy/dataset/dataset.py
project-defiant: `from_parquet("gs:/some_file.json")` is semantically wrong. We could make the method more dynamic though, as the implementation is the same between `from_parquet` and `from_json`. Either way it is no longer needed for the gold standard, I will remove it.

src/gentropy/method/l2g/trainer.py
project-defiant: I will check the error and produce the result without the additivity check if it is raised.

ireneisdoomed (approved):
Love the changes, thanks!

PR #913, opened 2024-11-12
feat: redefine neighbourhood features to represent similarity with best metric + other fixes

ireneisdoomed:
## ✨ Context
Changes here represent the result of iterating over the feature matrix definitions. More context here https://github.com/opentargets/issues/issues/3610

This has been coauthored with @addramir 

- L2G only prioritises protein coding genes: **non protein coding genes are filtered out from the training and the prediction set**.
    - Subsequently, `isProteinCoding` is no longer used as a training feature
- Redefinition of features in the neighbourhood. Same principle applies for distance, VEP, and Colocalisation:
    - We no longer use the average in the region to represent the feature value, we use the **max** instead. Eventually the neighbourhood feature represents the correspondence between the value for a specific credible set/gene with the max
    - We have also changed the way we compare: instead of doing `gene/credible set value - regional max`, we are now dividing `gene/credible set value / regional max`. According to @addramir, this is more powerful in terms of catching non-linearity, but unstable if the average or min/max is 0.

## 🙈 Missing

- [ ] Now that we are taking the max value, we can remove extra logic in the colocalisation features to have a better representation of the average.

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/l2g_features/vep.py
+++ b/src/gentropy/dataset/l2g_features/vep.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any
 
 import pyspark.sql.functions as f
+from pyspark.sql import Window
 
 from gentropy.common.spark_helpers import convert_from_wide_to_long
 from gentropy.dataset.gene_index import GeneIndex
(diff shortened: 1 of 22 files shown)


addramir (approved):

src/gentropy/dataset/l2g_features/vep.py
```diff
@@ -110,9 +110,11 @@ def common_neighbourhood_vep_feature_logic(
         local_metric.join(regional_max_per_study_locus, "studyLocusId", "left")
```
addramir, line 110: Shouldn't we add here: 
gene_index.df.filter(f.col("biotype") == "protein_coding").select("geneId"),
too?

PR #914, opened 2024-11-13
feat(variant index): variant description to summarise variant consequences in transcripts

DSuveges:
## ✨ Context

For an easier interpretation of variants, certain key fields from the variant annotation table are combined together to create a `variantDescription`. This is stored and will be shown on the UI.

**Fields to combine:**

- `biotype` - Shown if not protein coding, description always shows the variant impact on the closest protein coding transcript
- `most severe consequence` - always shows the label.
- `approvedSymbol` or `geneId` - always shown for the listed listed transcripts.
- `distanceFromFootprint` - shown if transcripts are not overlapping with the variant
- `aminoAcidChange` - shown if variant causes amino acid change.
- `lofteePrediction` - shown if the variant is high-confidence loss of function predicted by loftee

### DIFF
--- a/src/gentropy/datasource/ensembl/vep_parser.py
+++ b/src/gentropy/datasource/ensembl/vep_parser.py
@@ -95,7 +95,6 @@ class VariantEffectPredictorParser:
             _schema=VariantIndex.get_schema(),
             id_threshold=hash_threshold,
         )
-        # return VariantEffectPredictorParser.process_vep_output(vep_data, hash_threshold)
 
     @staticmethod
     def _extract_ensembl_xrefs(colocated_variants: Column) -> Column:
--- a/src/gentropy/datasource/ensembl/vep_parser.py
+++ b/src/gentropy/datasource/ensembl/vep_parser.py
@@ -95,6 +95,7 @@ class VariantEffectPredictorParser:
             _schema=VariantIndex.get_schema(),
             id_threshold=hash_threshold,
         )
+        # return VariantEffectPredictorParser.process_vep_output(vep_data, hash_threshold)
 
     @staticmethod
     def _extract_ensembl_xrefs(colocated_variants: Column) -> Column:
@@ -660,6 +661,8 @@ class VariantEffectPredictorParser:
                 map_column_by_dictionary(
                     f.col("most_severe_consequence"), sequence_ontology_map
                 ).alias("mostSevereConsequenceId"),
+                # Propagate most severe consequence:
+                "most_severe_consequence",
                 # Extract HGVS identifier:
                 f.when(
                     f.size("transcript_consequences") > 0,
@@ -732,6 +735,8 @@ class VariantEffectPredictorParser:
                                 "polyphenPrediction"
                             ),
                             transcript.transcript_id.alias("transcriptId"),
+                            transcript.biotype.alias("biotype"),
+                            transcript.gene_symbol.alias("approvedSymbol"),
                         ),
                     ),
                 ).alias("transcriptConsequences"),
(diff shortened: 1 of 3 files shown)


src/gentropy/datasource/ensembl/vep_parser.py
```diff
@@ -806,8 +845,278 @@ def process_vep_output(
...
+        Args:
+            transcript (Column): Transcript.
+
+        Returns:
+            Column: Loss of function annotation.
+        """
+        return f.when(
+            transcript.getField("lofteePrediction").isNotNull()
+            & (transcript.getField("lofteePrediction") == "HC"),
+            f.lit(" A high-confidence loss-of-function variant by loftee."),
```
d0choa, line 1011:
```suggestion
            f.lit(" A high-confidence loss-of-function variant by LOFTEE."),
```

d0choa (approved):

d0choa:
Looks amazing. It should be really simple to understand what needs to be changed in the time comes.

An opportunity for improvement could be to use the variant effect to improve the description. For example if the max normalised score is > .5 to report the method that predicts the variant to be likely deleterious. An idea for the future. This looks great for now

src/gentropy/dataset/variant_index.py
```diff
@@ -298,3 +298,292 @@ def get_loftee(self: VariantIndex) -> DataFrame:
...
+            score (Column): Pangolin score.
+
+        Returns:
+            Column: Normalised Pangolin score.
+        """
+        return f.when(
+            f.abs(score) > 0.14, cls._rescaleColumnValue(f.abs(score), 0.14, 1, 0.0, 1)
+        ).when(
+            f.abs(score) <= 0.14,
+            cls._rescaleColumnValue(f.abs(score), 0, 0.14, -1, 0.0),
```
d0choa, line 588: The description and the implementation don't agree. I think the description is better. Either that or: 0-0.5 and 0.5-1. But I wouldn't make it negative as we are not really sure we are talking about benign variation

src/gentropy/dataset/variant_index.py
```diff
@@ -298,3 +298,292 @@ def get_loftee(self: VariantIndex) -> DataFrame:
...
+
+        Returns:
+            Column: Normalised score for the in silico predictor.
+        """
+        return (
+            f.when(method == "LOFTEE", cls._normalise_loftee(assessment))
+            .when(method == "SIFT", cls._normalise_sift(score, assessment))
+            .when(method == "PolyPhen", cls._normalise_polyphen(assessment, score))
+            .when(method == "AlphaMissense", cls._normalise_alpha_missense(score))
+            .when(method == "phred scaled CADD", cls._normalise_cadd(score))
```
d0choa, line 366: can we make it `CADD`?

src/gentropy/datasource/ensembl/vep_parser.py
```diff
@@ -653,13 +686,17 @@ def process_vep_output(
                             method_name="phred scaled CADD",
```
d0choa, line 686: can we make it CADD?

d0choa (approved):

d0choa:
After reviewing the data and looking at the github repo we should probably leave spliceAI as is. From 0-1. It only assesses splice afecting variants using a recommended threshold of 0.5. But it has predictive value in the whole continuous range. I would not have negatives here

DSuveges:
> An opportunity for improvement could be to use the variant effect to improve the description. For example if the max normalised score is > .5 to report the method that predicts the variant to be likely deleterious. An idea for the future. This looks great for now

I was considering this option, but given the normalised assessments are still method specific, it is not clear what to focus, or how to summarise the range of predictors. However, when things are more solidified, we can take a closer look, see how often the methods are consistent or contradicting, and we can sort out something.

DSuveges:
Updates following @d0choa's comments:

- Removed spliceAI normalisation. Score left as is.
- Pangolin normalisation updated.
- `CADD` used instead of `Phred scaled CADD`.
- LOFTEE method name capitalised.

PR #919, opened 2024-11-19
fix: r2 for lead variant is always 1

addramir:
## ✨ Context

Related to bug when credible set size is zero for 10445 CSs.
The problem is that r2 for the lead variant is 0 for 100% cases.
There are two possible fixes:
1. Ensure that r2 for lead is always 1 in the ld annotation step. 
2. Add the flag "[LD block contains no variants at the required R^2 threshold]" as an exclusion criterion.

This PR fixes 1). Good way to go, but requires recalculation of PICS credible sets.  

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/study_locus.py
+++ b/src/gentropy/dataset/study_locus.py
@@ -74,7 +74,6 @@ class StudyLocusQualityCheck(Enum):
         LD_CLUMPED (str): Explained by a more significant variant in high LD
         WINDOW_CLUMPED (str): Explained by a more significant variant in the same window
         NO_POPULATION (str): Study does not have population annotation to resolve LD
-        NOT_QUALIFYING_LD_BLOCK (str): LD block does not contain variants at the required R^2 threshold
         FLAGGED_STUDY (str): Study has quality control flag(s)
         MISSING_STUDY (str): Flagging study loci if the study is not found in the study index as a reference
         DUPLICATED_STUDYLOCUS_ID (str): Study-locus identifier is not unique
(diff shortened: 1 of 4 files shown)


DSuveges (commented):
The newly introduced transformation of the ldSet makes sure that the lead variant always has r2 = 1.0. This will prevent nullifying the locus object. 

However, this solution raises some question:
- It is still possible that the lead variant for some reason not in the LD set, so the locus object remains empty. Downstream there's no validation for mandating non-empty locus object. 
- This action changes the data and will prevent these credible sets being flagged with the `NOT_QUALIFYING_LD_BLOCK` flag. It might be desired though...

addramir:
@DSuveges 

> It is still possible that the lead variant for some reason not in the LD set, so the locus object remains empty. Downstream there's no validation for mandating non-empty locus object.
The method for annotation has method "_rescue_lead_variant" that is run before it, so it is there in 100% cases.
However it adds it only when ldSet size is 0. In the freeze6 it is always there. 

> This action changes the data and will prevent these credible sets being flagged with the NOT_QUALIFYING_LD_BLOCK flag. It might be desired though...
It will be better. The flag clearly indicates that something is wrong here. Alternatively, we can just add the flag to exclusion and filter these CSs out. No needs to recalculate everything. I would actually say it is a better strategy now.

d0choa (approved):
@DSuveges and I had a chat about this PR and we think it's all good.

Have a look at the PR opened against this branch for some code cleanup.

Otherwise good to go. The moment this is merged, we can re-run PICS

PR #921, opened 2024-11-20
feat: changing studylocus validation to 95 percent credible sets

Daniel-Considine:
## ✨ Context

This changes the study_locus_validation step to filter down to 95% credible sets instead of 99%. 

Additionally: 
- it rearranges the order of functions so that the abnormal PIPs qc flag is added after credible set filtering. 
- it changes the number of partitions for the valid credible set dataframe to 3200. This function was defaulting to 100 partitions regardless of orchestration shuffle partition parameters, resulting in huge chunks of data crashing the coloc dataproc step in the ETL. 

## 🚦 Before submitting


d0choa:
I don't think we should hardcode the 3200 in the business logic. This same behavior can be achieved using the pyspark.sql.partitions in configuration.

Parameters that depend on the workload should not be part of the business logic

Daniel-Considine:
> I don't think we should hardcode the 3200 in the business logic. This same behavior can be achieved using the pyspark.sql.partitions in configuration and another one I don't remember about minimum partition size
> 
> Parameters that depend on the workload should not be part of the business logic

I agree, I just put this as a temporary measure to run the ETL. Ideally should be parameterised and part of the orchestration layer for credible set validation.

addramir (approved):

PR #924, opened 2024-11-22
feat(gold_standard): add traitFromSourceMappedId to schema

ireneisdoomed:
## ✨ Context
This PR adds the EFO ID to the gold standard so that we can split the training set by EFO/Gene during cross validation.
The change in the schema doesnt have implication in FE/BE as discussed.
Bare in mind this will create some explosion in the gold standard.

## 🙈 Missing

A following PR will make more thorough changes in the schema of the gold standard, and in the code to accommodate this dataset to the definition of a effector gene list.
The new field is set to nullable = True for the time being, so that tests don't complain. This field should be mandatory in the future.

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/l2g_feature_matrix.py
+++ b/src/gentropy/dataset/l2g_feature_matrix.py
@@ -36,9 +36,11 @@ class L2GFeatureMatrix:
             with_gold_standard (bool): Whether to include the gold standard set in the feature matrix.
         """
         self.with_gold_standard = with_gold_standard
-        self.fixed_cols = ["studyLocusId", "geneId", "traitFromSourceMappedId"]
+        self.fixed_cols = ["studyLocusId", "geneId"]
         if self.with_gold_standard:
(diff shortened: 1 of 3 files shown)


src/gentropy/dataset/l2g_feature_matrix.py
```diff
@@ -36,7 +36,7 @@ def __init__(
             with_gold_standard (bool): Whether to include the gold standard set in the feature matrix.
         """
         self.with_gold_standard = with_gold_standard
-        self.fixed_cols = ["studyLocusId", "geneId"]
+        self.fixed_cols = ["studyLocusId", "geneId", "traitFromSourceMappedId"]
```
ireneisdoomed, line 39: Feature matrix doesn't have a schema.
But I need to define that this column is expected as one of the fixed columns to pivot on.

addramir (approved):

PR #929, opened 2024-11-26
chore(gnomad): updating GnomAD version to 4.1 from 4.0 + using joint frequencies

DSuveges:
## ✨ Context

- [x] Default configuration for GnomAD variant annotation is updated to 4.1 from 4.0.
- [x] The configuration + parsing logic is updated to point to the joint variant data instead of the full genome data.
- [x] Removing in-silico prediction extraction from GnomAD data for various reason: questionable licecing, predictors are not available in the joint dataset.

The new dataset is saved here: `gs://gnomad_data_2/v4.1/variant_index`, confirmed that the allele frequencies are concordant with the GnomAD UI joint frequencies.

### DIFF
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -387,7 +387,7 @@ class GnomadVariantConfig(StepConfig):
         }
     )
     variant_annotation_path: str = MISSING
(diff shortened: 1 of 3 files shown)


d0choa (approved):
We should require all fixes to result in less lines of code 😅

PR #941, opened 2024-12-03
refactor: derive the _schema automatically

project-defiant:
## ✨ Context

The aim of this PR is to allow for skipping the `_schema` parameter when creating a new Dataset derived class instance.

Make the `_schema` field optional in the `Dataset` abstract class.

The new behavior is following:
1. When no `_schema` is passed to the `Dataset` derived class, the schema is derived from the static assets for this particular dataset.
2. When `_schema` is passed, it is used instead as a dataframe schema.

## 🚦 Before submitting


d0choa:
You might want look at this for inspiration. I remember I discussed with @DSuveges and we thought it was a good idea, but we didn't want to refactor too much at that point:
https://github.com/opentargets/gentropy/compare/dev...do_experiment_dataset

src/gentropy/dataset/dataset.py
```diff
@@ -8,7 +8,9 @@
 from functools import reduce
 from typing import TYPE_CHECKING, Any
 
-import pyspark.sql.functions as f
+from pyspark.sql import DataFrame
+from pyspark.sql import functions as f
+from pyspark.sql import types as t
```
DSuveges, line 13: Extremely marginal, but you are importing all types as t, while in line 14 there's an explicit import of DoubleType. It can be made consistent.

src/gentropy/dataset/dataset.py
```diff
@@ -172,7 +191,7 @@ def validate_schema(self: Dataset) -> None:
         Raises:
             SchemaValidationError: If the DataFrame schema does not match the expected schema
         """
-        expected_schema = self._schema
+        expected_schema = self.schema
```
DSuveges, line 194: Isn't `self.schema` inconsistent with line 40? Where the schema is specified as `_schema: StructType | None = None`.

tests/gentropy/dataset/test_dataset.py
```diff
@@ -68,6 +80,8 @@ def test_dataset_drop_infinity_values() -> None:
     rows = [(v,) for v in data]
     schema = StructType([StructField("field", DoubleType())])
     input_df = spark.createDataFrame(rows, schema=schema)
+    input_df.printSchema()
```
DSuveges, line 83: Do you need the print for debugging?

DSuveges (approved):
Very good! I like the solution that remains compatible with the existing solution.

PR #945, opened 2024-12-12
build: define dependency version constraints + other fixes

ireneisdoomed:
## ✨ Context
A recent update to the `scikit-learn` package has caused the L2G step in Gentropy to fail due to an incompatibility between `scikit-learn` and another dependency, `skops`.  

Specifically:  
- `scikit-learn` has been updated to version `1.6.0`, which is incompatible with `skops==0.10.0`.  
- Our current TOML specifies the following version constraints:  
  ```toml
  skops = ">=0.9,<0.11"
  scikit-learn = "^1.3.2"

because skops does not pin the version of scikit-learn, dependency resolution results in incompatible package versions when a user installs the Gentropy wheel now:
- the latest version of scikit-learn (e.g., 1.6.0) will be installed.
- skops==0.10.x will be installed, which is not compatible with scikit-learn==1.6.0.

In this PR, we define upper limits to the version constraints of all dependencies. This way that we avoid resolving to versions including minor updates that have not been tested.

- Definition of an upper bound version for all dependencies
- Bump scikit-learn to 1.6 and skops to 0.11. I have tested that with these updates the step runs.
- Addition of a test to make sure that all gentropy modules are importable
- Moving all dev dependencies to the same group

## 🙈 Missing

<!--- If there are things that are requested in the task and were not implemented, list them here -->

## 🚦 Before submitting


project-defiant (commented):
To sum this PR:
* fix on the upper bound of the package versions (this makes the `caret` operator redundant in all cases (https://python-poetry.org/docs/dependency-specification/#caret-requirements)
* Test to check if all modules are importable.

Note: We could fix the versions in more famiialar way `(>=X, <Y)` (a bit simpler to understand then the caret operator.

ireneisdoomed:
@SzymonSzyszkowski I changed the caret so that it is more explicit.Thank you ^^

project-defiant (approved):

project-defiant:
@ireneisdoomed well done!

PR #946, opened 2024-12-12
refactor: remove gene_index step

vivienho:
## ✨ Context

Gentropy uses a `gene_index` that is generated from the target index in the platform etl via the `gene_index` step. The `gene_index` step is redundant, and we now want to use the target index from the platform etl directly.

This PR is also connected to these PRs in the [orchestration](https://github.com/opentargets/orchestration/pull/89) and [platform-etl-backend](https://github.com/opentargets/platform-etl-backend/pull/369) repos.

Note: After this PR is merged, gentropy pipelines will depend on a `target_index` dataset generated by the platform etl that has a `tss` column and thus the gentropy pipeline cannot be run on its own. I have generated a patched `target_index` dataset using `gs://open-targets-pre-data-releases/24.12-uo_test-3/output/etl/parquet/targets` patched with the `tss` column, just in case the gentropy pipelines need to be run on their own. The patched `target_index` can be found here: `gs://ot-team/vivien/gentropy_patched_datasets/target_index_with_tss_column`

- The PR removes all files/code related to the `gene_index` step.
- All files/code with `gene_index` are renamed to `target_index`.
- Although the `gene_index` essentially comprises of a subset of columns from the `target_index`, some field names differ, so the PR renames field names to be compatible with the `target_index` where necessary.
- The `gene_index` json schema has been replaced with the `target_index` schema derived from a `target_index` dataset generated by the platform etl.
- The `gene_index` step was used an example in the docs [here](https://opentargets.github.io/gentropy/howto/command_line/run_step_in_cli/) and [here](https://opentargets.github.io/gentropy/howto/command_line/run_step_using_config/). The `gwas_catalog_sumstat_preprocess` step is now used as the example as it has inputs of similar complexity.

## 🚦 Before submitting


docs/howto/command_line/run_step_in_cli.md
```diff
@@ -28,16 +27,16 @@ Available options:
 Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
 ```
 
-As indicated, you can run a step by specifying the step's name with the `step` argument. For example, to run the `gene_index` step, you can run:
+As indicated, you can run a step by specifying the step's name with the `step` argument. For example, to run the `gwas_catalog_sumstat_preprocess` step, you can run:
```
DSuveges, line 30: Thanks for updating the documentation! I keep forgetting about it.

tests/gentropy/conftest.py
```diff
@@ -569,34 +569,33 @@ def sample_target_index(spark: SparkSession) -> DataFrame:
 
 
 @pytest.fixture()
-def mock_gene_index(spark: SparkSession) -> GeneIndex:
-    """Mock gene index dataset."""
-    gi_schema = GeneIndex.get_schema()
+def mock_target_index(spark: SparkSession) -> TargetIndex:
+    """Mock target index dataset."""
+    ti_schema = TargetIndex.get_schema()
```
DSuveges, line 574: Theoretical question: wouldn't make more sense to use an actual data sample from the target index?

DSuveges (approved):
Very thorough refactoring of this huge codebase! The target index is phased out in favour of the target index, which is now shared between the platform and gentropy. Luckily there was not much logic refactoring.

tests/gentropy/conftest.py
vivienho: Probably? There is a `sample_target_index` [right above it in the code](https://github.com/opentargets/gentropy/blob/dev/tests/gentropy/conftest.py#L564), but it doesn't seem to be used.

PR #947, opened 2024-12-13
feat: FoldX predicted energies in variant annotation

DSuveges:
## Context

OTAR2081 mapped the mutational landscape of the full human proteome by systematically changing every amio-acid to all other amino acid in the AlphaFold structures of all human protein. The mutation effect is assessed by the comparing energies of the structures via FoldX. The raw and scaled ddG values are used as `inSilicoPredictors` for all variants in the variant index that causes the same amio-acid change in the given protein.

## The PR contains

- Logic and step to process the FoldX dataset generated by OTAR2081.
- Logic to integrate amino-acid base in-silico predictors into the the variant index.


src/gentropy/dataset/variant_index.py
```diff
@@ -395,6 +452,29 @@ def _rescaleColumnValue(
...
+
+        Args:
+            score (Column): column with ddG values
+
+        Returns:
+            Column: Normalised energies
+        """
+        return f.when(f.abs(score) >= 5, f.lit(1.0)).otherwise(
+            cls._rescaleColumnValue(f.abs(score), 0.0, 5.0, 0.0, 1.00)
+        )
```
vivienho, line 476: From the function documentation I expected that scores above 2 should be normalised to 1, but here it is 5?

vivienho (approved):
Just one minor question, otherwise everything looks good!

PR #948, opened 2024-12-13
fix(`isProteinCoding`): get all overlapping genes from the variant index 

ireneisdoomed:
## ✨ Context
This PR is related to the `isProteinCoding` L2G feature and these 2 issues:
- https://github.com/opentargets/issues/issues/3674#issuecomment-2541104753 - in that SOX5 is no longer considered a non protein coding gene;
- https://github.com/opentargets/issues/issues/3677 - in that the feature computation is no longer based on the gene's TSS

This PR largely solves an issue in the feature matrix where the protein coding annotation of the gene was wrongly attributed.

Reimplementation of `is_protein_coding_feature_logic`: instead of using the TSS location in the gene index to extract which gene the sentinel variant overlaps with, we now:
- Use intersecting genes extracted by VEP that are present in the variant index. This covers all genes 500kb up and dowstream the variant
- We use all variants in the locus instead of only the lead

Tests have been adjusted accordingly.

## 🙈 Missing
This solution is still not ideal as we still might be missing credible set/gene annotation for genes that are outside of the defined window. In that case we could still be wrongly annotating genes as non protein coding.
- [ ] To run the feature matrix step and evaluate if this is still an actual problem

## 🚦 Before submitting


xyg123 (approved):
Looks good!

PR #954, opened 2024-12-16
test: improve fixture definition by constraining dataset IDs

ireneisdoomed:
## ✨ Context
We use `dbldatagen` to build mock dataframes given a specific schema. We use this mock dataframes to test that the logic returns data in the correct **format**. For **content** testing, we have to manually define how we want the data to look like.

In a recent [PR](https://github.com/opentargets/gentropy/pull/948), there is one test that is failing because the tested logic involves joining different mock datasets. The join doesn't return any common rows, therefore the test fails. 

**This PR aims to make the automatically generated datasets more useful by adding a constraint in the ID fields of each dataset so that joins will be successful.** 
After this PR is merged, the failing test in the above mentioned PR will no longer fail.

- All main ID fields in each dataset will be between 1 and 400 (the number of rows we randomly generate)
- Other minor constraints like providing a list of all possible study types

## 🚦 Before submitting


tests/gentropy/conftest.py
```diff
@@ -76,6 +76,14 @@ def mock_colocalisation(spark: SparkSession) -> Colocalisation:
             randomSeedMethod="hash_fieldname",
         )
         .withSchema(coloc_schema)
+        .withColumnSpec(
```
project-defiant, line 79: I am not sure I follow why you define the cast explicitly here
```
In [61]: df = data_spec.withSchema(schema).build().select('leftStudyLocusId', 'rightStudyLocusId').printSchema()
root
 |-- leftStudyLocusId: string (nullable = false)
 |-- rightStudyLocusId: string (nullable = false)
```

project-defiant (approved):

PR #957, opened 2024-12-16
fix: repair SusieFinemapperStep to work with new SL schema and fix lo…

Daniel-Considine:
## ✨ Context

In some cases the SusieFinemapperStep was not correctly assigning the variant with the highest posteriorProbability as the lead. This was because the code was sorting the column as a string instead of as a float/double. 

This PR fixes the issue by casting the posteriorProbability column as a double. 

Additionally, it also fixes the schema for one of the other functions which was causing issues when 'confidence' was added to the StudyLocus schema. This is a temporary fix to ensure the function works, and a more robust change with tests will be implemented when I tidy up the finemapping step/functions.

### DIFF
--- a/src/gentropy/susie_finemapper.py
+++ b/src/gentropy/susie_finemapper.py
@@ -13,9 +13,7 @@ import scipy as sc
 from pyspark.sql import DataFrame, Row, Window
 from pyspark.sql.functions import desc, row_number
 from pyspark.sql.types import (
-    ArrayType,
     DoubleType,
-    FloatType,
     IntegerType,
     StringType,
     StructField,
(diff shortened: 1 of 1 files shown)


src/gentropy/susie_finemapper.py
addramir, line 1: I feel that it is not a great idea to specify a big schema explicitly. Is there a way to avoid it? 
@project-defiant what do you think about it?

src/gentropy/susie_finemapper.py
```diff
@@ -896,7 +898,63 @@ def susie_finemapper_one_sl_row_gathered_boundaries(  # noqa: C901
             logging.warning("Analysis Flags check failed for this study")
             return None
 
-        schema = StudyLocus.get_schema()
+        schema = StructType(
```
project-defiant, line 901: @Daniel-Considine what is the difference between this schema and StudyLocus schema?

src/gentropy/susie_finemapper.py
project-defiant: I posted a comment, lets see if we can use the original StudyLocus schema

src/gentropy/susie_finemapper.py
Daniel-Considine: It was breaking because the locus column from LocusBreaker (specifically `annotate_locus_statistics_boundaries`) has less columns than the StudyLocus schema.

project-defiant (approved):

PR #971, opened 2025-01-14
chore: removing symbols from QTL study identifiers

DSuveges:
## Context

When concatenating various fields of QTL studies to generate the study identifier, there is an unexpected side effect: the resulting string might contain special characters that interfere with using study identifiers as URLs. To resolve this problem, upon ingestion, study id string is lower cased, and all non-alphanumeric (plus `_` and `-`) characters are replaced with underscores.


tests/gentropy/data_samples/QTD000584.credible_sets.tsv
```diff
@@ -1,17 +1,17 @@
 molecular_trait_id	gene_id	cs_id	variant	rsid	cs_size	pip	pvalue	beta	se	z	cs_min_r2	region
-RBP7.14208.3.3..1	ENSG00000162444	RBP7.14208.3.3..1_L1	chr1_10000458_T_C	rs3003377	22	0.0170703689724805	1.33564e-10	0.139135	0.0215922	6.45120520807609	0.649351082174683	chr1:8997206-10997206
```
DSuveges, line 2: The sample files were changed to provide a more complete representation of the issues caused by the special characters in the studyId.

ireneisdoomed: Nice, thank you!

ireneisdoomed (approved):
All good! Thank you for the changes!

PR #973, opened 2025-01-15
feat(qtls): flagging trans QTL credible sets

DSuveges:
## ✨ Context

QTL credible stets can be in trans or in cis position respective to the relative location of the measured gene. A QTL credible set is considered trans if:

- All taggging variants are in trans position
- A variant is in trans position if the chromosome is different from the target chromosome
- A variant is in trans position if the distance from the TSS of the gene is higher than a provided threshold (eg 5e6)

A newly introduced boolean column is added to studyLocus schema. This column is nulls for credible sets of gwas (non-qtl) studies.


DSuveges:
This PR depends on merging #946

src/gentropy/dataset/study_locus.py
```diff
@@ -681,6 +682,91 @@ def get_QC_mappings(cls: type[StudyLocus]) -> dict[str, str]:
...
+                f.when(
+                    (f.col("chromosome") != f.col("geneChromosome"))
+                    | (f.abs(f.col("tss") - f.col("position")) > trans_threshold),
+                    f.lit(True),
+                ).otherwise(f.lit(False)),
+            )
+            .groupby("studyLocusId")
+            .agg(
+                # If any of the tags of a locus is in trans position, the QTL is considered trans:
+                f.array_contains(f.collect_set("isTagTrans"), True).alias("isTransQtl")
```
vivienho, line 761: Here it states that the QTL is considered trans if at least one tag is trans, but in the PR description it's mentioned that a QTL is considered trans when all tagging variants are trans?

tests/gentropy/dataset/test_study_locus.py
```diff
@@ -1189,3 +1191,96 @@ def test_duplication_flag_correctness(
...
+            )
+        )
+        self.study_index = StudyIndex(
+            _df=spark.createDataFrame(self.STUDY_DATA, self.STUDY_COLUMNS)
+        )
+        self.target_index = TargetIndex(
+            _df=(
+                spark.createDataFrame(self.GENE_DATA, self.GENE_COLUMNS).select(
+                    f.struct(
+                        "strand",
```
vivienho, line 1240: The first `strand` seems redundant

vivienho (approved):
The PR adds a function to flag trans qtls during study locus validation. All makes sense!

src/gentropy/dataset/study_locus.py
DSuveges: Yes, thank you for finding this discrepancy. I have changed my mind, but haven't updated the description.

tests/gentropy/dataset/test_study_locus.py
DSuveges: Oh, yes, you are totally right.

DSuveges:
@vivienho thank you for reviewing the PR! I have addressed both comments. If the tests are all passing, I'll merge the branch.

PR #974, opened 2025-01-16
chore: multiple python version support with latest pyspark and hail

project-defiant:
## ✨ Context

This PR closes [#3189](https://github.com/opentargets/issues/issues/3189) 
This PR closes [#3680](https://github.com/opentargets/issues/issues/3680)
 and links to https://github.com/opentargets/orchestration/pull/94

This PR focuses on allowing pyspark>=3.5.0, <3.6 along with newest hail version, that supports it.
The changes above caused trouble (errors) for poetry to resolve the dependencies. This lead to more developments, the full list is described below:

- [x] replaced poetry with uv as a dependency management tool (~5x faster dependency resolution without issues)
- [x] added support for `X.Y.Z-devXX` releases on push to `dev` branch
- [x] updated way to create dev dataproc cluster (Open Targets users only) - see details below
- [x] add support for multiple python versions in gentropy (py310, py311, py312) py313 support failure due to missing [BLAS when installing scipy from wheel](https://stackoverflow.com/questions/69954587/no-blas-lapack-libraries-found-when-installing-scipy)
- [x] support for pyspark>=3.5.0, <3.6
- [x] bump hail version to sync with pyspark dependency
- [x] dropped `google` as dependency, added more granular `google-cloud-storage` (To be removed in future release with other google dependent packages)
- [x] github actions update to use uv and test for dependency matrix of python versions (requires changes in the repository rules) 

### dev cluster setup

Previously dev cluster was created from the package pushed to the google cloud storage bucket under the namespace that matched the git ref name. This has now changed, as cluster now install gentropy directly from the github branch (or tag) specified in the orchestration - see changes in https://github.com/opentargets/orchestration/pull/94. 

#### Benefits of new solution
 - keeping just one copy of the static files used for running gentropy steps(install_dependencies_on_cluster.sh and `cli.py`) in the GCS.
- spped up the development process, since the one does not need to wait for the github actions to finish, before they can set up the cluster from airflow on the branch.
- speed up the dependency resolution with `uv pip install` (pip resolution of gentropy with support for multiple python versions takes ~ 20minutes leading to initialization actions timeout, uv does the same thing with around 1.5min).

## 🚦 Before submitting


ireneisdoomed (approved):
Thank you!
2 questions:
1. Should we also give support to pyspark versions lower than 3.5? Afaik, there is nothing in the API that is not compatible. I know we have custom code to compare dataframes, something that Pyspark 3.5 supports
2. We need to update the Dataproc version too. Should we make a pipeline run to test nothing breaks?

project-defiant:
> Thank you! 2 questions:
> 
> 1. Should we also give support to pyspark versions lower than 3.5? Afaik, there is nothing in the API that is not compatible. I know we have custom code to compare dataframes, something that Pyspark 3.5 supports
> 2. We need to update the Dataproc version too. Should we make a pipeline run to test nothing breaks?

These are very nice points!

1. I am aftraid, that poetry will not allow us to supoport more pyspark versions, latest [hail requires version between 3.5 to 3.6](https://github.com/hail-is/hail/blob/4c60fddb171a52c21f41a81995c53a28e375c26b/hail/python/requirements.txt#L14), otherwise we could do it.
2. Yes I am planning to do that, just right after I finish the harmonisation :). Let's wait for this check before we merge.

utils/clean_status.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+echo "Fetching version changes..."
+git fetch
+if output=$(git status --porcelain) && [[ -n "$output" ]]; then
```
ireneisdoomed, line 5: Nice, I didn't know about `--porcerlain` :D

pyproject.toml
```diff
@@ -9,28 +9,30 @@ authors = [
...
+    "scikit-learn (>=1.6.1, <1.7.0)",
+    "pandas[gcp,parquet] (>=2.2.3, <2.3.0)",
+    "skops (>=0.11.0, <0.12.0)",
+    "shap (>=0.46, <0.47)",
+    "matplotlib (>=3.10.0, <3.11.0)",
+    "google-cloud-secret-manager (>=2.12.6, <2.13.0)",
+    "google-cloud-storage (>=2.14.0, <2.15.0)"
+]
+classifiers = [
+  "Programming Language :: Python :: 3.10",
```
ireneisdoomed, line 32: 🥳

Dockerfile
ireneisdoomed, line 1: The Gentropy image will be automatically updated, right?

utils/install_dependencies.sh
```diff
@@ -1,33 +1,18 @@
...
-    PYTHON_VERSION=$(cat .python-version)
-    pyenv install --skip-existing $PYTHON_VERSION
-    pyenv shell $PYTHON_VERSION
-else
-    echo ".python-version file not found."
-    exit 1
+if ! command -v uv &>/dev/null; then
+    echo "uv was not found, installing uv..."
+    curl -LsSf https://astral.sh/uv/install.sh | sh
+    source $HOME/.local/bin/env
```
ireneisdoomed, line 7: After installing the environment, `uv` wasn't on my path. Should we add `export PATH="$HOME/.local/bin:$PATH"` here?

Makefile
```diff
@@ -29,37 +30,48 @@ setup-dev: ## Setup development environment
...
 	@echo "Linting docstrings..."
-	@poetry run pydoclint --config=pyproject.toml src
-	@poetry run pydoclint --config=pyproject.toml --skip-checking-short-docstrings=true tests
+	@uv run pydoclint --config=pyproject.toml src
+	@uv run pydoclint --config=pyproject.toml --skip-checking-short-docstrings=true tests
 
 test: ## Run tests
 	@echo "Running Tests..."
-	@poetry run pytest
+	@uv run pytest
```
ireneisdoomed, line 40: Please test this yourself, but the current command goes idle without running any tests
```suggestion
	@uv run pytest .
```

ireneisdoomed (approved):
Really cool!! New PySpark, support for several Python versions, and migration to UV.

image
I suggest changing the PR title, this is clearly a lie haha 

Setting up the dev environment was smooth simply by doing `make setup-dev` (except that uv wasnt in my path, see minor comment below).

I haven't used `uv` myself, so I can't really evaluate pros and cons. M only questions is to confirm that we are not missing anything from CI/CD after the migration?

Makefile
project-defiant: The tests still go ok on my side, although it seems that the long lookup, not sure why the dot helps here, assuming that we already add the `testpath(s)` to the pytest options, the dot just overwrites it. While testing your solution I got following errors cache related:
``` 
import file mismatch:
imported module 'a_creating_spark_session' has this __file__ attribute:
  /home/mindos/Projects/OpenTargets/gentropy/docs/src_snippets/howto/python_api/a_creating_spark_session.py
which is not the same as the test file we want to collect:
  /home/mindos/Projects/OpenTargets/gentropy/site/src_snippets/howto/python_api/a_creating_spark_session.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

This seems to be due to the fact that I have previously generated the docs that contain some test duplicate.

On the note of test collection speeds:

```
(base)  mindos@mindos  ~/Projects/OpenTargets/gentropy   pyspark-bump ±  time uv run pytest --collect-only . 1> /dev/null
uv run pytest --collect-only . > /dev/null  11,81s user 2,56s system 115% cpu 12,390 total
(base)  mindos@mindos  ~/Projects/OpenTargets/gentropy   pyspark-bump ±  time uv run pytest --collect-only 1> /dev/null 
uv run pytest --collect-only > /dev/null  11,70s user 2,62s system 116% cpu 12,269 total
```

utils/install_dependencies.sh
project-defiant: This is a good point! The UV setup script looks over 3 locations (the `$HOME/.local/bin/{uv,uvx}` is the last fallback if `XDG_BIN_HOME` nor `XDG_DATA_HOME` are not defined, which is one of the standards, that most [Unix based machines](https://specifications.freedesktop.org/basedir-spec/latest/#index) follow (apparently not Mac). 

I will make additional fix for this !

project-defiant: I looked over how uv is set up, the installation path is specified at runtime, so referencing it means recreating the lookup I described above. On the good note, the uv adds itself to the PATH in the shell rc file, so ideally one just need to source the shell rc to have it run as executable. There should be no need to add the `${HOME}/.local/bin/` to the path, as it might not be correct by default on linux machines.

project-defiant: To explain in short:
-  uv sets the export for the env file that holds the paths to the `uv` and `uvx`.
-  You just have to source the shellrc file after the succesful installation to make it available in the terminal (this is due to the fact that Make runs in a separate shell and parent shell does not inherit from child shell).

Dockerfile
project-defiant: yes, this does not change

ireneisdoomed (approved):
Thank you for addressing the comments!

PR #989, opened 2025-02-05
fix: foldX normalisation to cover full range

DSuveges:
## ✨ Context

With the latest PR (#947) a bug got introduced into Fold-X energy normalisation. This fix ensures the full free energy range is covered.


vivienho (approved):
Thank you for fixing!

PR #1001, opened 2025-02-24
chore(trainer): sample train set to base model importances on

ireneisdoomed:
## ✨ Context
SHAP calculation on the samples from the training set makes the L2G train step to run for over 1h.
Downsampling the training set for SHAP calculation to 1000 instances makes the step to run in 10min ([job](https://console.cloud.google.com/dataproc/jobs/2e14da53-3d3b-4478-9442-71a6bd2b2326/configuration?region=europe-west1&authuser=1&inv=1&invt=AbqcZw&project=open-targets-genetics-dev))

This partially solves the issue described here https://github.com/opentargets/issues/issues/3779

SHAP calculation happens here to understand the global feature importances, and to understand general model patterns.
Because the training set is ~200k samples, we can downsample for the SHAP calculation. We lose granularity, but makes training much faster.

We have less datapoints, but I think this approach is still useful to grasp general patterns in the model. The predictions dataset will have the full list.
image

## 🚦 Before submitting


project-defiant:
@ireneisdoomed, I guess this is just an approximation. By looking at the SHAP docs, I have found that the GPUTreeExplainer can use the CUDA support for the calculations, although it would need some more investigation if we could use it.

PR #1008, opened 2025-02-28
feat: adding metadata from dataset schemas

DSuveges:
## ✨ Context

Column level metadata can be defined/added based on the dataset schema JSON files. This information then can be baked into the exported data, so any time the dataset is opened, the column level metadata information is available. This can also be directly used for generating data documentation via ml croissant.

## Contents

1. Variant index dataset annotated with column level metadata.
2. Schemas package is updated to extract metadata from the expected schema and added to observed schema.
3. Dataset.validate method optionally propagates the metadata to the dataset.

After the dataset is saved, the metadata is in the parquet files.

### DIFF
--- a/src/gentropy/assets/schemas/variant_index.json
+++ b/src/gentropy/assets/schemas/variant_index.json
@@ -1,37 +1,49 @@
 {
   "fields": [
     {
-      "metadata": {},
+      "metadata": {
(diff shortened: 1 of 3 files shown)


DSuveges:
No longer relevant.

PR #1023, opened 2025-03-18
feat: slimmer docker image

ffrnz:
## ✨ Context

This PR optimizes the Docker image for the gentropy project by implementing a multi-stage build process and upgrading dependencies, resulting in performance improvements and a smaller image size.

This PR introduces the following changes to the Dockerfile:

- Upgraded base image from Python 3.10-bullseye to Python 3.11-bookworm/slim-bookworm
- Implemented a multi-stage build:
  - Builder stage: Handles dependency installation and package building
  - Runtime stage: Creates a minimal final image
- Improved dependency management:
  - Replaced direct pip install with UV installation script
  - Added virtual environment isolation
- Optimized image size:
  - Reduced final image size by approximately 50% (from ~2.4GB to ~1.2GB)
  - Better layer caching with separate dependency and source copy steps
- Performance improvements:
  - Faster container startup times
  - Significant speed increase for batch jobs due to smaller image and updated dependencies
- Changed entrypoint from `uv run gentropy` to direct `python cli.py` execution
- Added proper environment variables for PySpark and Hail

These changes result in:
- mostly, the benefits are noticeable for batch task, where the reduced image size leads to quicker runtimes. 

## 🙈 Missing

- This would also require changes in the orchestration layer, i.e. the commands that run the tasks need to be changed for `/bin/sh -c python cli.py`. See https://github.com/opentargets/orchestration/pull/112

## 🚦 Before submitting


project-defiant:
Dear @ffrnz thank you for the continuous interest in our pacakge improvement. 

I have one question upon the PR, as you are trying to update the java version in the gentropy docker container to openjdk-17, I am not sure if all functionalities provided by Hail will be compatible then - see [hail requirements](https://hail.is/docs/0.2/install/macosx.html). I had attempted to run gentropy tests with `Temurin-17.0.14+7` wihout any failure, but can not guarantee that everything will work after the update.

ffrnz:
> Dear @ffrnz thank you for the continuous interest in our pacakge improvement.
> 
> I have one question upon the PR, as you are trying to update the java version in the gentropy docker container to openjdk-17, I am not sure if all functionalities provided by Hail will be compatible then - see [hail requirements](https://hail.is/docs/0.2/install/macosx.html). I had attempted to run gentropy tests with `Temurin-17.0.14+7` wihout any failure, but can not guarantee that everything will work after the update.
> 
> With regards to the container image size I am looking forward to see the speedup while using the multi-stage builds. We are currently using the image for fine-mapping and harmonisation in the orchestration repo (as you might already know). Due to the inability of installing the Cloud Ops Agent at the cos images (that blocks us from collecting the telemetry for RAM usage on VMs during batch run) I am reflecting wether we might stop using the containerised google batch job and roll back to script usage.

Good point, I blindly assumed backward compatibility, but as some libraries are usually replaced between versions, functionality cannot be 100% guaranteed. I'll revert to jdk-11 for safety, then :) 

Interesting point on the telemetry, I can't comment on GCP, as I run exclusively on AWS, which has me covered for the most part with Container Insights. Script usage sounds like it would introduce quite a significant overhead though? Maybe a custom solution using gmond / ganglia is an option.

project-defiant:
Dear @ffrnz, let me know if the PR is ready (switch from draft), cheers!

ffrnz:
> Dear @ffrnz, let me know if the PR is ready (switch from draft), cheers!

Sure, just wanted to implement the necessary changes in the orchestration layer first:
https://github.com/opentargets/orchestration/pull/112

Dockerfile
```diff
@@ -1,20 +1,62 @@
...
-COPY pyproject.toml uv.lock ./
+# Copy project configuration and source code
+COPY ./pyproject.toml .
+COPY ./src src
+
+# Create an empty README file (required by some package managers)
 RUN touch README.md
 
+# Synchronize dependencies using UV (installs requirements from pyproject.toml)
 RUN uv sync
```
project-defiant, line 30: I think it would be beneficial to preserve and use the `uv.lock` file for the build instead of regenerating it. This will ensure the dependencies are not updated during the docker build (although it is relatively unlikely to happen)
```suggestion
COPY ./src src
COPY ./uv.lock uv.lock

# Create an empty README file (required by some package managers)
RUN touch README.md

# Synchronize dependencies using UV (installs requirements from pyproject.toml)
RUN uv sync --frozen

```

Dockerfile
```diff
@@ -1,20 +1,62 @@
...
 
-COPY src ./src
+# Install the package (and its dependencies) into a virtual environment
+RUN uv pip install .
+
+# Stage 2: Runtime stage - Creates the final minimal image
+FROM python:3.11-slim-bookworm
+
+# Install Java runtime (required for PySpark/Hail) and procps for process management
+RUN echo "deb http://deb.debian.org/debian oldstable main" >> /etc/apt/sources.list
```
project-defiant, line 38: From build perspective, the java version update will change less frequently then gentropy. Do you know if there is a way to move it to the build step. I have experimented with 
```
COPY --from=builder /usr/lib/jvm /usr/lib/jvm
COPY --from=builder /etc/alternatives /etc/alternatives
COPY --from=builder /usr/bin/java /usr/bin/java
```
But some of the libraries are still missing after that.

project-defiant (commented):
Thank you very much for your contribution. This change will significantly decresase the image size and speed up build time for our container. Please have a look at the comments. When resolved I am happy to review again.

Dockerfile
ffrnz: Hey @project-defiant,

Thanks for the feedback! I’ve refactored the build process to ensure we use the same base image with Java installed, then add only the UV binaries from the official corresponding UV Docker image.

Additionally, by installing the project into the virtual environment, we no longer need to copy the cli.py file separately. Let me know what you think—I’d love to hear your thoughts!

project-defiant (approved):
Thank you for the contribution to the gentropy codebase, this change will increase the build time of our image.

PR #1034, opened 2025-04-15
feat: tweaking colocalisation step to limit the right-hand side of the overlaps step

Daniel-Considine:
## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/study_locus.py
+++ b/src/gentropy/dataset/study_locus.py
@@ -488,14 +488,14 @@ class StudyLocus(Dataset):
     def _overlapping_peaks(
         credset_to_overlap: DataFrame,
         intra_study_overlap: bool = False,
-        molecular_qtl_only: bool = False,
+        restrict_to_studyIds: list[str] = [],
     ) -> DataFrame:
         """Calculate overlapping signals (study-locus) between GWAS-GWAS and GWAS-Molecular trait.
 
         Args:
             credset_to_overlap (DataFrame): DataFrame containing at least `studyLocusId`, `studyType`, `chromosome` and `tagVariantId` columns.
             intra_study_overlap (bool): When True, finds intra-study overlaps for credible set deduplication. Default is False.
-            molecular_qtl_only (bool): When True, does not find overlaps for GWAS vs GWAS. Defaults to False.
+            restrict_to_studyIds (list[str]): List of studyIds to restrict finding overlaps on the right side. Default is empty list.
 
         Returns:
             DataFrame: containing `leftStudyLocusId`, `rightStudyLocusId` and `chromosome` columns.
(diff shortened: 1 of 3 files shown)


src/gentropy/dataset/study_locus.py
```diff
@@ -486,13 +486,16 @@ def qc_abnormal_pips(
...
+        restrict_to_studyIds: list[str] | None = None,
+        qtl_study_overlap: bool = False,
     ) -> DataFrame:
         """Calculate overlapping signals (study-locus) between GWAS-GWAS and GWAS-Molecular trait.
 
         Args:
             credset_to_overlap (DataFrame): DataFrame containing at least `studyLocusId`, `studyType`, `chromosome` and `tagVariantId` columns.
-            intra_study_overlap (bool): When True, finds intra-study overlaps for credible set deduplication. Default is False.
+            restrict_to_studyIds (list[str] | None): List of studyIds to restrict finding overlaps on the right side. Default is None.
+            qtl_study_overlap (bool): When True, finds overlaps with just molecular-QTLs on the right-hand side. Default is False.
```
ireneisdoomed, line 498: This is GWAS vs mQTLs only, right? It is not clear that this doesn't compute mQTLxmQTL. Would it be interesting?

ireneisdoomed (commented):
Thank you! This is a good approach. Filtering by credible set type or by studies makes sense and makes the step much more flexible.
I still see a problem in how you pass the list of studies, though. Do you think it'd make sense if this list were generated dinamically in the coloc step itself? Maybe theres a better solution but the step could have 2 inputs: 
1. list of credible sets (including new ones we haven't colocalised)
2. optional colocalisation dataset

To compute, you find all the overlaps, and drop those that match up with the pairs already present in the colocalisation dataset. This solution is not optimal as it still requires computing overlaps.

My q is, did you consider how to approach this?

src/gentropy/dataset/study_locus.py
Daniel-Considine: `qtl_study_overlap` will only do GWAS vs molQTLs correct. 
I can see how this is confusing now that you mention it - would you recommend `qtl_right_side` or something similar as the variable name? 
@Tobi1kenobi has previously mentioned that at some point it might be interesting to do QTL vs QTL overlaps too, so this could also be added as a potential option. 

`restrict_to_studyIds`  can do GWAS vs molQTLs (if you've specified QTL studyIds) and can do GWAS vs GWAS if you've specified a GWAS study ID on the right. I'll change this to `restrict_right_studyIds` to also be more explicit.

Daniel-Considine:
@ireneisdoomed 
I'm not a fan of calculating the list dynamically within the step, especially if it involves calculating overlaps and filtering - one of the main benefits of these options is that majorly reduces the computation required to do the step. For example, previous coloc implementation for 36k sceqtl (eQTLGen) credible sets took ~50min on airflow w/ autoscaling using large/expensive machines. Whereas running just GWAS vs sceqtl took less than 3min airflow with smaller machines & autoscaling. This could probably even be run on a laptop. 

Instead of passing the list of studyIds directly to the ColocalisationStep, it may be better storing them as a csv/json and the path is passed to the Step, and then it's read and passed to the downstream functions. What do you think?

project-defiant (commented):
@Daniel-Considine I get the two points that this PR tries to achieve:
1. Allow for running the ColocalisationStep with for example 1 GWAS studyId
2. Allow for skipping the GWAS-GWAS colocalisation.

This are valid use cases, but enriching this step might cause more trouble around the usage then benefit.

What I suggest is to create a new step that will allow exactly what you want, rather then adding the flags here and then properly documenting the use case:

1. ColocalisationStep allows for running the colocalisation in bulk
2. ColocaliseGwasToQTL allows for defining the studyIds that one want to colocalise agains the qtls
* These are example names (not the best ones)

project-defiant (approved):
Great work on the tests, looks very clean and solid.

PR #1068, opened 2025-07-01
chore: adding pubmed identifier to evidence

DSuveges:
## ✨ Context

Addressing [#3936](https://github.com/opentargets/issues/issues/3936): 
- Carrying over pubmed identifier from the study index to evidence
- Adding some tests to cover evidence generation.

### DIFF
--- a/src/gentropy/dataset/l2g_prediction.py
+++ b/src/gentropy/dataset/l2g_prediction.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING
 import pyspark.sql.functions as f
 import shap
 from pyspark.sql import DataFrame
-from pyspark.sql.types import StructType
+from pyspark.sql.types import StringType, StructType
 
 from gentropy.common.schemas import parse_spark_schema
 from gentropy.common.session import Session
@@ -22,7 +22,6 @@ from gentropy.method.l2g.model import LocusToGeneModel
 
 if TYPE_CHECKING:
     from pandas import DataFrame as pd_dataframe
-    from pyspark.sql.types import StructType
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/gentropy/dataset/l2g_prediction.py
```diff
@@ -120,11 +120,16 @@ def to_disease_target_evidence(
...
-            DataFrame: Disease target evidence
+            DataFrame | None: Disease target evidence
         """
         datasource_id = "gwas_credible_sets"
         datatype_id = "genetic_association"
 
+        # A set of optional columns need to be in the input datasets:
+        for c in ["diseaseIds", "pubmedId"]:
+            if c not in study_index.df.columns:
+                return None
```
ireneisdoomed, line 131: Why would you return `None` instead of raising an error?
I think I would check that the output contains the columns, not the input. What do you think?

src/gentropy/l2g.py
```diff
@@ -419,16 +419,16 @@ def __init__(
         # Reading the study index
         study_index = StudyIndex.from_parquet(session, study_index_path)
 
-        # Generate evidence and save file:
-        (
-            locus_to_gene_prediction.to_disease_target_evidence(
-                credible_sets, study_index, locus_to_gene_threshold
+        # Generate evidence and save file - if there's any evidence:
```
ireneisdoomed, line 422: Related to the comment above, do you think that the lack of study metadata is a big enough reason not to return evidence?

tests/gentropy/dataset/test_l2g.py
ireneisdoomed, line 1: Thank you so much for this test!! <3

ireneisdoomed (commented):
Please see my comments below!

DSuveges:
**Comments addressed:**

- Evidence generation will not return `None` => Step logic updated.
- diseaseIDs is required column => if missing, Value error is raised.
- pubmedId is optional column => if missing, evidence still generated with nulls in the literature field.
- tests updated to reflect changes listed above.

PR #1108, opened 2025-08-29
build(docker,xgboost): update docker build

project-defiant:
## ✨ Context

This PR attempts to close https://github.com/opentargets/issues/issues/4005

There is a bug in the way how we build the docker image for gentropy, which is due to the fact that the `openjdk-11-jdk-headless` is no longer available from the apt-repo and has been replaced by `openjdk-17-jdk-headless`. 

We can not change the java version, **because hail does not support java > 11**. Hence we need an alternative method to install java. 

We also found that introduction of the `xgboost` library brought ~1Gb of depencency size, since we do not use all of the `xgboost` features we can easily limit ourselves to the `xgboost-cpu` library that is much smaller in size reducing the total size of the docker image from 2.8G to 1.8G.

Additionally, since we use the docker image only for the harmonisation batch job, we can limit the number of artifacts on the Google Artifact Registry 
image_720

and remove entire `gentropy-app` repo. The optimal solution would be to build the image only on demand by triggering github actions from specific tag, rather then on each branch update. This will prevent the registry from exploding the number of already large images. 

This PR implements following:

- [x] Refactoring of gentropy docker image so it copies the java from `amazoncorretto:11.0.28-al2023-headless` image
- [x] Aligning image build with the [uv guidelines](https://github.com/project-defiant/uv-docker-example/blob/main/multistage.Dockerfile)
- [x] Now github action generating the artifact will be only triggered from the github ui (wokrlfow_dispatch) rather then build the image on each push reducing the size of the GAR (google artifact registry)
- [x] Github action generating the artifact no longer uses the long-term credentials stored in the secrets, rather uses the OIDC provider to allow github runner accessing the GAR -   [see more](https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account)
- [x] Docker image is now also build to the Github Container Registry (GHCR)
- [x] Addition of the target to test if docker build command works correctly
- [x] Swap of `xgboost`  to `xgboost-cpu`  

> [!NOTE]
> The new image in GAR will now be stored under different registry `opentargets` instead of `gentropy-app`. This repository is set up with Workload Identity Provider to make sure that the authentication no longer requires long term stored credentials.

## 🚦 Before submitting


pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   "matplotlib (>=3.10.0, <3.11.0)",
   "google-cloud-secret-manager (>=2.12.6, <2.25.0)",
   "google-cloud-storage (>=2.14.0, <3.3.0)",
-  "xgboost>=3.0.3",
+  "xgboost-cpu>=3.0.4",
```
DSuveges, line 27: I'm personally not sure about this one. I mean is the cpu only as performant as the one that supports gpu utilisation? Probably in our case it doesn't matter, because the cluster doing the computation doesn't have gpu anyway. It's fine.

Dockerfile
```diff
@@ -1,46 +1,37 @@
...
 
-# Set working directory in the runtime container
-WORKDIR /app
+# # Stage 2: Runtime stage - Creates the final minimal image
+FROM python:3.12.11-slim-trixie AS production
 
-# Copy the virtual environment with all dependencies from the builder stage
-COPY --from=builder /app/.venv .venv
+# # Set working directory in the runtime container
+COPY --from=uv_builder --chown=app:app /app /app
```
DSuveges, line 25: I don't get the concept of the different stages. How many images are we building here? It's not a cardinal issue, as this design was already in place.

pyproject.toml
project-defiant: Yes, I am not sure 100%.  Maybe it would be good to ask @ireneisdoomed if it works fine this way

.github/workflows/artifact.yml
```diff
@@ -2,88 +2,95 @@ name: Build and Push to Artifact Registry
...
-      - "*"
-    tags:
-      - "v[0-9]+.[0-9]+.[0-9]+-*"
 
 env:
-  PROJECT_ID: open-targets-genetics-dev
-  REGION: europe-west1
-  GAR_LOCATION: europe-west1-docker.pkg.dev/open-targets-genetics-dev
-  REPOSITORY: gentropy-app
-  PYTHON_VERSION_DEFAULT: "3.11.11"
```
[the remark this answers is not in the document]
project-defiant: Yes. We now use default python from image which is 3.12.  We do not need to build the package anymore

DSuveges (approved):
A smart re-design of the automated build process of the repository, which allows for a leaner image.

PR #1187, opened 2026-02-04
feat: initial commit with logic to run LDSC h2 estimates

xyg123:
## ✨ Context

To add LD score regression functionality to gentropy (https://github.com/opentargets/issues/issues/4136), starting with h2 estimates.

Core logic to calculate h2 using precomputed LD scores.
Incorporate h2 estimates into summary statistics qc step.

## 🙈 Missing

Update paths to precomputed LD scores, weights and hapmap3 SNP inputs required for h2 calculations.

## 🚦 Before submitting


xyg123:
The logic for h2 estimates has been ported and refactored from the ldsc python3 implementation.

Due to the requirements of sample size and ancestry information, this step cannot be completed at the summary stats QC stage.

Instead, the step requires study index as one of the inputs, and there is a separate PR for orchestration to run each study as batch jobs.

src/gentropy/assets/schemas/study_index.json
```diff
@@ -288,6 +288,65 @@
       },
       "nullable": true,
       "metadata": {}
+    },
+    {
+      "name": "ldscH2",
```
project-defiant: If we are amending to the schema, I would not bind the name of the method into the column name. Better would be to make it generic, or include in the sumstatQCValue array of structs.

src/gentropy/assets/schemas/study_index.json
```diff
@@ -288,6 +288,65 @@
...
+            "nullable": true,
+            "metadata": {}
+          },
+          {
+            "name": "M_ldsc",
+            "type": "double",
+            "nullable": true,
+            "metadata": {}
+          },
+          {
```
project-defiant: What is the difference between M and n_snps ?

src/gentropy/dataset/study_index.py
```diff
@@ -497,6 +497,39 @@ def validate_target(self: StudyIndex, target_index: TargetIndex) -> StudyIndex:
...
+        """Collecting heritability information from LDSC results.
+
+        Args:
+            heritability_df (DataFrame): A dataframe containing heritability information for a subset of studies.
+
+        Returns:
+            StudyIndex: with heritability information annotated.
+        """
+        h2_annotations = (
+            heritability_df.filter(f.col("runStatus") == "success")
```
project-defiant, line 513: Coalesce the heritability (very small table) & broadcast it.

src/gentropy/dataset/study_index.py
```diff
@@ -497,6 +497,39 @@ def validate_target(self: StudyIndex, target_index: TargetIndex) -> StudyIndex:
...
+                    f.col("h2_se"),
+                    f.col("intercept"),
+                    f.col("intercept_se"),
+                    f.col("mean_chisq"),
+                    f.col("lambda_gc"),
+                    f.col("n_snps_used").cast("long"),
+                    f.col("M_ldsc"),
+                    f.col("ld_ancestry"),
+                ).alias("ldscH2"),
+            )
```
project-defiant, line 525: coalesce afterwards to 1 partition and sort within the partition by studyType, projectId, studyId (will dramatically downstream)

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            row["studyId"] for row in sumstats_df.select("studyId").distinct().collect()
+        ]
+
+        if len(study_ids) != 1:
+            raise ValueError(
+                f"Expected one study in summary statistics, got {len(study_ids)}: {study_ids}"
+            )
+
+        return study_ids[0]
+
```
project-defiant, line 220: Do not do that, one could potentially want to run the heritability on multiple summary statistics, then it will break.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        self.min_samples = min_samples
+        self.m_ldsc_override = m_ldsc_override
+        self.results: dict[str, Any] | None = None
+
+        self._run()
+
+    def _run(self) -> None:
+        """Execute the LDSC heritability estimation pipeline."""
+        sumstats_df = self._read_sumstats()
+        study_id = self._extract_single_study_id(sumstats_df)
```
project-defiant, line 80: As below, do not expect that dataset can have only one studyId.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            skip_reasons.append("Could not infer ancestry")
+
+        return {
+            "run_status": "skipped" if skip_reasons else "success",
+            "skip_reasons": skip_reasons,
+            "analysis_flags": analysis_flags,
+            "ancestry": ancestry,
+        }
+
+    @staticmethod
```
project-defiant, line 594: Again, there is an expectation that this will just run one study at a time, I am not sure if that should be the case.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        if analysis_flags is None:
+            return []
+
+        normalised: list[str] = []
+        for flag in analysis_flags:
+            if flag is None:
+                continue
+            normalised.append(str(flag).strip().lower())
+
+        return normalised
```
project-defiant, line 707: What this normalization is needed for ?

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        row = (
+            study_index_df.filter(F.col("studyId") == study_id)
+            .select("ldPopulationStructure", "analysisFlags", "nSamples")
+            .first()
+        )
+
+        if row is None:
+            raise ValueError(f"studyId {study_id} not found in study index")
+
+        analysis_flags = self._normalise_analysis_flags(row["analysisFlags"])
```
project-defiant, line 568: Note, that the analysisFlags are already predefined in the StudyAnalysisFlag enum, please rely on these flags, instead of re-parsing them here.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            - no flags
+            - flags are a subset of the allowed set (exwas, wgsgwas, metabolite)
+
+        Args:
+            analysis_flags (Any): Raw analysis flags.
+
+        Returns:
+            bool: Whether the study is allowed.
+        """
+        allowed_flags = {"exwas", "wgsgwas", "metabolite"}
```
project-defiant, line 723: For that check, reuse the StudyAnalysisFlag enum.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        if ld_pop_struct is None:
+            raise ValueError("ldPopulationStructure is None, cannot infer ancestry")
+
+        if not isinstance(ld_pop_struct, Iterable):
+            raise TypeError(
+                f"ldPopulationStructure has unexpected type {type(ld_pop_struct)}; "
+                "expected iterable of population structures."
+            )
+
+        pop_map = {
```
project-defiant, line 654: Make this as a constant map at the top of the module.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+                f"Could not map any populations from ldPopulationStructure: {ld_pop_struct}"
+            )
+
+        max_weight = max(agg.values())
+        tied = [pop for pop, weight in agg.items() if weight == max_weight]
+
+        if "nfe" in tied:
+            return "nfe"
+
+        return sorted(tied)[0]
```
project-defiant, line 686: again, there is a function in gentropy that does this exactly for LD, make sure you reuse it. Also do not do it on single study basis.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        analysis_flags = self._normalise_analysis_flags(row["analysisFlags"])
+        skip_reasons: list[str] = []
+
+        if not self._is_allowed_by_analysis_flags(row["analysisFlags"]):
+            skip_reasons.append("Invalid study design")
+
+        n_samples = row["nSamples"]
+        if n_samples is None:
+            skip_reasons.append("Sample size missing")
```
project-defiant, line 576: You should include additional field (array) with flags for exclusion that will capture error codes for computing the heritability.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            skip_reasons.append(f"Sample size too small: {n_samples} < {min_samples}")
+
+        ancestry: str | None
+        try:
+            ancestry = self._infer_ld_ancestry(row["ldPopulationStructure"])
+        except Exception:
+            ancestry = None
+            skip_reasons.append("Could not infer ancestry")
+
+        return {
```
project-defiant, line 587: this would be be needed to be post joined, here you create 2 statuses (with enum) and skip_reasons (with enum), dump the data into two tables, depending on the run_status, after success is computed, rejoin both and put into the output.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            "n_snps_used": [n_snps_used],
+            "h2": [result["h2"] if result else None],
+            "intercept": [result["intercept"] if result else None],
+            "h2_se": [result["h2_se"] if result else None],
+            "intercept_se": [result["intercept_se"] if result else None],
+            "mean_chisq": [result["mean_chisq"] if result else None],
+            "lambda_gc": [result["lambda_gc"] if result else None],
+        }
+
+        schema = StructType(
```
project-defiant, line 521: Make a separate dataset with the fields (json)

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            / (F.col("nCases").cast("double") + F.col("nControls").cast("double"))
+        )
+
+        fallback_n_expr = F.when(
+            F.col("nCases").isNotNull()
+            & F.col("nControls").isNotNull()
+            & (F.col("nCases") > 0)
+            & (F.col("nControls") > 0),
+            neff_expr,
+        ).otherwise(F.col("nSamples").cast("double"))
```
project-defiant, line 288: These would be nice utililty functions.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+                ),
+            )
+            .withColumn("variant_parts", F.split(F.col("variantId"), "_"))
+            .withColumn("ref", F.col("variant_parts").getItem(2))
+            .withColumn("alt", F.col("variant_parts").getItem(3))
+            .drop("variant_parts", "nSamples", "nCases", "nControls")
+            .filter(F.col(beta_col).isNotNull())
+            .filter(F.col(se_col).isNotNull())
+            .filter(F.col(n_col).isNotNull())
+            .filter(F.col(se_col) > 0)
```
project-defiant, line 305: Any reason for that? I assume that if SE < 0, then this should be flagged somewhere else.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            .filter(F.col(chrom_col).isNotNull())
+            .filter(F.col(pos_col).isNotNull())
+            .filter(F.col("ref").isNotNull())
+            .filter(F.col("alt").isNotNull())
+            .dropDuplicates([study_col, chrom_col, pos_col, "ref", "alt"])
+        )
+
+    def _read_and_prepare_ld_scores(self, ldscore_input_path: str) -> DataFrame:
```
project-defiant, line 313: The LDScores should be it's own datasource with a valid parser.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        Raises:
+            ValueError: If required LD-score columns are missing.
+        """
+        chrom_col = "chromosome"
+        pos_col = "position"
+
+        ld_df = self.session.spark.read.csv(ldscore_input_path, header=True, sep="\t")
+
+        if "BP_hg38" in ld_df.columns:
```
project-defiant, line 331: is the BP_hg38 position in hg38 build?

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            ld_df = ld_df.withColumnRenamed("BP_hg38", pos_col)
+        else:
+            raise ValueError("LD score file must contain 'BP_hg38' column.")
+
+        if "CHR" in ld_df.columns:
```
project-defiant, line 336: Please model the input schema, rather then do one column at a time.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        missing_ld_cols = required_ld_cols - set(ld_df.columns)
+        if missing_ld_cols:
+            raise ValueError(
+                f"LD score file is missing required columns: {sorted(missing_ld_cols)}"
+            )
+
+        return (
+            ld_df.select(chrom_col, pos_col, "ref", "alt", "L2")
+            .withColumn(pos_col, F.col(pos_col).cast("int"))
+            .withColumn(chrom_col, F.col(chrom_col).cast("string"))
```
project-defiant, line 351: Probably worth ensuring the chromosome fields are 1,2,3...

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            .withColumn("L2", F.col("L2").cast("double"))
+            .filter(F.col("L2").isNotNull())
```
project-defiant, line 353: Probably also worth checking if the L2 (assuming that is the ldscore for variant j, is >= 1.0).

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            .dropDuplicates([chrom_col, pos_col, "ref", "alt"])
+        )
+
+    def _merge_with_ld_scores(
+        self,
```
project-defiant, line 358: the Step classes are not objects per se, you should not rely on the inner state of `self....`, instead pass the arguments explicitly.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        merged_ld = prepared_sumstats.join(
+            ld_df,
+            on=[chrom_col, pos_col, "ref", "alt"],
+            how="inner",
+        )
+
+        dedup_df = merged_ld.dropDuplicates([chrom_col, pos_col, "ref", "alt"])
+
+        return dedup_df, m_ldsc
```
project-defiant, line 390: m should be a column based feature, or count table per studyId.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        dedup_df, m_ldsc = self._merge_with_ld_scores(
+            prepared_sumstats=prepared_sumstats,
+            ld_df=ld_df,
+        )
+
+        n_rows = dedup_df.count()
+
+        if n_rows == 0:
+            self._write_result_row(
```
project-defiant, line 120: Writing result_row should be done once, this would simplify a lot, first collect all info, then write, not write on every exit issue.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+                m_ldsc=m_ldsc,
+                n_snps_used=n_rows,
+            )
+            return
+
+        beta, se, n_array, ld = self._collect_ldsc_arrays(
+            dedup_df=dedup_df,
+            study_id=study_id,
+        )
+        result = self._run_ldsc(
```
project-defiant, line 158: Probably this is a target for pandas_udf function, avoid casting from pyspark -> list[float] -> numpy, this will be a lot of overhead on this simple task.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            beta=beta,
+            se=se,
+            n_array=n_array,
+            ld=ld,
+            m_ldsc=m_ldsc,
+        )
+
+        self.results = result
+
+        self._write_result_row(
```
project-defiant, line 168: It would be very usefull to see some logging in the step.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            "slope", "slope_se", "mean_chisq", "lambda_gc",
+            "coef", "coef_se".
+    """
+    beta = np.asarray(beta, dtype=float)
+    se = np.asarray(se, dtype=float)
+    N = np.asarray(N, dtype=float)
+    ld = np.asarray(ld, dtype=float)
+    w_ld = np.asarray(w_ld, dtype=float)
+
+    if not (beta.ndim == se.ndim == N.ndim == ld.ndim == w_ld.ndim == 1):
```
project-defiant, line 1445: this would not be needed, given you convert the function to pandas_udf.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        raise ValueError("All inputs beta, se, N, ld, w_ld must be 1D arrays.")
+
+    Z = beta / se
```
project-defiant, line 1448: pass z-score instead of beta and se to the function parameters.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+                f"LD score file is missing required columns: {sorted(missing_ld_cols)}"
+            )
+
+        return (
+            ld_df.select(chrom_col, pos_col, "ref", "alt", "L2")
+            .withColumn(pos_col, F.col(pos_col).cast("int"))
+            .withColumn(chrom_col, F.col(chrom_col).cast("string"))
+            .withColumn("L2", F.col("L2").cast("double"))
+            .filter(F.col("L2").isNotNull())
+            .dropDuplicates([chrom_col, pos_col, "ref", "alt"])
```
project-defiant, line 354: Probably you could prepare the ldscores dataset once (per ancestry) and then reuse it directly with `LDScoreDataset.from_parquet(...)`

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        return study_ids[0]
+
+    def _read_study_index(self) -> DataFrame:
+        """Read the study index parquet.
+
+        Returns:
+            DataFrame: Study index dataframe with metadata required for validation.
+        """
+        return self.session.spark.read.parquet(self.study_index_input_path).select(
```
project-defiant, line 227: You probably should extract the major population here and ensure the studyIndex is valid (ex. nCases + nControls = nSamples), etc.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            dict[str, Any]: Validation result with status, reasons, flags, and ancestry.
+        """
+        row = (
+            study_index_df.filter(F.col("studyId") == study_id)
+            .select("ldPopulationStructure", "analysisFlags", "nSamples")
+            .first()
+        )
+
+        if row is None:
+            raise ValueError(f"studyId {study_id} not found in study index")
```
project-defiant, line 566: Check should be generic, not include study_id, as it will eventually be a data that contains the studyId in the table. Easier to filter over.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+        skip_reasons: list[str] = []
+
+        if not self._is_allowed_by_analysis_flags(row["analysisFlags"]):
+            skip_reasons.append("Invalid study design")
+
+        n_samples = row["nSamples"]
+        if n_samples is None:
+            skip_reasons.append("Sample size missing")
+        elif float(n_samples) < float(min_samples):
+            skip_reasons.append(f"Sample size too small: {n_samples} < {min_samples}")
```
project-defiant, line 578: The same as above.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+
+        return (
+            sumstats_df.join(n_df, on=study_col, how="left")
+            .withColumn(
+                n_col,
+                F.when(F.col(n_col).isNull(), fallback_n_expr).otherwise(
+                    F.col(n_col).cast("double")
+                ),
+            )
+            .withColumn("variant_parts", F.split(F.col("variantId"), "_"))
```
project-defiant, line 298: So the overall condition of `n` is scattered around the logic of LDSC step.

1. if nSamples is Null => run_status = `skipped`
2. if nCases > 0 & nControls > 0 => { n = (4.0 * `nCases` * `nControls`) / (`nCases` + `nControls`)} otherwise `nSamples`

There are a few questions which may be corner cases to explore:
```
+--------------------------------------------------+------+
|nCheck                                            |count |
+--------------------------------------------------+------+
|nSamples > 0, nCases zero, nControls zero         |181036|
|nSamples = nCases + nControls                     |31927 |
|nSamples > 0, nCases > 0, nControls zero          |713   |
|nSamples null, nCases not null, nControls not null|602   |
|nSamples != not equal nCases + nControls          |401   |
|nSamples > 0, nCases zero, nControls > 0          |34    |
|All null                                          |3     |
+--------------------------------------------------+------+
```
See issue https://github.com/opentargets/issues/issues/3879.

Now, we should probably explore all corner cases and ensure we flag these studies for Heritability when `nCases + nControls != nSamples` or one of `nCases` or `nControls` is zero, but other ones are not.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            .drop("variant_parts", "nSamples", "nCases", "nControls")
+            .filter(F.col(beta_col).isNotNull())
+            .filter(F.col(se_col).isNotNull())
+            .filter(F.col(n_col).isNotNull())
+            .filter(F.col(se_col) > 0)
+            .filter(F.col(chrom_col).isNotNull())
+            .filter(F.col(pos_col).isNotNull())
+            .filter(F.col("ref").isNotNull())
+            .filter(F.col("alt").isNotNull())
+            .dropDuplicates([study_col, chrom_col, pos_col, "ref", "alt"])
```
project-defiant, line 310: No need to drag full dataframe, use select post filtering, so the exact structure of the result is well understood.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            return
+
+        if n_rows > self.max_rows_for_collection:
+            self._write_result_row(
+                study_id=study_id,
+                heritability_output_path=self.heritability_output_path,
+                run_status="skipped",
+                skip_reasons=[
+                    (
+                        "Too many joined SNPs for collection: "
```
project-defiant, line 142: Again, do not make the skip_reason specific. (easier filtering of results)

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            ValueError: If multiple study identifiers are found or no valid rows remain.
+        """
+        beta_col = "beta"
+        se_col = "standardError"
+        n_col = "sampleSize"
+        study_col = "studyId"
+
+        rows = dedup_df.select(beta_col, se_col, n_col, "L2", study_col).collect()
+
+        study_ids_rows = {row[study_col] for row in rows}
```
project-defiant, line 417: this should be redundant check in this implementation.

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            se_list.append(row[se_col])
+            n_list.append(row[n_col])
+            ld_list.append(row["L2"])
+
+        if not beta_list:
+            raise ValueError(f"No valid rows after filtering for studyId {study_id}")
+
+        beta = np.array(beta_list, dtype=float)
+        se = np.array(se_list, dtype=float)
+        n_array = np.array(n_list, dtype=float)
```
project-defiant, line 439: Am I correct assuming, that `n` is a vector of duplicated values, given we have a single studyId ?

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+    def _run(self) -> None:
+        """Execute the LDSC heritability estimation pipeline."""
+        sumstats_df = self._read_sumstats()
+        study_id = self._extract_single_study_id(sumstats_df)
+
+        study_index_df = self._read_study_index()
+        validation = self._validate_study(
+            study_index_df=study_index_df,
+            study_id=study_id,
+            min_samples=self.min_samples,
```
project-defiant, line 86: This may be important, the check you perform to figure out if nSamples >= min_samples is done before you actually calculate the `n_effective` so in case of large imbalance, we will get very low sampleSize (possibly below the expected cutoff.)

src/gentropy/ldsc.py
```diff
@@ -0,0 +1,725 @@
...
+            ld (np.ndarray): LD scores for the matched variants.
+            m_ldsc (float): Number of SNPs in the LD-score universe.
+
+        Returns:
+            dict[str, Any]: LDSC result dictionary returned by
+                ``run_ldsc_h2_from_arrays``.
+        """
+        w_raw = 1.0 / np.maximum(ld, 1.0)
+        w_ld = w_raw / np.mean(w_raw)
+
```
project-defiant, line 467: Please describe how the weight is calculated and why it is required. Possibly this should be included in the method itself.

I assume it is due to the heteroskedasticity problem with weighted regression and SNPs with high LD (hith LDScore) will get lower weights in the regression.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+    chisq = Z**2
+
+    n = chisq.shape[0]
+    y = chisq.reshape((n, 1))
+    x = ld.reshape((n, 1))
+    w = w_ld.reshape((n, 1))
+    N_mat = N.reshape((n, 1))
+    M_mat = np.array([[float(M_ldsc_scalar)]])
```
project-defiant, line 1456: not reshaping to the (n,1) ?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+    n_annot = x.shape[1]
```
project-defiant, line 1458: why recompute? the `x.shape[1] == 1`, right?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+    old_weights = False
+    if n_annot == 1:
+        step_cutoff = twostep if intercept is None else None
+    else:
```
project-defiant, line 1462: **NOTE:
The current code does not allow you to build the partitioned ldsc config, as n_annot is forced to be (n,1) (shape[1] == 1) always, so this code is dead and we will always hit the case when `n_annot == 1`.**

Below I asked to rewrite this logic into pseudo-code, it would be very useful to explain all of the hsq configuration modes that we can end up with. 

### Determine how to configure LDSC regression based on annotation count
annotation_count = number of columns in design_matrix_x

### Default to the modern IRWLS-based weighting path
use_legacy_weights = false

if annotation_count equals 1 then
    # Two-step LDSC is only applicable when the intercept is free
    if intercept is not fixed then
        step_one_chisq_cutoff = twostep_threshold
    else
        step_one_chisq_cutoff = null
    end if
else
    # Multi-annotation regression falls back to the original LDSC weights
    use_legacy_weights = true
    step_one_chisq_cutoff = null
end if

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+        Args:
+            chisq (np.ndarray): Chi-square statistics of shape (n_snp, 1).
+
+        Returns:
+            tuple[float, float]: Mean chi-square and lambda_GC.
+        """
+        mean_chisq = float(np.mean(chisq))
+        lambda_gc = float(np.median(np.asarray(chisq)) / 0.4549)
+        return mean_chisq, lambda_gc
```
project-defiant, line 1303: this helper introduces to much noise, ex. the input chisq is already the ndarray, no need to cast it back to `np.asarray`. It would be better to store the function to compute lambda_gc in stats.py module.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            M (np.ndarray): Row vector of annotation sizes of shape
+                (1, n_annot).
+            n_blocks (int): Number of jackknife blocks.
+            intercept (float | None): Fixed intercept value, if any.
+
+        Raises:
+            TypeError: If any argument is not array-like with 2 dimensions.
+            ValueError: If shapes are incompatible.
+        """
+        for i in [y, x, w, M, N]:
```
project-defiant, line 642: You are rechecking the size of the nd arrays again, as in the `run_ldsc_h2_from_arrays`.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            try:
+                if len(i.shape) != 2:
+                    raise TypeError("Arguments must be 2D arrays.")
+            except AttributeError as err:
```
project-defiant, line 646: The type annotations should force the np.ndarray, so the exception should be never raised.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+                raise TypeError("Arguments must be arrays.") from err
+
+        n_snp, self.n_annot = x.shape
```
project-defiant, line 649: Again checking the sizes, I would argue this should be done in a single place (in the LDSC method validation.)

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        if any(i.shape != (n_snp, 1) for i in (y, w, N)):
+            raise ValueError("N, weights and response must have shape (n_snp, 1).")
+        if M.shape != (1, self.n_annot):
+            raise ValueError("M must have shape (1, n_annot).")
+
+        self.constrain_intercept: bool = intercept is not None
```
project-defiant, line 655: You should not set the attributes in private methods, do the staticmethod that validate the shape of the inputs and return it back in the constructor. 

Currently any method that would call the attribute assigned in the `_validate_*` private method will be forced to check if the attribute exists on the class, else this will raise typing issues.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+                Tuple containing:
+                    x (np.ndarray): Rescaled design matrix.
+                    x_tot (np.ndarray): Total LD scores per SNP.
+                    yp (np.ndarray): Response with intercept handled.
+                    Nbar (float): Mean sample size.
+                    M_tot (float): Total number of SNPs used in LD scores.
+                    initial_w (np.ndarray): Initial weights for IRWLS.
+        """
+        n_snp = x.shape[0]
+        M_tot: float = float(np.sum(M))
```
project-defiant, line 692: uh, now I understand why you do not reshape the `m` to (n,1) size ...
i would argue, this is completely redundant operation, as M is array of size one that you forced from float upsteam in callstack.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        x: np.ndarray,
+        N: np.ndarray,
+        M: float,
+        intercept: float | None = None,
+    ) -> float:
+        """Initial aggregate estimate used to set starting weights.
+
+        This matches the approximate LDSC formula:
+
+            E[y] ≈ intercept + (N / M) * x * h2
```
project-defiant, line 1014: E[χ²] ≈ intercept + (N / M) * ℓ​ * h2.

you can use latex in the docstrings. Please do.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            N (np.ndarray): Sample sizes of shape (n_snp, 1).
+            M (float): Total number of SNPs used in LD-score estimation.
+            intercept (float | None): Intercept value. If None, uses
+                `cls.__null_intercept__`.
+
+        Returns:
+            float: Initial aggregate heritability or covariance estimate.
+        """
+        if intercept is None:
+            intercept = cls.__null_intercept__  # type: ignore[attr-defined]
```
project-defiant, line 1028: this will break in the LDSC class, it can be only called in Hsq, as the __null_intercept__ is defined only there, also please do not use double underscores for class variables, use uppercase and make it constant.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        num = M * (np.mean(y) - intercept)
```
project-defiant, line 1029: is this logic for setting up the initial values for h2? Genuine question from laik about GLS -> can we specify random values for h2 instead?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+                    Nbar (float): Mean sample size.
+                    M_tot (float): Total number of SNPs used in LD scores.
+                    initial_w (np.ndarray): Initial weights for IRWLS.
+        """
+        n_snp = x.shape[0]
+        M_tot: float = float(np.sum(M))
+        x_tot = np.sum(x, axis=1).reshape((n_snp, 1))
+
+        tot_agg: float = self.aggregate(y, x_tot, N, M_tot, intercept)
+        initial_w = self._update_weights(x_tot, w, N, M_tot, tot_agg, intercept)  # type: ignore[attr-defined]
```
project-defiant, line 696: This line is broken, `_update_weights` exists in Hsq class, not on `LD_Score_Regression`, hence we can only call it without an error on Hsq class.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        Returns:
+            np.ndarray: Regression weights of shape (n_snp, 1).
+        """
+        M = float(M)
+        if intercept is None:
+            intercept = 1.0
+
+        hsq = max(hsq, 0.0)
+        hsq = min(hsq, 1.0)
+        ld = np.fmax(ld, 1.0)
```
project-defiant, line 1392: again redundant, the ld is already clipped (1, inf)

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            intercept (float | None): Intercept used in the heteroskedastic
+                variance model. Defaults to 1 if None.
+            ii (np.ndarray | None): Unused. Present for API compatibility.
+
+        Returns:
+            np.ndarray: Regression weights of shape (n_snp, 1).
+        """
+        M = float(M)
+        if intercept is None:
+            intercept = 1.0
```
project-defiant, line 1388: why not use __null_* variable ?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+    @classmethod
+    def weights(
+        cls,
+        ld: np.ndarray,
+        w_ld: np.ndarray,
+        N: np.ndarray,
+        M: float,
+        hsq: float,
+        intercept: float | None = None,
```
project-defiant, line 1364: You are dragging the intercept as None, but it can not be none, (Hsq.weights is called in _update_weights method which already implements the fallback to __null_intercept__ if intercept is None)

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+        Returns:
+            np.ndarray: Regression weights of shape (n_snp, 1).
+        """
+        M = float(M)
+        if intercept is None:
+            intercept = 1.0
+
+        hsq = max(hsq, 0.0)
+        hsq = min(hsq, 1.0)
```
project-defiant, line 1391: hsq initially needs to be cropped to <0.0, 1.0>

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        ld = np.fmax(ld, 1.0)
+        w_ld = np.fmax(w_ld, 1.0)
+        c = hsq * N / M
+        het_w = 1.0 / (2 * np.square(intercept + np.multiply(c, ld)))
+        oc_w = 1.0 / w_ld
+        w = np.multiply(het_w, oc_w)
+        return w
```
project-defiant, line 1398: Add the formula to compute the updated weight

wj​=1/2(intercept+MNh^2​⋅ℓj​)^2

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        n_snp = x.shape[0]
+        M_tot: float = float(np.sum(M))
+        x_tot = np.sum(x, axis=1).reshape((n_snp, 1))
+
+        tot_agg: float = self.aggregate(y, x_tot, N, M_tot, intercept)
+        initial_w = self._update_weights(x_tot, w, N, M_tot, tot_agg, intercept)  # type: ignore[attr-defined]
+
+        Nbar: float = float(np.mean(N))  # keep condition number low
+        x = np.multiply(N, x) / Nbar
+
```
project-defiant, line 700: reason for scaling the ldscores ?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        """
+        if np.any(w <= 0):
+            raise ValueError("Weights must be > 0")
+        n, p = x.shape
+        if w.shape != (n, 1):
+            raise ValueError(f"w has shape {w.shape}. w must have shape (n, 1).")
+        w_n = w / float(np.sum(w))
+        return np.multiply(x, w_n)
+
+def append_intercept(x: np.ndarray) -> np.ndarray:
```
project-defiant, line 480: Call to this method results in (nrow, 2) matrix
```
[[1],[2],[3]] -> 
[
    [1, 1],
    [1, 2],
    [1, 3],
]
```
(n,1) -> (n,2)

You should add a doctest to ensure this is happeing.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+        if not self.constrain_intercept:
+            x = append_intercept(x)
+            x_tot = append_intercept(x_tot)
+            yp = y
+        else:
+            yp = y - intercept
+            self.intercept_se = "NA"
+
+        return x, x_tot, yp, Nbar, M_tot, initial_w
```
project-defiant, line 709: Please document it, 
1) when intercept is not constraint => estimate intercept (add to design matrix)
2) when intercept is constrait => condition on known fixed effect and remove it from the response before estimating slopes.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            ValueError: If two-step LDSC is requested with a constrained
+                intercept or with partitioned LD Score (n_annot > 1).
+        """
+        if step1_ii is None:
+            return
+
+        if self.constrain_intercept:
+            raise ValueError("twostep is not compatible with constrained intercept.")
+        if self.n_annot > 1:
+            raise ValueError("twostep not compatible with partitioned LD Score yet.")
```
project-defiant, line 731: This block of code should be at the top of the callstack, so you do not need to do all of the computations, before that. Note it is deeply nested and clearly explains that we can not run `two_step` when intercept if fixed or we want to run partitioned hsq.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            slow (bool): Kept for compatibility with older interfaces.
+            step1_ii (np.ndarray | None): Boolean mask for two-step LDSC.
+            old_weights (bool): Use original LDSC weighting scheme if True.
+            initial_w (np.ndarray): Initial weights for IRWLS.
+
+        Returns:
+            Any: A jackknife object with attributes `est`, `jknife_cov`,
+                `jknife_se`, and `delete_values`.
+        """
+        if step1_ii is not None:
```
project-defiant, line 770: You should totally describe the modes enum:

```py
from enum import StrEnum

HsqRunModes(StrEnum):
    S_LDSC = "Partitioned LDSC"
    FIXED_INTERCEPT_LDSC = "Fixed..."
    TWO_STEP_LDSC = "Two step..."
```

Later each mode should describe it's own input conditions (like `FIXED_INTERCEPT_LDSC` requires intercept to be specified apriori, `S_LDSC` requires (n,s) design matrix where n is the number of variants, s is the partitioning col.)

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+                N=N,
+                M_tot=M_tot,
+                Nbar=Nbar,
+                n_blocks=n_blocks,
+                slow=slow,
+                step1_ii=step1_ii,
+                initial_w=initial_w,
+            )
+
+        if old_weights:
```
project-defiant, line 785: The name is not super explainable, why the weights are old?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            step1_ii (np.ndarray): Boolean mask for first-step SNP subset.
+            initial_w (np.ndarray): Initial weights for IRWLS.
+
+        Returns:
+            Any: Jackknife-like namedtuple with fields:
+                est, jknife_se, jknife_est, jknife_var, jknife_cov, delete_values.
+        """
+        n_snp = x.shape[0]
+        n1 = int(np.sum(step1_ii))
+        self.twostep_filtered = int(n_snp - n1)
```
project-defiant, line 844: the `twostep_filtered` attr is defined somewhere else..., hard to track dependency between methods.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+        x1 = x[np.squeeze(step1_ii), :]
```
project-defiant, line 846: NIT: 
```py
mask = np.squeeze(step1_ii)
yp1, w1, N1, initial_w1 = [a[mask, :] for a in (yp, w, N, initial_w)]
```

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            N (np.ndarray): Sample sizes of shape (n_snp, 1).
+            M (float): Total number of SNPs used for LD scores.
+            Nbar (float): Mean sample size.
+            intercept (float | None): Intercept used when building weights.
+            ii (np.ndarray | None): Optional mask for subsetting N and LD.
+
+        Returns:
+            np.ndarray: New weights of shape (n_snp, 1).
+        """
+        raise NotImplementedError
```
project-defiant, line 1062: Good! this has to be defined in parent as not implemented, if you ever want to use this in child.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        n1 = int(np.sum(step1_ii))
+        self.twostep_filtered = int(n_snp - n1)
+
+        x1 = x[np.squeeze(step1_ii), :]
+        yp1, w1, N1, initial_w1 = [
+            a[step1_ii].reshape((n1, 1)) for a in (yp, w, N, initial_w)
+        ]
+
+        update_func1 = lambda a: self._update_func(
+            a,
```
project-defiant, line 852: this is very confusing, the `a` passed in the function call is actually passed to `x` in function def.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            x1,
+            w1,
+            N1,
+            M_tot,
+            Nbar,
+            ii=step1_ii,
```
project-defiant, line 858: Do we need to pass the mask downstream, if it was already used to filter the data?

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        )
+        step1_jknife = IRWLS(
+            x1,
+            yp1,
```
project-defiant, line 862: Just for the note: 
x1 is (n1, 1)

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+        Args:
+            x (np.ndarray): Design matrix of shape (n, p).
+            y (np.ndarray): Response vector of shape (n, 1).
+            update_func (Callable[[tuple[Any, ...]], np.ndarray]): Callable taking the output of
+                `np.linalg.lstsq` and returning new weights of shape (n, 1).
+            n_blocks (int): Number of jackknife blocks.
+            w (np.ndarray | None): Initial weights of shape (n, 1). If None,
+                all weights are set to 1.
+            slow (bool): Kept for API compatibility. Ignored (only fast
```
project-defiant, line 342: What API ? You did not define any contract, only abstract class for LD_Score_Regression.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+
+from collections import namedtuple
+from collections.abc import Callable, Sequence
+from typing import Any
+
+import numpy as np
+
+np.seterr(divide="raise", invalid="raise")
+
+def _check_shape(x: np.ndarray, y: np.ndarray) -> tuple[int, int]:
```
project-defiant, line 21: Am I correct assuming x and y be the same size ? this function seem to be called when intercept is free.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            jknife_var (np.ndarray): Jackknife variances of shape (1, p).
+            jknife_cov (np.ndarray): Jackknife covariance matrix of shape
+                (p, p).
+            delete_values (np.ndarray): Delete values of shape (n_blocks, p).
+            separators (np.ndarray): Block boundaries used by the jackknife.
+
+        Raises:
+            ValueError: If shapes of `x`, `y` or `w` are incompatible.
+        """
+        n, p = _check_shape(x, y)
```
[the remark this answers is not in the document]
project-defiant: Agreed, if it will be reused for fixed slope or partitioned hsq, then we shall see `p` defined, otherwise it will always be eq to 1.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        if w is None:
```
project-defiant, line 361: When weights can be None? Possibly this is an artifact that would allow to reuse this kind of regression algo for other applications, but!

1) is should not live in the ldsc.py in that case
2) update_func callable typing should have better design to understand it's inputs/outputs without checking example implementations in ldsc.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+            w = np.ones_like(y)
+        if w.shape != (n, 1):
```
project-defiant, line 363: You should check the shape of w in the `_check_shape`, since you are already checking it for x and y

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        Raises:
+            ValueError: If `y` or `w` shapes are incompatible with `x`, or if
+                `update_func` returns weights with mismatched shape.
+        """
+        n, p = x.shape
+        if y.shape != (n, 1):
+            raise ValueError(f"y has shape {y.shape}. y must have shape ({n}, 1).")
+        if w.shape != (n, 1):
+            raise ValueError(f"w has shape {w.shape}. w must have shape ({n}, 1).")
+
```
project-defiant, line 417: You are repeating checks from the init of this class.

src/gentropy/method/ldsc_h2.py
```diff
@@ -0,0 +1,1491 @@
...
+        Args:
+            x (np.ndarray): Design matrix of shape (n, p).
+            y (np.ndarray): Response vector of shape (n, 1).
+            n_blocks (int | None): Number of jackknife blocks. If provided,
+                evenly spaced block boundaries are constructed.
+            separators (Sequence[int] | None): Monotonically increasing
+                integer indices (including 0 and n) that define block
+                boundaries.
+
```
project-defiant, line 112: Would be very nice to see the example in docstring with separators.

project-defiant (changes requested):
This PR implements the Heritability step that calculates the LDSC regression on gnomAD LDScores and     
Gentropy SummaryStatistics.

I was not able to finish the PR review fully (pvalue calculation) as I did not understnad most of the code and operations you do. There is clearly a lot of room to improvement of the code understanding, fortunately (by my means) I could not find a thing that would make the business logic very wrong. 

This code does not look healthy, there is a lot of redundancy, it does not allow to run Hsq on multiple summary statistics at once, the logic of Hsq and LD_Score_Regression classes is highly convoluted as well as the code running for Jackknife and it's derivative class. The method assumes it could potentially run in 3 modes: 1) fixed intercept, 2) free intercept + twostep 3) partitioned hsq. Unfortunately the amount of noise in the code needs to be reduced before we can merge and release the hsq step. 


Overall I feel, we need to reduce 3/4 of the redundant codebase and properly define the method. (**We need to do that before Tobi starts working on partitioned heritability and before the genetics correlation**). We will not be able to to that before the release, so we need to discuss the next actions. Glad to see the change to StudyIndex is there though, that is one bit I was afraid we would miss. 

Good job overall, from the code and the maths derivations, I see this is very complicated, but you managed to make it work.

project-defiant:
@xyg123 last comment: 
Why we did not utilize existing ldsc implementation ? There are at least 3 different implementations
* https://github.com/abdenlab/ldsc-python3
* https://github.com/CBIIT/ldsc
* https://github.com/bulik/LDSC

project-defiant (approved):
Approving, but please have a look at the comments. We are not going to refactor the logic for the actual implementation of LDScore. For future we should:
1) allow to calculate hsq for multiple studies.
2) align the hsq implementation to the gentropy standard.

PR #1209, opened 2026-03-11
feat(dependencies): remove google secret manager

### LINKED ISSUE #3645: Hardcoded project id in Gentropy L2G
The `run_predict` method in L2G is pulling the GCP Secret for wandb with a hardcoded project id.

We need to take that out so, otherwise anybody who wants to run Gentropy requires permissions for that project.

### LINKED ISSUE #3835: Remove Google Secret Manager dependency from Gentropy
As a **Gentropy dev** I want **to remove Google Secret Manager from the list of dependencies** because **handling of environment variables like API tokens should be handled at the orchestration level and not by the library**.

## Background
The `access_gcp_secret` function is used to:
1. Fetch API key to log L2G evaluation to W&B
2. Fetch API key to pull/push L2G model from/to the Hub

The L2G steps shouldn't rely on access to GCP to run

## Tasks

- [ ] Refactor L2G step from accessing secrets stored in GCP
- [ ] Remove `google-cloud-secret-manager` as a dependency
- [ ] To provide environment variables at the orchestration level

project-defiant:
## ✨ Context

This PR removes google cloud secret manager from gentropy dependencies.

https://github.com/opentargets/issues/issues/3835https://github.com/opentargets/issues/issues/3835

- [x] Removal of secret manager dependency
- [x] Inclusion of pydantic to build model classes
- [x] Inclusion of models for W&B and HF credentials
- [x] Refactoring of gentropy L2G steps to use the paths to the credential files instead of tokens
- [x] Fallback for using the env variables `HF_TOKEN` and `WANDB_API_KEY`  if access to the services is requrested, but no credential files are provided
- [x] Example credential files
- [x] Tests   

## 🚦 Before submitting


src/gentropy/l2g.py
```diff
@@ -183,27 +191,35 @@ def __init__(
...
+        self.wandb_run_name = wandb_run_name
+        self.wandb_credentials_path = wandb_credentials_path
+
+        # Predict io
+        self.download_from_hub = download_from_hub
+        self.model_path = model_path or "opentargets/locus2gene"
+        self.predictions_path = predictions_path
+
+        # Predict parameters
+        self.l2g_threshold = l2g_threshold or 0.05
```
ireneisdoomed: If 0.05 is going to be the default, I'd move this to the function definition. It is clearer and better documented

src/gentropy/l2g.py
```diff
@@ -183,27 +191,35 @@ def __init__(
...
-            else model_path
-        )
         self.hf_model_version = hf_model_version
+        self.hf_credentials_path = hf_credentials_path
+        self.wandb_run_name = wandb_run_name
+        self.wandb_credentials_path = wandb_credentials_path
+
+        # Predict io
+        self.download_from_hub = download_from_hub
+        self.model_path = model_path or "opentargets/locus2gene"
```
ireneisdoomed: Same comment as `l2g_threshold`? In any case, this needs changing to `opentargets/locus_to_gene`

src/gentropy/l2g.py
```diff
@@ -214,7 +230,14 @@ def __init__(
             _df=session.load_data(feature_matrix_path, "parquet"),
         )
 
+
         if run_mode == "predict":
+            if download_from_hub:
+                if not self.hf_hub_repo_id or not self.hf_model_version:
```
ireneisdoomed: I wouldn't require `hf_model_version`. I think the former approach of "unless a specific commit is specified, let's use the latest" is better

Nevertheless, i introduced this when the L2G model were expected to live in the same repo. Now we have one repo for each model we release, so this is less relevant. I think the usecase is minimal and it is not worth the complexity

tests/gentropy/no_spark/test_external_credentials.py
ireneisdoomed, line 1: Very nice testing suite <3

src/gentropy/external/gcs/__init__.py
```diff
@@ -1,22 +1,4 @@
-"""Common functions for Google cloud storage (GCS) operations."""
-
-
-def access_gcp_secret(secret_id: str, project_id: str) -> str:
```
ireneisdoomed, line 4: I agree it's not worth keeping this util for devs to create the env files

src/gentropy/external/wandb/__init__.py
ireneisdoomed, line 1: This module is really similar to the `HuggingFaceHubCredentials` class. Do you think we could merge them into an `SecretsHandler` class or something like that? It'd require less testing as well. Same comments goes for the `Missing...ApiKeyError` classes

src/gentropy/external/hf_hub/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+            import os
+            token = os.getenv("HF_TOKEN")
+            if not token:
+                raise MissingHFTokenError("HF_TOKEN environment variable is not set.")
+            return cls(HF_TOKEN=SecretStr(token))
+
+        return cls.from_json(path)
+
+
+class HuggingFaceModelRepoHandle(BaseModel):
```
ireneisdoomed, line 35: Nice

src/gentropy/l2g.py
```diff
@@ -364,8 +387,15 @@ def run_train(self) -> None:
...
-            wandb_key = access_gcp_secret("wandb-key", "open-targets-genetics-dev")
-            wandb_login(key=wandb_key)
+            wandb_credentials = WandbCredentials.read(self.wandb_credentials_path)
+            wandb_login(key=wandb_credentials.WANDB_API_KEY.get_secret_value())
+
+        # Initialize access to Hugging Face Hub
+        hf_token = None
+        if self.hf_hub_repo_id and self.hf_model_commit_message:
+            # Fails when the HF_TOKEN env is not set
+            hf_hub_credentials = HuggingFaceHubCredentials.read(self.hf_credentials_path)
```
ireneisdoomed: Should we check credentials earlier in the step?

src/gentropy/l2g.py
```diff
@@ -175,6 +181,8 @@ def __init__(
             hf_model_commit_message (str | None): Commit message when we upload the model to the Hugging Face Hub
             hf_model_version (str | None): Tag, branch, or commit hash to download the model from the Hub. If None, the latest commit is downloaded.
             explain_predictions (bool | None): Whether to extract SHAP importances for the L2G predictions. This is computationally expensive.
+            hf_credentials_path (str | None): Optional path to the Hugging Face Hub credentials JSON file. If not provided, the HF_TOKEN environment variable will be used.
```
ireneisdoomed, line 212: Should we document where/how to fetch them from?

ireneisdoomed (changes requested):
Thank you so much, this is a great contribution. We remove the Google Secrets dependency and now the L2G step expects reading these credentials from a local file.

I am asking for request changes because I have comments about some default values in the L2G specification, please have a look.
Other than that, my main comment is whether it's doable to make a generic class for handling credentials.

Nice use of pydantic as well, I didn't know about `SecretStr`

src/gentropy/l2g.py
project-defiant: you are right! thanks for pointing this out

src/gentropy/l2g.py
project-defiant: Ok, I assume the `hf_model_version` is going to be `latest` by default. Agree with less complexity though.

src/gentropy/l2g.py
```diff
@@ -324,6 +347,11 @@ def run_predict(self) -> None:
         Raises:
             ValueError: If predictions_path is not provided for prediction mode
         """
+        hf_token = None
+        if self.download_from_hub:
+            hf_hub_credentials = HuggingFaceHubCredentials.read(self.hf_credentials_path)
+            hf_token = hf_hub_credentials.HF_TOKEN
```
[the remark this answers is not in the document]
project-defiant: This is by design. I do not want to make the the HF_TOKEN optional when download_from_hub is set to True.

src/gentropy/l2g.py
project-defiant: It's a tricky question. By design how the step is now, it's hard to split the logic, but in general I would agree with you that we should first check the environment, later start reading datasets. We can make a follow up change that deals with splitting the L2G step into train & test and implement it there.

src/gentropy/external/wandb/__init__.py
project-defiant: Yes, I think it would be beneficial.

project-defiant:
@copilot resolve the merge conflicts in this pull request

project-defiant:
@ireneisdoomed please see the changes.

src/gentropy/l2g.py
```diff
@@ -168,17 +172,19 @@ def __init__(
...
         predictions_path: str | None = None,
-        l2g_threshold: float | None = None,
-        hf_hub_repo_id: str | None = None,
-        hf_model_commit_message: str | None = "chore: update model",
-        hf_model_version: str | None = None,
-        explain_predictions: bool | None = None,
+        l2g_threshold: float = 0.05,
+        hf_hub_repo_id: str = "locus_to_gene",
+        hf_model_commit_message: str = "chore: update model",
+        hf_model_version: str = "latest",
```
ireneisdoomed: See the comment I did on `config.py` for this parameter

src/gentropy/config.py
```diff
@@ -352,13 +352,15 @@ class LocusToGeneConfig(StepConfig):
         }
     )
     wandb_run_name: str | None = None
-    hf_hub_repo_id: str | None = None
-    hf_model_commit_message: str | None = None
-    hf_model_version: str | None = None
+    hf_hub_repo_id: str = "locus_to_gene"
+    hf_model_commit_message: str = "chore: update model"
+    hf_model_version: str = "latest"
```
ireneisdoomed: I am not sure this would work. Basically, if you pass a commit a string, it has to match the commit or tag IDs available for that repo. But I might be on the wrong, have you tried it?

You get the latest version if `hf_model_version` is set to None. My suggestion was to leave it out of the conditional block (which you have done)

src/gentropy/external/wandb/__init__.py
ireneisdoomed: Just to note, this hasn't been implemented and would be tackled in another PR

src/gentropy/l2g.py
ireneisdoomed: Cool! I have a branch that splits the train and test logic. I can update it once we merge your changes

ireneisdoomed (commented):
Thank you, all good. Just have a look at the HF model versioning comment

src/gentropy/config.py
project-defiant: Yes, you are probably correct.

src/gentropy/external/wandb/__init__.py
project-defiant: @ireneisdoomed I did the base class for secret credentials, but as you may have seen the logic prevents us from merging the two classes into a single class, that is due to the differences in `static` keys. I wanted to make this distinction clear, as the json key-value will also reflect how the env variable should look like, but if you think it's too verbose, we can collapse both models in the future. 

Thanks!

ireneisdoomed (approved):

PR #1224, opened 2026-05-13
feat(l2g): add dark matter loci filter for training set

addramir:
## Background

Analysis of the 26.03 L2G training set revealed a structural problem: gold-standard positives with no functional genomics signal — termed **dark matter** — cannot be learned from by the model. The model sees no QTL colocalisation, no E2G signal, no coding variant, only distance features that are present for every gene in the locus. Retaining them adds noise without providing any learnable pattern.

## Definition of dark matter

A positive is dark matter when **both** of the following hold simultaneously:

**1. No functional genomics signal** — all of these are zero or absent (NULL treated as zero):
- `eQtlColocClppMaximum`, `eQtlColocH4Maximum`
- `pQtlColocClppMaximum`, `pQtlColocH4Maximum`
- `sQtlColocClppMaximum`, `sQtlColocH4Maximum`
- `transPQtlColocH4Maximum`
- `e2gMean`
- `vepMaximum` < 0.6 (below protein-altering variant threshold, SO:0001818)

NULL values are treated as zero because colocalisation and VEP features are produced via inner joins — a missing row means no evidence, not unknown evidence.

**2. Not the nearest gene** — all neighbourhood distance features < 1.0:
- `distanceSentinelFootprintNeighbourhood`
- `distanceFootprintMeanNeighbourhood`
- `distanceTssMeanNeighbourhood`
- `distanceSentinelTssNeighbourhood`

A value of 1.0 in any neighbourhood distance feature means the gene is the nearest in the locus window. Nearest-gene positives are kept even without functional signal, because proximity-based learning is still valid for them.

## Locus removal logic

A `studyLocusId` is removed only when **all** of its positives are dark matter. If a locus has two positives and one has signal, the whole locus is kept. When a locus is removed, all rows (positives and negatives) are dropped to preserve class balance within retained loci.

### `L2GFeatureMatrix.filter_dark_matter_loci()` (new method)
- Builds the no-signal condition from `_DARK_MATTER_SIGNAL_FEATURES` (8 QTL/E2G features) + VEP threshold check; all comparisons coalesce NULL to 0.0 so that inner-join-absent evidence is correctly treated as no signal
- Builds the not-nearest condition from `_DARK_MATTER_NEAREST_FEATURES` (4 distance neighbourhood features)
- Returns early (no-op + warning) if the required signal or distance features are absent from the matrix
- Counts all positives per locus and dark matter positives per locus; marks a locus for removal only when the two counts are equal (all positives are dark matter)
- Removes all rows for dark matter loci via `left_anti` join; persists the filtered matrix so downstream training does not recompute it
- Returns `(filtered_matrix, stats_dict)` with before/after row counts, positive counts, studyLocusId counts, and % reductions

### `LocusToGeneStep`
- New parameter `filter_dark_matter: bool = False` (opt-in, no change to default behaviour)
- When enabled, calls `filter_dark_matter_loci()` after gold standard annotation and before model fitting
- Logs a warning if the filter has no effect (required features absent from `features_list`)
- Writes a JSON stats file to `{model_path}_dark_matter_stats.json`; creates the parent directory if needed for local paths

### `LocusToGeneConfig`
- Adds `filter_dark_matter: bool = False` — enable via Hydra override `step.filter_dark_matter=true`

## Stats file format

```json
{
  "before":        { "rows": 200536, "positives": 13060, "study_locus_ids": 8412 },
  "dark_matter":   { "positive_rows": 2928,  "study_locus_ids_removed": 2639 },
  "after":         { "rows": 152687, "positives": 10132, "study_locus_ids": 5773 },
  "pct_reduction": { "rows": 23.9,   "positives": 22.4,  "study_locus_ids": 31.4 }
}
```

## How to enable

```yaml
# Hydra config
step:
  filter_dark_matter: true
```

or CLI:
```bash
python -m gentropy step=locus_to_gene step.filter_dark_matter=true
```

### DIFF
--- a/src/gentropy/config.py
+++ b/src/gentropy/config.py
@@ -361,6 +361,7 @@ class LocusToGeneConfig(StepConfig):
     explain_predictions: bool = False
     wandb_credentials_path: str | None = None
     hf_credentials_path: str | None = None
+    filter_dark_matter: bool = False
     _target_: str = "gentropy.l2g.LocusToGeneStep"
 
 
--- a/src/gentropy/dataset/l2g_feature_matrix.py
+++ b/src/gentropy/dataset/l2g_feature_matrix.py
@@ -345,6 +345,7 @@ class L2GFeatureMatrix:
             positives.filter(no_signal & not_nearest)
             .groupBy("studyLocusId")
             .agg(f.count("*").alias("dark_matter_count"))
+            .persist()
         )
         dark_matter_loci = (
             total_positives_per_locus.join(
@@ -374,6 +375,7 @@ class L2GFeatureMatrix:
 
         self._df = self._df.join(dark_matter_loci, "studyLocusId", "left_anti")
         dark_matter_loci.unpersist()
+        dark_matter_positives_per_locus.unpersist()
 
         # Consolidate after-stats into a single aggregation (1 action)
         after_row = self._df.agg(
--- a/src/gentropy/dataset/l2g_feature_matrix.py
+++ b/src/gentropy/dataset/l2g_feature_matrix.py
@@ -309,19 +309,22 @@ class L2GFeatureMatrix:
             )
             return self, {}
 
-        # "No functional signal" condition
+        # "No functional signal" condition — coalesce to 0.0 so that NULL
+        # (produced by inner joins when no coloc/VEP evidence exists) is treated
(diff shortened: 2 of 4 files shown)


addramir:
@copilot resolve the merge conflicts in this pull request

PR #1227, opened 2026-05-26
feat(l2g): add LocusToGeneTrainTestSplitStep pipeline step

addramir:
## Summary

Introduces a dedicated \`locus_to_gene_train_test_split\` pipeline step and wires \`LocusToGeneStep\` to consume its output, separating gold-standard annotation + train/test splitting from model training.

### Motivation

Previously the \`locus_to_gene\` training step rebuilt the annotated gold-standard feature matrix and performed the train/test split on every run. This was expensive and made it impossible to:
- keep the held-out test set **stable** across feature matrix updates or hyperparameter sweeps
- inspect / audit the split before committing to a training run
- reuse the same split across multiple training configurations

### What this adds

**\`LocusToGeneTrainTestSplitStep\`** (\`src/gentropy/l2g.py\`)

A self-contained step that:
1. Loads the credible set and pre-computed feature matrix
2. Parses the gold standard curation (supports both the current native format and the legacy OTG curation format)
3. Annotates the gold standard with feature matrix values and selects the requested features
4. Performs a **hierarchical train/test split** (stratified by study locus to prevent data leakage) and writes \`train.parquet\` + \`test.parquet\`
5. Optionally re-uses a **predefined test set** from a previous run: the positive genes in the predefined test are used to exclude contaminating study loci from the new training set, while test features are refreshed from the current feature matrix

Split statistics (\`n_train\`, \`n_test\`, records lost due to predefined-test re-anchoring, etc.) are written as a JSON sidecar next to the training parquet.

In the predefined-test branch, \`train_sdf\` / \`test_sdf\` are persisted before the \`count()\` actions and unpersisted after the parquet writes so the join chain is not recomputed twice.

**\`LocusToGeneTrainTestSplitConfig\`** (\`src/gentropy/config.py\`)

Hydra config dataclass with a sensible default \`features_list\` matching the current production feature set. Registered under the \`step\` group as \`locus_to_gene_train_test_split\`.

**\`LocusToGeneStep\` changes**

- Removes \`gold_standard_curation_path\`, \`variant_index_path\`, \`gene_interactions_path\` — gold-standard parsing now lives exclusively in the split step
- Removes \`prepare_gold_standard()\` and \`_annotate_gold_standards_w_feature_matrix()\`
- \`run_train()\` now requires \`train_parquet_path\` and \`test_parquet_path\` (raises \`ValueError\` if absent) and loads them directly
- \`credible_set_path\` and \`feature_matrix_path\` are now **optional** (\`str | None = None\`); they are validated and loaded **only in predict mode** — train mode skips the I/O entirely
- \`LocusToGeneConfig\` reflects the same optionality (no longer \`MISSING\` for train-only runs)
- \`LocusToGeneTrainer.train()\` accepts \`presplit_train_df\` / \`presplit_test_df\` to bypass the internal \`generate_train_test_split\` call; raises \`ValueError\` if exactly one of the two is provided

**\`_parse_gold_standard\` robustness**

The \`unexpected_columns\` match case is guarded with \`if "missing_mandatory_columns" not in schema_issues\` so it cannot fire when mandatory columns are also missing, preventing the OTG-format branch from being silently shadowed.

### Usage

\`\`\`yaml
# Step 1 — split
step: locus_to_gene_train_test_split
credible_set_path: gs://…/credible_set
feature_matrix_path: gs://…/feature_matrix
gold_standard_curation_path: gs://…/gold_standard.parquet
train_parquet_path: gs://…/train
test_parquet_path: gs://…/test
test_size: 0.15

# Step 2 — train (credible_set_path / feature_matrix_path not required)
step: locus_to_gene
run_mode: train
train_parquet_path: gs://…/train
test_parquet_path: gs://…/test
\`\`\`

To lock the test set across runs, pass \`predefined_test_parquet_path\` pointing to the \`test.parquet\` from a previous run.

## Test plan

- [x] \`_write_split_stats\`: JSON sidecar written correctly for local paths
- [x] \`_parse_gold_standard\`: exact schema, extra columns silently dropped, unrecognised schema raises \`TypeError\`
- [x] \`LocusToGeneTrainer.train\` with \`presplit_train_df\`/\`presplit_test_df\`: confirms bypass path is used (identity check on DataFrames)
- [x] \`LocusToGeneTrainer.train\` with only one presplit df: confirms \`ValueError\` is raised
- [x] Integration smoke test: run \`locus_to_gene_train_test_split\` on a dev feature matrix and confirm parquets are written with expected row counts
- [x] Re-run with \`predefined_test_parquet_path\` set to the previous \`test.parquet\` and confirm the test gene overlap is zero between train and test positives


src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+                "n_train": n_train,
+                "n_lost_total": n_original_total - n_test_new - n_train,
+            }
+            self._write_split_stats(split_stats, train_parquet_path)
+            logging.info("Split stats: %s", split_stats)
+        else:
+            train_df, test_df = annotated_fm.generate_train_test_split(
+                test_size=test_size,
+                verbose=True,
+                label_encoder=label_encoder,
```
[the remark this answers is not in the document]
addramir: Fixed in b6fe9c5. Split stats are now written in both branches — the fresh-split path writes `{n_train, n_test, test_size}` and the predefined-test path writes the full stats dict as before.

src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+
+            parsed = urlparse(log_path)
+            bucket = storage.Client().bucket(parsed.hostname)
+            bucket.blob(parsed.path.lstrip("/")).upload_from_string(
+                payload, content_type="application/json"
+            )
+        else:
+            from pathlib import Path
+
+            Path(log_path).write_text(payload)
```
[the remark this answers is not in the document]
addramir: Fixed in b6fe9c5. Added `p.parent.mkdir(parents=True, exist_ok=True)` before `p.write_text(payload)` so the directory is created if it doesn't exist yet.

src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+                label_encoder=label_encoder,
+                label_col=annotated_fm.label_col,
+            )
+
+        session.spark.createDataFrame(train_df).coalesce(1).write.mode(
+            session.write_mode
+        ).parquet(train_parquet_path)
+        session.spark.createDataFrame(test_df).coalesce(1).write.mode(
+            session.write_mode
+        ).parquet(test_parquet_path)
```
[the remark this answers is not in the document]
addramir: Fixed in b6fe9c5. Removed `coalesce(1)` from both write calls. The predefined-test branch now writes Spark DataFrames directly (no pandas round-trip), and the fresh-split branch lets Spark use its default partitioning.

src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+            variant_index_path=variant_index_path,
+            gene_interactions_path=gene_interactions_path,
+        )
+
+        # Build annotated feature matrix for gold standard loci
+        annotated_fm = (
+            gold_standard.build_feature_matrix(feature_matrix, credible_set)
+            .select_features(features_list)
+            .persist()
+        )
```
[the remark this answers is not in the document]
addramir: Fixed in b6fe9c5. Added `annotated_fm._df.unpersist()` after both branches complete their writes.

src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+            "Train/test split written: %d train rows, %d test rows.",
+            len(train_df),
+            len(test_df),
+        )
+
+    @staticmethod
+    def _parse_gold_standard(
+        session: Session,
+        gold_standard_curation_path: str,
+        credible_set: StudyLocus,
```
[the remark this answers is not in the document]
addramir: Not applicable after the refactor. `LocusToGeneStep.prepare_gold_standard` was removed entirely in this PR — `_parse_gold_standard` in `LocusToGeneTrainTestSplitStep` is now the sole implementation, so there is no duplication.

src/gentropy/l2g.py
```diff
@@ -157,6 +158,264 @@ def __init__(
...
+            )
+
+            # Apply the same label encoding that generate_train_test_split uses.
+            for col in {annotated_fm.label_col, "goldStandardSet"}:
+                for df in [train_df, test_df]:
+                    if col in df.columns and df[col].dtype == object:
+                        df[col] = df[col].map(label_encoder)
+
+            n_test_new = len(test_df)
+            n_train = len(train_df)
```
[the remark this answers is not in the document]
addramir: Fixed in b6fe9c5. The predefined-test branch is now fully Spark-based: `left_anti` join to build the train set, `inner` join to refresh test features, and `create_map` for label encoding. No `toPandas()` calls remain in that path, so the data never leaves the cluster.

src/gentropy/l2g.py
```diff
@@ -157,6 +159,267 @@ def __init__(
...
+            case {**extra} if not extra:
+                return L2GGoldStandard(
+                    _df=gold_standard_raw,
+                    _schema=L2GGoldStandard.get_schema(),
+                )
+            case {"unexpected_columns": extra_columns}:
+                return L2GGoldStandard(
+                    _df=gold_standard_raw.drop(*extra_columns),
+                    _schema=L2GGoldStandard.get_schema(),
+                )
```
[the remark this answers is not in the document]
addramir: Fixed in 60c1055. Added `if "missing_mandatory_columns" not in schema_issues` as a guard on the `unexpected_columns` case so it only fires when the schema has no missing mandatory columns — preventing it from shadowing the OTG-specific branch or masking a genuinely invalid schema.

src/gentropy/l2g.py
```diff
@@ -157,6 +159,267 @@ def __init__(
...
+                "n_test_new": n_test_new,
+                "n_lost_test": n_original_test - n_test_new,
+                "n_train": n_train,
+                "n_lost_total": n_original_total - n_test_new - n_train,
+            }
+            self._write_split_stats(split_stats, train_parquet_path)
+            logging.info("Split stats: %s", split_stats)
+            train_sdf.write.mode(session.write_mode).parquet(train_parquet_path)
+            test_sdf.write.mode(session.write_mode).parquet(test_parquet_path)
+            n_written_train, n_written_test = n_train, n_test_new
```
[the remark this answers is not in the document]
addramir: Fixed in 60c1055. Added `.persist()` on both `train_sdf` and `test_sdf` before the `count()` calls, and `.unpersist()` after the parquet writes. The count actions now materialise the cached data rather than re-running the join chain a second time during `.write`.

src/gentropy/method/l2g/trainer.py
```diff
@@ -297,22 +299,28 @@ def train(
...
-            verbose=True,
-            label_encoder=self.model.label_encoder,
-            label_col=self.feature_matrix.label_col,
-        )
+        if presplit_train_df is not None and presplit_test_df is not None:
+            self.train_df = presplit_train_df
+            self.test_df = presplit_test_df
+        else:
+            # Create held-out test set using hierarchical splitting
+            self.train_df, self.test_df = self.feature_matrix.generate_train_test_split(
```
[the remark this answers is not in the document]
addramir: Fixed in 60c1055. Added an early check: `if (presplit_train_df is None) != (presplit_test_df is None): raise ValueError(...)` so passing exactly one of the two DataFrames is immediately surfaced as a programming error rather than silently falling back to an internal split.

src/gentropy/config.py
```diff
@@ -290,10 +290,7 @@ class LocusToGeneConfig(StepConfig):
     feature_matrix_path: str = MISSING
     predictions_path: str | None = None
     l2g_threshold: float = 0.05
```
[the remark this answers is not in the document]
addramir: Fixed in 60c1055. Changed `credible_set_path` and `feature_matrix_path` in `LocusToGeneConfig` from `MISSING` to `str | None = None`. The step now validates and loads them only when `run_mode == "predict"`, so a train-mode config that omits these paths is valid.

src/gentropy/l2g.py
```diff
@@ -248,9 +509,8 @@ def __init__(
...
 
         # Train io
-        self.gold_standard_curation_path = gold_standard_curation_path
-        self.gene_interactions_path = gene_interactions_path
-        self.variant_index_path = variant_index_path
+        self.train_parquet_path = train_parquet_path
+        self.test_parquet_path = test_parquet_path
         self.train_on_full_dataset = train_on_full_dataset
 
         # Train parameters
```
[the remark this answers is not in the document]
addramir: Fixed in 60c1055. Moved the `StudyLocus.from_parquet` and `L2GFeatureMatrix` loads inside the `run_mode == "predict"` branch (with `ValueError` guards for missing paths). Train mode no longer touches these inputs at all.

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            feature_matrix_path (str): Path to the L2G feature matrix input dataset
+            gold_standard_curation_path (str): Path to the gold standard curation file (parquet or JSON)
+            train_parquet_path (str): Output path for the training split parquet
+            test_parquet_path (str): Output path for the held-out test split parquet
+            features_list (list[str]): Features to select from the feature matrix
+            test_size (float): Proportion of study loci assigned to the test split. Defaults to 0.15.
+            variant_index_path (str | None): Path to the variant index (required for OTG gold standard)
+            gene_interactions_path (str | None): Path to the PPI dataset (required for OTG gold standard)
+            predefined_test_parquet_path (str | None): Path to an existing test-split parquet (produced by a previous run of this step). When provided the test set is loaded as-is and the training set is derived by removing all studyLocusIds from the annotated feature matrix whose positive genes overlap with the test set's positive genes. ``test_size`` is ignored. Defaults to None (performs a fresh hierarchical split).
```
project-defiant: NIT: split the line please.

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            split_stats_path (str | None): Explicit path for the split statistics JSON file. Defaults to ``<train_parquet_path>_split_stats.json``.
+        """
+        credible_set = StudyLocus.from_parquet(
+            session, credible_set_path, recursiveFileLookup=True
+        )
+        feature_matrix = L2GFeatureMatrix(
```
project-defiant, line 216: Is there any reason why L2GFeatureMatrix does not implement the from_parquet method?

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            .persist()
+        )
+
+        label_encoder = {"negative": 0, "positive": 1}
+
+        if predefined_test_parquet_path:
+            predefined_test_sdf = session.spark.read.parquet(predefined_test_parquet_path)
+
+            n_original_total: int = annotated_fm._df.count()
+            n_original_test: int = predefined_test_sdf.count()
```
project-defiant: Persist before counting. Will reduce IO ops

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+
+            # Positive gene IDs from the predefined test set.
+            test_positive_genes_sdf = predefined_test_sdf.filter(
+                f.col("goldStandardSet").isin([1, "positive"])
+            ).select("geneId").distinct()
```
project-defiant, line 248: What is the type of goldStandardSet field. I would say that it should be either str or int. (Hopefully small int or Boolean) to reduce size

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+        # Build annotated feature matrix for gold standard loci
+        annotated_fm = (
+            gold_standard.build_feature_matrix(feature_matrix, credible_set)
+            .select_features(features_list)
+            .persist()
+        )
+
+        label_encoder = {"negative": 0, "positive": 1}
+
+        if predefined_test_parquet_path:
```
project-defiant, line 238: I can imagine case when the future run will have encoder labels swapped 0->1, so there will be discrepancy probably

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            # Apply label encoding in Spark (handles both string and already-encoded int values).
+            label_map = f.create_map(f.lit("negative"), f.lit(0), f.lit("positive"), f.lit(1))
+            train_sdf = train_sdf.withColumn(
+                "goldStandardSet",
+                f.coalesce(label_map[f.col("goldStandardSet")], f.col("goldStandardSet").cast("int")),
+            )
+            test_sdf = test_sdf.withColumn(
+                "goldStandardSet",
+                f.coalesce(label_map[f.col("goldStandardSet")], f.col("goldStandardSet").cast("int")),
+            )
```
project-defiant: Kinda duplicate ops, but I get that due to previous test set labels from current annotated fm are not 100% certain and need to be aligned with predefined test set

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+
+            train_sdf = train_sdf.persist()
+            test_sdf = test_sdf.persist()
```
project-defiant: Ideally do before many joins to prevent rereading

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            split_stats: dict[str, Any] = {
+                "n_original_total": n_original_total,
+                "n_original_test": n_original_test,
+                "n_test_new": n_test_new,
+                "n_lost_test": n_original_test - n_test_new,
+                "n_train": n_train,
+                "n_lost_total": n_original_total - n_test_new - n_train,
+                "train": self._compute_set_stats(train_sdf.toPandas()),
+                "test": self._compute_set_stats(test_sdf.toPandas()),
+            }
```
project-defiant, line 295: Please define the schema for that as pydantic model

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            self._write_split_stats(split_stats, train_parquet_path, split_stats_path)
+            logging.info("Split stats: %s", split_stats)
```
project-defiant: Set up non default logger for this module before using it

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+                session.write_mode
+            ).parquet(test_parquet_path)
+
+        annotated_fm._df.unpersist()
+        logging.info(
+            "Train/test split written: %d train rows, %d test rows.",
+            n_written_train,
+            n_written_test,
+        )
+
```
project-defiant, line 326: You have a clear contract in two logic branches that is repeated. Please simplify and ensure branching only generates data frames which are later written after code branches are merged.

src/gentropy/l2g.py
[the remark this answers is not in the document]
project-defiant: Please, check for the Gold Standard in data sources and datasets and ensure the dead code there is also removed.

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+                    "goldStandardSet",
+                ],
+                "unexpected_columns": [
+                    "association_info",
+                    "gold_standard_info",
+                    "metadata",
+                    "sentinel_variant",
+                    "trait_info",
+                ],
+            }:
```
project-defiant, line 383: Do we really still need to allow for both gold standard schemas post this PR?

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+                if gene_interactions_path is None:
+                    raise ValueError("Interactions are required for parsing OTG curation.")
+                if variant_index_path is None:
+                    raise ValueError("Variant Index is required for parsing OTG curation.")
```
project-defiant, line 387: Remove the optionality, make the paths required as str, remove these checks

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+                    variant_index=variant_index,
+                    study_locus_overlap=study_locus_overlap,
+                    interactions=interactions,
+                )
+            case _:
+                raise TypeError("Incorrect gold standard dataset provided.")
+
+    @staticmethod
+    def _compute_set_stats(df: pd.DataFrame) -> dict[str, Any]:
+        """Compute descriptive statistics for one split partition.
```
project-defiant: Mention that the split partition is either `test` or `train` (row-base slice from annotated feature matrix)

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+
+        Args:
+            df (pd.DataFrame): Split DataFrame with integer-encoded ``goldStandardSet`` (0/1).
+
+        Returns:
+            dict[str, Any]: Stats dict containing counts of positive/negative rows, unique loci,
```
project-defiant: Make a pydantic schema and typed dict with proper value types (always int in your case, never anything else)

src/gentropy/l2g.py
```diff
@@ -157,6 +159,308 @@ def __init__(
...
+            bucket.blob(parsed.path.lstrip("/")).upload_from_string(
+                payload, content_type="application/json"
+            )
+        else:
+            from pathlib import Path
+
+            p = Path(log_path)
+            p.parent.mkdir(parents=True, exist_ok=True)
+            p.write_text(payload)
+        logging.info("Split stats written to %s", log_path)
```
project-defiant: BREAKING:  
1. The thing you write is not logs, but statistics (metrics from the step)
2. **Please do not use the `google` packages, if you need to write the dataset to multiple paths, use pyspark, or alternatively use pandas, they have proper dispatcher methods that write to gcs or local paths based on location. We aim to remove google packages from gentropy.** This is one thing that has to be updated if we want to include this PR in dev.
3. If you need to assume parent directories in the output path that does not exist, just use pyspark, it will make them in proper file system for you,
4. for logging again use the non-default logger.

src/gentropy/l2g.py
```diff
@@ -169,14 +473,11 @@ def __init__(
         download_from_hub: bool,
         cross_validate: bool,
         train_on_full_dataset: bool,
-        credible_set_path: str,
-        feature_matrix_path: str,
+        credible_set_path: str | None = None,
```
project-defiant, line 478: If I am completely honest, this PR makes a lot of things more convoluted. Currently both `tran` and `predic` steps were required to use `credible_set_path` and `feature_matrix_path`, now you make them optional while including `train_parquet_path` and `test_parquet_path` which are again optional. 

This step is a mess. 

The only way to make it cleaner is to split this into 
1. train (requires `train_parquet_path` and `test_parquet_path` to be non optional that are output from `LocusToGeneTrainTestSplitStep`)
2. predict (requires still `credible_set_path` and `feature_matrix_path`) 

Probably Irene has some changes there @ireneisdoomed and we should merge them first. 


One could imagine, if someone else would like to run L2G, how difficult would it be to understand which parameters are actually required for prediction.

src/gentropy/l2g.py
```diff
@@ -274,15 +575,17 @@ def __init__(
...
+            if not credible_set_path:
+                raise ValueError("credible_set_path is required for predict mode.")
+            if not feature_matrix_path:
+                raise ValueError("feature_matrix_path is required for predict mode.")
+            self.credible_set = StudyLocus.from_parquet(
+                session, credible_set_path, recursiveFileLookup=True
+            )
+            self.feature_matrix = L2GFeatureMatrix(
+                _df=session.load_data(feature_matrix_path, "parquet"),
+            )
```
project-defiant, line 590: This gets super messy, splitting the step into 2 (train and predict) should clean up all these checks.

src/gentropy/l2g.py
```diff
@@ -436,10 +640,16 @@ def run_train(self) -> None:
         """Run the training step.
 
         Raises:
-            ValueError: If features list is not provided for model training.
+            ValueError: If features list or presplit parquet paths are not provided.
         """
         if self.features_list is None:
             raise ValueError("Features list is required for model training.")
+        if not (self.train_parquet_path and self.test_parquet_path):
```
project-defiant, line 649: Given step split this would no longer be needed.

src/gentropy/l2g.py
```diff
@@ -461,8 +670,25 @@ def run_train(self) -> None:
...
+        # Labels are decoded back to strings so generate_train_test_split works if
+        # export_to_hugging_face_hub is called later.
+        label_decoder = {0: "negative", 1: "positive"}
+        combined = pd.concat([presplit_train_df, presplit_test_df], ignore_index=True)
+        if combined["goldStandardSet"].dtype != object:
+            combined["goldStandardSet"] = combined["goldStandardSet"].map(label_decoder)
+        feature_matrix = L2GFeatureMatrix(
+            _df=self.session.spark.createDataFrame(combined),
+            with_gold_standard=True,
+        ).select_features(self.features_list)
```
project-defiant, line 693: use native spark operations instead of mapping to pandas and then to spark back and forth. As far as I understand the previous step (TrainTestSplit) mapped the integer values, why are you remapping them here to strings again ? Since we control the process of building GoldStandard, we should simplify it and use just one encoding. 

```
            label_map = f.create_map(f.lit("negative"), f.lit(0), f.lit("positive"), f.lit(1))
            train_sdf = train_sdf.withColumn(
                "goldStandardSet",
                f.coalesce(label_map[f.col("goldStandardSet")], f.col("goldStandardSet").cast("int")),
            )
            test_sdf = test_sdf.withColumn(
                "goldStandardSet",
                f.coalesce(label_map[f.col("goldStandardSet")], f.col("goldStandardSet").cast("int")),
            )
```

tests/gentropy/method/test_l2g_trainer.py
```diff
@@ -136,6 +136,41 @@ def test_train_on_full_dataset_logs_second_wandb_run(
...
+        verbose=False,
+        label_encoder=l2g_model.label_encoder,
+        label_col=filled_fm.label_col,
+    )
+    trainer = LocusToGeneTrainer(
+        model=l2g_model,
+        feature_matrix=filled_fm,
+        features_list=features_list,
+    )
+    trained_model = trainer.train(
```
project-defiant, line 159: Hope running training does not log to W&B during tests.

project-defiant (changes requested):

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. Wrapped the docstring line across 5 lines.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. Added `logger = logging.getLogger(__name__)` at module level and replaced all bare `logging.info()` calls in `LocusToGeneTrainTestSplitStep` with `logger.info()`.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. `predefined_test_sdf` is now `.persist()`-ed immediately after loading, before any counts or joins.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. `predefined_test_sdf` is persisted right after `session.spark.read.parquet()`, before the `test_positive_genes_sdf`, `contaminating_sdf`, and `test_pairs_sdf` joins.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. Both branches now only produce `train_sdf`/`test_sdf`/`split_stats`. The `_write_split_stats` call and both `.write.parquet()` calls are in a single block after the `if/else`. Unpersisting is handled via a `to_unpersist` list populated inside the branches.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. Docstring now reads: *"A train or test partition — a row-based slice of the annotated feature matrix with integer-encoded `goldStandardSet` (0 = negative, 1 = positive)."*

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. Added a `_SetStats(TypedDict)` with all-`int` fields at module level. The optional `n_unique_positive_gene_disease_pairs` field uses `NotRequired[int]` (Python 3.11+). `_compute_set_stats` now returns `_SetStats` instead of `dict[str, Any]`.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6.

1. Renamed the concept: the function/variable is about *statistics*, not logs — `log_path` → `stats_path` throughout.
2. Removed `google.cloud.storage` entirely. `_write_split_stats` now calls `pd.DataFrame([stats]).to_json(stats_path, orient="records", indent=2)`. `pandas[gcp]` is already a declared dependency, so `gcsfs` is available and pandas dispatches `gs://` paths transparently — no manual bucket/blob handling needed.
3. The local-path `mkdir(parents=True)` guard is preserved for non-GCS paths.

src/gentropy/l2g.py
addramir: Fixed in e41b9f6. `run_train()` now:
1. Loads both parquets as Spark DataFrames and `.persist()`s them.
2. Builds the `L2GFeatureMatrix` via `train_sdf.union(test_sdf)` with a Spark `create_map` label decode (int→string), no pandas involved.
3. Calls `.toPandas()` once per split *after* the union (hits the cache), then `.unpersist()`s.

The label decode in the union is needed because `export_to_hugging_face_hub` calls `generate_train_test_split` which expects string labels.

src/gentropy/l2g.py
addramir: Agreed — splitting `LocusToGeneStep` into dedicated train and predict steps is the right long-term direction. We're deferring that refactor to a follow-up PR to keep this one focused on the split step itself.

src/gentropy/l2g.py
addramir: Agreed — splitting `LocusToGeneStep` into dedicated train and predict steps is the right long-term direction. We're deferring that refactor to a follow-up PR to keep this one focused on the split step itself.

src/gentropy/l2g.py
addramir: Agreed — splitting `LocusToGeneStep` into dedicated train and predict steps is the right long-term direction. We're deferring that refactor to a follow-up PR to keep this one focused on the split step itself.

src/gentropy/l2g.py
addramir: No strong reason — `L2GFeatureMatrix` is a plain Python class rather than a `Dataset` subclass, so it doesn't inherit `from_parquet`. Adding one would require threading through the `with_gold_standard` flag, which is doable but a separate concern from this PR. Happy to add it as a follow-up if that's the preference.

src/gentropy/l2g.py
addramir: Post this PR, `LocusToGeneTrainTestSplitStep` always writes integer-encoded labels (0/1) via the Spark `create_map` encoder. The `.isin([1, "positive"])` filter is defensive for parquets produced before this step existed (where `goldStandardSet` was a string). Once the step is the sole entry point for producing train/test splits this dual-type guard can be dropped.

src/gentropy/l2g.py
addramir: Acknowledged. The encoding is a fixed contract defined once in `LocusToGeneTrainTestSplitStep` as `{"negative": 0, "positive": 1}` and applied via Spark `create_map`. The `coalesce(label_map[col], col.cast("int"))` pattern in the predefined-test branch passes through already-encoded int values as-is — so if a future run produced a parquet with a swapped encoding, the join would silently use wrong labels. The safeguard is that the encoding constant lives in one place; any change there would be explicit. Could add an assertion on the distinct values in the loaded parquet as a runtime guard if preferred.

src/gentropy/l2g.py
addramir: Correct — the redundancy is intentional. The predefined test set was produced by a previous run and may carry label assignments that have since changed in the annotated FM (e.g. a study-gene pair reclassified positive→negative). Re-joining against the current annotated FM refreshes the feature values while the `test_pairs_sdf` join preserves the original (studyLocusId, geneId) identity of the test set.

src/gentropy/l2g.py
addramir: Partially addressed in e41b9f6: added `_SetStats(TypedDict)` with all-`int` fields (plus `NotRequired[int]` for the optional gene-disease-pairs field) and tightened `_compute_set_stats` to return `_SetStats`. The top-level `split_stats` dict has two different shapes depending on the branch (fresh split carries `n_train/n_test/test_size`; predefined-test carries `n_original_total/n_original_test/...`), so a single model would need to be a union or have optional fields. Happy to add explicit TypedDicts for both shapes if you'd like, but left it as `dict[str, Any]` for now given the two-shape situation.

src/gentropy/l2g.py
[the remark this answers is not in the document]
addramir: `OpenTargetsL2GGoldStandard` in `datasource/open_targets/l2g_gold_standard.py` is still live — it is imported and called by `L2GGoldStandard.from_otg_curation`, which is invoked from `_parse_gold_standard` in this step when the OTG schema is detected. No dead code to remove there.

src/gentropy/l2g.py
addramir: Intentionally keeping it. The OTG curation format (`genetics-gold-standards` repo) is still the source of truth for some workflows. `_parse_gold_standard` detects the schema automatically so callers don't need to know which format they're passing.

src/gentropy/l2g.py
addramir: `variant_index_path` and `gene_interactions_path` are optional precisely because they are only required for the OTG curation branch, which we're keeping. Both are validated at runtime with a `ValueError` if the OTG schema is detected but the paths are absent. Making them required unconditionally would break the common case where only the native format is used.

tests/gentropy/method/test_l2g_trainer.py
addramir: Confirmed — the test passes `wandb_run_name=None` and all W&B calls in the trainer are guarded by `if wandb_run_name:`, so no W&B logging occurs.

addramir:
@project-defiant thank you for reviewing! Answered your comments.

project-defiant (approved):

PR #1229, opened 2026-05-28
feat(method): add FUMA Gene2Func hypergeometric gene-set enrichment method

xyg123:
## ✨ Context

To leverage the available gene prioritisation results from the open targets platform and produce relevant biosamples for each study and trait, this PR implements the computationally simple hypergeometric test used in FUMA to infer biosample significance and can be a starting point for the biosample enrichment catalog for studies which do not have full summary statistics. 

https://github.com/opentargets/issues/issues/4389

- This PR adds FUMA Gene2Func tissue enrichment functionality in src/gentropy/method/fuma_gene2func.py, which is a hypergeometric enrichment approach for GWAS-prioritised genes
- Takes any scored gene DataFrame (L2G predictions or OT association scores) and a long-format gene sets DataFrame (e.g. GTEx DEGs), and tests whether prioritised genes are over-represented in each gene set relative to a per-group background universe.
- Automatically resolves studyLocusId -> studyId via a credible set DataFrame, and optionally joins to a study index to produce per-(study, disease) results — group columns are inferred from whatever identifiers are present in the input.
- Returns one row per (group × gene set) with enrichment counts, fold enrichment, one-sided hypergeometric p-value, Bonferroni correction, and BH FDR.

## 🚦 Before submitting


src/gentropy/method/fuma_gene2func.py
```diff
@@ -0,0 +1,355 @@
...
+    - diseaseId / targetId / score (OT association scores, no extras)
+      -> output grouped by diseaseId
+    - Any other scored DataFrame -- group columns are inferred automatically
+      as everything that is not gene_col or score_col.
+
+    Gene sets are supplied as a long-format Spark DataFrame with one row per
+    (set label, gene).  GTEx DEG sets, MSigDB gene sets, or any other collection
+    can be passed in this format.
+
+    Example::
```
project-defiant, line 39: This example will not render properly in gentropy docs. Have a look how to implement doctest that are not expected to run.

src/gentropy/method/fuma_gene2func.py
```diff
@@ -0,0 +1,406 @@
...
+                )
+            else:
+                disease_mapping = disease_mapping.withColumnRenamed(
+                    study_disease_col, "diseaseId"
+                )
+            working = working.join(disease_mapping, on="studyId", how="inner")
+
+        # 3. Infer group columns
+        non_group = {gene_col, score_col}
+        group_cols = [c for c in working.columns if c not in non_group]
```
addramir: Inferring the group columns by subtraction means any column in `scored_df` that is neither `gene_col` nor `score_col` silently becomes part of the enrichment key, including columns that are not identifiers at all.

The case that worries me is `L2GPrediction.df`, whose schema is `studyLocusId, geneId, score, features`, with `features` an `array<struct<name, value>>` feature vector that differs for essentially every row. `studyLocusId` is resolved and dropped, so the inferred grouping becomes `["studyId", "features"]`, and `(studyId, features)` is effectively a row identifier. Spark accepts an array-of-struct as an aggregation key, so nothing fails: every gene ends up in its own group with `N = K = n = k = 1`, and every p-value is exactly 1.0.

Reproduced on this branch by passing an L2G-shaped frame with a `features` column and a two-set gene set collection:

```
features    |studyId|setName|k_gene_set|k_overlap|n_background|n_input|p_value
[{f1, 1.0}] |S1     |SET1   |1         |1        |1           |1      |1.0
[{f1, 2.0}] |S1     |SET1   |1         |1        |1           |1      |1.0
... one row per gene, all p_value = 1.0
```

The only visible clue is `features` appearing as a group column in the output; the counts and statistics all look well-formed, so a user would most likely read this as "no tissue enrichment" rather than as a malformed input. The same applies to any other per-row column someone happens to leave in, such as `beta` or `pValueMantissa`.

A few options, in increasing strictness:

1. Accept an explicit `group_cols: list[str] | None` argument and only fall back to inference when it is not supplied.
2. Restrict inference to a known set of identifier columns (`studyId`, `diseaseId`, and similar) and ignore the rest.
3. Keep inference, but raise a clear error if any inferred group column has a non-primitive type, since a struct, array, or map is never a sensible enrichment key.

Option 3 alone would already turn this from a silent wrong answer into an actionable error message.

src/gentropy/method/fuma_gene2func.py
```diff
@@ -0,0 +1,406 @@
...
+                    f"that can be renamed automatically, but found columns: "
+                    f"{gene_sets_df.columns}. Rename the gene column to '{gene_col}' "
+                    f"before passing gene_sets_df."
+                )
+            gene_sets_df = gene_sets_df.select(
+                set_col, f.col(candidate_cols[0]).alias(gene_col)
+            )
+
+        # Gene set universe: restrict background to genes in any set
+        universe = gene_sets_df.select(gene_col).distinct()
```
addramir, line 169: `universe` is deduplicated here, but the counting joins further down are not: `k_gene_set` and `k_overlap` both join the raw `gene_sets_df` and aggregate with `f.count("*")`, which counts join rows rather than distinct genes.

`scored` is guaranteed unique per (group, gene) by the collapsing `groupBy` in `gene2func_enrichment`, so `n_background` and `n_input` are safe. `gene_sets_df` carries no such guarantee, and duplicated `(setName, geneId)` pairs are easy to hit in practice: per-subregion GTEx rows, several transcripts mapping to one gene, or a union of up- and down-regulated tables.

With the sets duplicated and everything else held constant:

```
clean : SET1 N=10 K=5  n=5 k=5  p=0.003968
dupset: SET1 N=10 K=10 n=5 k=10 p=0.0
```

`n_background` stays at 10 because it comes from the deduplicated `scored` side, while `K` and `k` double. That leaves `K == N` and `k == 2n`, which is not a valid hypergeometric configuration, and `hypergeom.sf` returns 0.0 without complaining. So duplication pushes p-values towards zero, which is the worst direction for an enrichment method, and the degree of inflation depends on each set's duplication rate, so some tissues would inflate more than others.

Suggested fix: deduplicate once when `gene_sets_df` is reduced to `(set_col, gene_col)` just above, and reuse that frame for the universe and for both counting joins. Using `f.countDistinct(gene_col)` in the two aggregations would work too.

It would also be worth adding a cheap sanity guard after the counts are assembled, for example asserting `k_gene_set <= n_background` and `k_overlap <= n_input`, so any future violation surfaces as an error instead of a p-value of 0.0.

src/gentropy/method/fuma_gene2func.py
xyg123: Implemented the suggested **options 3 and 1** together.

**Option 3** is the minimum fix. Inference still runs, but any inferred column with an `ArrayType`, `MapType` or `StructType` now raises instead of becoming a grouping key:

```
ValueError: Inferred group columns ['features'] have non-primitive types and
cannot be used as enrichment grouping keys. Pass group_cols explicitly to
select the identifier columns to group by. Columns available: [...]
```

**Option 1** is implemented on top of this, because option 3 alone converts the silent wrong answer into a hard block.  So there's now an optional `group_cols: list[str] | None`, validated against the post-resolution columns, with inference as the fallback when it's omitted:

```python
FumaGene2Func.gene2func_enrichment(
    scored_df=l2g_predictions,   # studyLocusId, geneId, score, features
    gene_sets_df=gtex_deg,
    gene_col="geneId",
    score_col="score",
    credible_set_df=cs.select("studyLocusId", "studyId"),
    group_cols=["studyId"],      # `features` is ignored
)
```

On the L2G-shaped frame that collapses all genes into one study group and drops `features` from the output.

Three new tests cover it:

| case | expected |
|---|---|
| `features` left in, no `group_cols` | raises, matching `non-primitive` |
| same frame, `group_cols=["studyId"]` | one row, grouped by `studyId`, no `features` column |
| `group_cols=["nonexistentColumn"]` | raises, matching `not present after identifier` |

src/gentropy/method/fuma_gene2func.py
xyg123: a .distinct() function has been added and a sanity check guard for the various scenarios outlined has also been added.

PR #1232, opened 2026-06-12
perf(coloc): native and fused coloc_pip_ecaviar

d0choa:
Part of opentargets/issues#4415.

## Summary

Optimises the `colocalisation` step (`coloc_pip_ecaviar`), which at baseline took **4h18m** with **3,021 `ExecutorLostFailure`**. This PR brings it to **~22.5 min / 539 OOM** on the same input (`26.06.0-dev1`) and same-shaped cluster — **≈11.5× faster, ~14× cheaper** (actual billed **£469.81 → £33.58 per run**) — with output **semantically identical** (full-output comparison below; characterization + R-reference semantic tests unchanged).

## Changes

**Overlap pipeline** (`study_locus.py`)
- Drop `.repartition("chromosome")` in `StudyLocus._overlapping_peaks` (it collapsed the combinatorial self-join into ≤25 severely skewed partitions, chr6/HLA).
- The aligned-overlap recompute (up to 4× in the original, across the two sub-methods + beta-ratio) is eliminated by the fused single-pass aggregation below — so no extra caching of the final overlaps is needed (`find_overlaps` still persists the genuinely-reused intermediate).

**Methods → fully native Spark SQL** (no Python UDFs, no `collect_list`)
- New `get_logsum_column` (native logsumexp) in `common/stats.py`.
- `Coloc` rewritten native (deleted `_get_posteriors`).
- `ColocPIP` rewritten native — the PIP approximation reduces to three per-overlap sums (`S1=Σpip1`, `S2=Σpip2`, `B=Σpip1·pip2`); `_floored_pip` / `_pip_posteriors` are shared helpers.
- `ECaviar` native.
- `numberColocalisingVariants` as a native `sum(when(source=='both', 1))` count.

**`ColocPIPECaviar` fused**
- A single `groupBy` computes every metric for both methods (ColocPIP `h3/h4`, eCAVIAR `clpp`, variant count) **and** the beta-ratio (folded in — `rightStudyType` is functionally determined by the right locus, so the grouping granularity matches), replacing the previous two groupBys + inner merge join + a separate ~2.4 TiB beta-ratio shuffle.

## A/B (full runs to DONE, same input + same-shaped cluster)

| Stage | Wall-clock | `ExecutorLostFailure` | Cost (£/run, billed) |
|---|---|---|---|
| Baseline | 4h18m | 3,021 | 469.81 |
| native methods + native count | 51m | 1,817 | |
| fused agg + single beta-ratio | 43m | 1,006 | |
| **+ beta-ratio folded (PR tip)** | **22.5m** | **539** | **33.58** |

Costs are actual billed amounts (GBP, net) from the Google Cloud Console billing report (CSV export), by cluster label. The baseline cluster billed £939.62 for **two** step executions → £469.81/run; the optimised run is one execution.

> **Measurement note.** The first post-baseline run (`opt-001`) — dropping the `repartition("chromosome")` skew — **OOM-aborted before completion** (stage 38, YARN `exit-143`): the row-wise `pandas_udf` still present at that point exhausted off-heap Python memory. So the skew fix has no standalone completed A/B row; it's folded into the first completed row (51m, "native methods"). (An interim step-level `.persist()` on the aligned overlaps was carried through the early runs but later removed as redundant — the fused single-pass aggregation already reads them once.)

## Semantic validation (full output)

Keyed full-outer join of this PR's output (`opt-006`) against the baseline output on the **same input**, over **all 235,953,548 rows**:

- identical row count and **identical key set** — 0 only-in-baseline, 0 only-in-optimised; keys unique in both;
- `numberColocalisingVariants` and `chromosome`: **0 mismatches**;
- `h3` / `h4` / `clpp` / `betaRatioSignAverage`: 0 rows exceed a 1e-6 tolerance; **actual max abs diff ~5e-14** (`h3`/`h4`), ~1.7e-15 (`clpp`), 0.0 (`betaRatioSignAverage`) — float-summation-order noise. `betaRatioSignAverage` null in **both** for 98,162,488 pairs (agree).

**The optimised output is semantically equivalent to baseline.**

## Testing

Characterization tests pin Coloc / ColocPIP / `ColocPIPECaviar` numeric output at ≤1e-9; the existing R-reference semantic tests are unchanged. Full colocalisation method + step + dataset + `common/stats` suites pass; ruff clean.

## Not in this PR

- **Cluster topology** (non-preemptible workers, executor packing, drop EFM) — orchestration PR opentargets/orchestration#209, also under opentargets/issues#4415.
- The residual ~539 `ExecutorLostFailure` are a per-executor physical-memory overrun in the fused agg+parquet-write stage (**not** skew, **not** disk — zero spill, uniform across the fleet). Diagnosed and reduced **~80% (535 → 108)** by raising `spark.sql.shuffle.partitions` 4000 → 8000 at no wall-clock cost — shipped in orchestration PR opentargets/orchestration#209.


Daniel-Considine (approved):
Looks all good to me, great work 👍

PR #1233, opened 2026-06-14
feat!: introduce scoped dependencies

### LINKED ISSUE #4032: Gentropy scoped dependencies
As a **gentropy user** I want **Introduce scoped dependencies** because **it can help to reduce the size of the package dependencies**

## Background

Currently gentropy [v2.4.1] contains following list of dependencies:
```
 "pyspark[pandas_on_spark] (>=3.5.0, <3.6)",
  "hail (>=0.2.133, <0.3.0)",
  "scipy (>=1.11.4, <1.16.0)",
  "hydra-core (>=1.3.2, <1.4.0)",
  "pyliftover (>=0.4.1, <0.5.0)",
  "numpy (>=1.26.4, <2.3.0)",
  "wandb (>=0.19.4, <0.22.0)",
  "omegaconf (>=2.3.0, <2.4.0)",
  "typing-extensions (>=4.12.2, <4.15.0)",
  "scikit-learn (>=1.6.1, <1.8.0)",
  "pandas[gcp,parquet] (>=2.2.3, <2.4.0)",
  "skops (>=0.11.0, <0.14.0)",
  "shap (>=0.46, <0.49)",
  "matplotlib (>=3.10.0, <3.11.0)",
  "google-cloud-secret-manager (>=2.12.6, <2.25.0)",
  "google-cloud-storage (>=2.14.0, <3.4.0)",
  "xgboost-cpu>=3.0.4 ; platform_machine == 'amd64' or platform_machine == 'x86_64' ",
  "xgboost>=3.0.4 ; platform_machine == 'aarch64' or platform_machine == 'arm64'",
```

Not all of the steps that we run, use all of the dependencies. Good examples are:
- L2G steps are the only ones to use `xgboost`, `wandb`, `scikit-learn`, `skops`, `shap`, `google-cloud-*`
- finngen finemapping ingestion, gnomad ingestion and ld matrix interface uses `hail` 

We could introduce the dependency scopes and provide guards though the codebase to ensure the use of the additional features (like L2G) is only available once the `gentropy[scope]` is installed. 

Two candidates for scopes are:

- `gentropy[l2g]` including `xgboost`, `wandb`, `scikit-learn`, `skops`, `shap`, `google-cloud-*` and allowing to run all L2G based steps
- `gentropy[hail] including `hail` and allowing to run `FinngenFinemappingIngestionStep`, `GnomadVariantIndexStep`, `SusieFinemmapperStep`

## Tasks

- [ ] Create dependency scopes
- [ ] Provide guards that ensure the scopes are installed
- [ ] Provide tests to ensure guards are working

project-defiant:
## Rationale

Full gentropy stack of dependencies is not needed to run most of the steps except L2G. The same idea is behind fine-mapping & LD access that requires hail. Current gentropy build on the cluster takes ~5minutes due to excessive amount of dependencies. 

This PR attempts to isolate steps that require L2G dependencies and Hail dependencies into 2 separate installation scopes. 

The benefit of this addition should be in simplified dependency resolution (hence startup time) for variant, coloc, credible_set cluster, while retaining L2G dependencies for L2G steps on a separate cluster.

 ## Summary

  Carves gentropy's heavy ML and hail stacks behind opt-in extras. After this PR `pip install gentropy` ships a Spark-only core; users opt into `gentropy[hail]`, `gentropy[l2g]`,
  or `gentropy[all]` based on what steps they run. Docker, Dataproc, and CI continue to install the fat image (`[all]`), so production behaviour is unchanged.

  **BREAKING CHANGE for end users:** plain `pip install gentropy` no longer pulls `hail`, `xgboost`, `scikit-learn`, `skops`, `shap`, `matplotlib`, `wandb`, or `huggingface-hub`.
  A step that needs one of those raises `ImportError` with the exact pip command to fix it.

  ## Install matrix (now in README + docs/installation.md)

  | Use case | Command |
  | --- | --- |
  | Spark-only steps | `pip install gentropy` |
  | + hail-backed datasources (gnomAD LD, FinnGen finemapping, PanUKBB LD, susie) | `pip install gentropy[hail]` |
  | + L2G training, prediction, HF Hub publishing | `pip install gentropy[l2g]` |
  | Full pipeline | `pip install gentropy[all]` |

  ## What changed

  **Packaging (`pyproject.toml`):**
  - Move `hail`, `xgboost{,-cpu}`, `scikit-learn`, `skops`, `shap`, `matplotlib`, `wandb`, `huggingface-hub` from `[dependencies]` to `[project.optional-dependencies]` under
  - `[dependency-groups].test` self-references `gentropy[all]` so the test env stays fat.
  - Register `hail` and `l2g` pytest markers; partition by them.

  **Source-side guards:**
  - New shared helper `src/gentropy/common/imports.py` with `install_hint(extra)` and `optional_imports(extra)` context manager — every guard funnels missing-import errors through
   one voice (`"pip install gentropy[<extra>]"`).
  - **Module-level guards** at the top of intrinsically extra-bound files: `datasource/{gnomad/ld,gnomad/variants,finngen/finemapping,pan_ukbb_ld/ld}.py` (hail) and
  `method/l2g/{model,trainer}.py`, `dataset/l2g_prediction.py` (l2g).
  - **Lazy in-function guards** where heavy imports were already deferred: `common/session.py`, `common/genomic_region.py`, plus `gentropy.l2g` (top-level step module). The four
  non-training L2G step classes — `LocusToGeneFeatureMatrixStep`, `LocusToGeneTrainTestSplitStep`, `LocusToGeneEvidenceStep`, `LocusToGeneAssociationsStep` — remain importable and
   runnable on a core-only install; only `LocusToGeneStep.run_train`/`run_predict` pull the heavy stack.
  - `SessionConfig.hail_home` flips from eagerly-resolved `os.path.dirname(hail.__file__)` to `str | None = None`. `Session._setup_hail_config` already lazy-resolved it, so the
  eager default was just leaking `import hail` into the CLI boot path.
  - `utils/spark.py::get_spark_testing_conf` gains a `with_hail: bool = False` flag; the testing Spark conf is hail-free by default and only mixes in the Kryo registrator / hail
  jar paths when explicitly opted in.

  **Test infrastructure:**
  - `conftest.py` `spark` fixture introspects collected items: builds a hail-enabled session iff at least one collected test carries the `hail` marker. Keeps `pytest -m "not
  hail"` honestly hail-free.
  - Existing hail- and L2G-using tests gain file-scope `pytestmark = pytest.mark.hail` / `.l2g` and `pytest.importorskip` so they're safe to collect on a core-only install.
  - New `tests/gentropy/no_spark/test_hail_optional.py` and `test_l2g_optional.py` pin the public contract — install hint wording, the specific pip command substring, that the
  lazy step module imports cleanly with the L2G stack stubbed out via a `sys.meta_path` finder. New `test_config.py` cases pin `hail_home` default + import-without-hail.

  **Build & CI:**
  - `Dockerfile`: both `uv sync` invocations gain `--all-extras` so the production image still installs everything.
  - `utils/install_dependencies_on_cluster.sh`: Dataproc init switches from `gentropy[hail]` to `gentropy[all]`.
  - `.github/workflows/pr.yaml`: install step now `uv sync --all-groups --all-extras`.
  - `Makefile`: `make test` partitions runs into `hail` vs non-`hail` and `no_shared_spark` vs shared — required because the `spark` fixture needs to know the partition's
  hail-marker state at fixture-build time. Also (small fix) the `uv sync` step uses `--all-groups` so the docs test can collect (`mkdocs` lives in `[dependency-groups].docs`).
  - `tool.deptry`: `xgboost-cpu` keeps the `DEP002` ignore (no module of its own — platform-conditional drop-in); the `xgboost` ignore is dropped since it's now correctly declared
   via the extra.

  ## Test plan

  - [x] `make test` passes locally with `--all-extras --all-groups`.
  - [x] `tests/gentropy/no_spark/test_hail_optional.py` simulates a hail-free env via a custom `sys.meta_path` finder and asserts `pip install gentropy[hail]` appears in the
  raised `ImportError`.
  - [x] `tests/gentropy/no_spark/test_l2g_optional.py` does the same for the L2G stack, and additionally asserts `gentropy.l2g` imports cleanly without the heavy deps
  (lazy-imports work).
  - [x] `tests/gentropy/test_config.py::TestSessionConfig` asserts `hail_home` defaults to `None`, is typed `str | None`, and is importable with hail blocked.


src/gentropy/common/imports.py
```diff
@@ -0,0 +1,55 @@
...
+        str: User-facing message suggesting the pip-install command.
+    """
+    return (
+        f"This functionality requires the `{extra}` extra. "
+        f"Install with: `pip install gentropy[{extra}]`"
+    )
+
+
+@contextmanager
+def optional_imports(extra: str) -> Iterator[None]:
```
ireneisdoomed, line 28: Interesting pattern 😵‍💫

src/gentropy/common/session.py
```diff
@@ -310,7 +310,10 @@ def __init__(
             )
             # Initialize Hail if requested
             if start_hail:
-                import hail as hl
+                from gentropy.common.imports import optional_imports
```
ireneisdoomed, line 313: Could we import thin function in `gentropy/common/__init__.py`? 
Looks a bit too repetitive to me. We'd go from `from gentropy.common.imports import optional_imports` to `from gentropy.common import optional_imports`

src/utils/spark.py
```diff
@@ -32,9 +36,26 @@ def get_spark_testing_conf() -> SparkConf:
...
-        .set("spark.jars", f"{hail_home}/backend/hail-all-spark.jar")
-        .set("spark.driver.extraClassPath", f"{hail_home}/backend/hail-all-spark.jar")
-        .set("spark.executor.extraClassPath", "./hail-all-spark.jar")
-        .set("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
-        .set("spark.kryo.registrator", "is.hail.kryo.HailKryoRegistrator")
     )
+
+    if with_hail:
+        try:
+            import hail as hl
```
ireneisdoomed, line 43: Could we use the context manager here?

tests/gentropy/conftest.py
```diff
@@ -33,13 +32,26 @@
...
+    from gentropy.dataset.l2g_prediction import L2GPrediction
+
 
 @pytest.fixture(scope="session")
-def spark() -> Generator[SparkSession, None, None]:
-    """Local spark session for testing purposes."""
+def spark(request: pytest.FixtureRequest) -> Generator[SparkSession, None, None]:
+    """Local spark session for testing purposes.
+
+    Builds the testing Spark conf with hail support iff at least one
```
ireneisdoomed:
```suggestion
    Builds the testing Spark conf with hail support if at least one
```

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -7,19 +7,24 @@
...
 import pyspark.sql.functions as f
-import shap
 from pydantic import SecretStr
 from pyspark.sql import DataFrame
 from pyspark.sql.types import StringType, StructType
 
+from gentropy.common.imports import optional_imports
+
+with optional_imports("l2g"):
+    import shap
```
ireneisdoomed, line 17: This could be moved inside `explain`, it's the only function that uses it.

src/gentropy/dataset/l2g_prediction.py
```diff
@@ -7,19 +7,24 @@
...
+
+    from gentropy.method.l2g.model import LocusToGeneModel
```
ireneisdoomed, line 19: `LocusToGeneModel` is only used for type checking. I think we should be able to move it under a TYPE_CHECKING block

If that's the case, we could import L2GPrediction without the optional dependencies

ireneisdoomed (changes requested):
Thank you for the initiative!!! My summary:
1. The PR creates different submodule for Gentropy: `l2g`, `hail`, `all`. `pip install gentropy` will only install the bare spark modules
2. Every module that imports hail or l2g dependencies adds some guardrails to check whether the dependency is in the environment. It raises a readable message otherwise
3. All tests that use the optional dependencies are skipped if the dependency is not found
4. Dataproc/Docker/CI install `gentropy[all]` so the behaviour in our pipelines shouldn't change

I have one comment that is not a deal-breaker: IMO, L2G prediction should be available in the bare Gentropy. This means that `xgboost` is installed as part of the main package. This is not currently straightforward to accomplish because L2GPrediction is bound to L2GModel, which imports wandb and huggingface. What do you think?

src/gentropy/common/session.py
project-defiant: totally agree!

project-defiant:
@ireneisdoomed, I can see usecases that people actually do not need to run L2G training, but they actually still want to run the feature matrix, predictions and evidence with our model. So maybe the l2g scope is too wide. The only thing that I do not like with this approach is that we still need to install the xgboost with default gentropy, as the package is quiet large, but we might not be able to avoid it.

ireneisdoomed:
> @ireneisdoomed, I can see usecases that people actually do not need to run L2G training, but they actually still want to run the feature matrix, predictions and evidence with our model. So maybe the l2g scope is too wide. The only thing that I do not like with this approach is that we still need to install the xgboost with default gentropy, as the package is quiet large, but we might not be able to avoid it.

Yes. It makes sense to me, but I agree it is a product decision

PR #1238, opened 2026-06-16
fix: handle OmegaConf DictConfig in Session connectors

project-defiant:
## ✨ Context

This is the part of the larger refactor of Finngen Meta Analysis & DeCODE harmonisation efforts. The overarching goal is to unify the harmonisation pipelines and include the 2-way meta analysis (subsequent PRs).

For deCODE genetics harmonisation source is s3 compatible storage
For finngen-meta the harmonisation source is their GCS bucket

This PR adds the `list_hadoop_paths` method, that uses pure Hadoop java connector from python to list files. This is crucial, so we do not need to include or keep vendor specific packages like (google cloud storage or boto3) to the dependency stack. This method is used to list the existing summary statistics files in all harmonisation pipelines that are part of this larger effort. 

Also testing the addition of hadoop-connector jars to the gentropy I have found issues with connector for GCS, that is too lightweight and does not ship with all requrired packages, it is changed, so we can use the shaded jar with all dependencies included. See links in the code for more details. 

- Coerce OmegaConf DictConfig to plain dict before passing to S3Config/GCSConfig so Hydra-managed configs work end-to-end without manual conversion
- Switch GCS connector from Maven package resolution (gcs-connector:4.0.4 via spark.jars.packages) to a direct shaded JAR URL (v3.1.17) to eliminate Ivy dependency conflicts at runtime
- Add _append_repository() static helper to Session for future Maven repo wiring
- Add list_hadoop_paths() to Session for driver-side glob listing over any Hadoop-compatible filesystem (GCS, S3A, local) using the already-configured connector
- Fix GCSConfig.from_env() to explicitly cast env var strings through GCSAuthType and GCSRequesterPaysMode enums rather than passing raw strings
- Widen s3_configuration / gcs_configuration type hints from dict[str, str] to dict[str, Any] to allow nested config values
- Add tests for list_hadoop_paths() using local tmp_path filesystem

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/common/session.py
+++ b/src/gentropy/common/session.py
@@ -176,6 +176,10 @@ class Session:
         for bucket-level operations or as a requester-pays billing project fallback.
         See :class:`~gentropy.external.gcs.GCSConfig` for the full set of options.
 
+        To allow running locally with GCS credentials, one can set the ``auth_type``
+        to ``APPLICATION_DEFAULT`` and ensure that ``gcloud auth application-default login``
+        has been run to populate the local ADC credentials file.
+
         >>> session = Session(
         ...     add_gcs_connector=True,
         ...     gcs_configuration={'project_id': 'my-gcp-project'},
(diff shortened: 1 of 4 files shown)


xyg123 (approved):
Lovely PR to allow the session to work with S3Config and GCSConfig, the new method to list hadoop files is well covered by the tmp_path tests, looks great!

PR #1240, opened 2026-06-18
feat(study_index): gc lambda distinction

project-defiant:

To reduce the ambiguity of the sumstatQC field names, this PR introduces mapping between sumstatQCField names. 
  collect_heritability in StudyIndex previously stored LDSC metrics in sumstatQCValues using their raw column names as QCCheckName values (e.g. h2, lambda_gc).

  Why it was needed

  There were two sources of ambiguity:

  1. gc_lambda vs lambda_gc: The LDSC output column is named lambda_gc, but the QC threshold parameter in annotate_sumstats_qc is called gc_lambda (e.g. threshold_min_gc_lambda). **Without namespacing, a consumer reading QCCheckName == "lambda_gc" from sumstatQCValues had no
   obvious link to the gc_lambda threshold that flagged the study — the two names looked like different things.**
  2. No namespace: Raw field names like h2 or intercept are generic and give no indication they originate from LDSC. The LDSC_ prefix makes the provenance explicit, which matters when sumstatQCValues can contain metrics from multiple QC sources concatenated together. In future we may want other types of h2 estimates, this ensures it's easy to allow that. (popcorn, sumHer, ...)

## 🚦 Before submitting

### DIFF
--- a/src/gentropy/dataset/study_index.py
+++ b/src/gentropy/dataset/study_index.py
@@ -498,7 +498,9 @@ class StudyIndex(Dataset):
         return StudyIndex(_df=validated_df, _schema=StudyIndex.get_schema())
 
     @qc_test
-    def collect_heritability(self: StudyIndex, heritability_df: DataFrame) -> StudyIndex:
+    def collect_heritability(
+        self: StudyIndex, heritability_df: DataFrame
+    ) -> StudyIndex:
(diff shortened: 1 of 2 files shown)


xyg123 (approved):
Looks great! Makes sense to have explicit names for these.  Thank you!

PR #1272, opened 2026-07-23
feat(l2g): add EGL and training set generation steps

addramir:
## Context

Closes the automation gap described in opentargets/issues#3783 — the L2G gold standard was being generated outside gentropy from ad-hoc notebooks (`EGL_and_training_set/2606/01_EGL_preparation.ipynb` and `02_training_set.ipynb`). This PR brings that logic into the pipeline as two parametrised steps.

### `EffectorGeneListStep` (`gentropy.effector_gene_list`)
Builds the Effector Gene List (EGL): a deduplicated set of `(diseaseId, targetId)` pairs. It combines up to three **optional** platform-ETL sources, each with its own **parametrised filter**:

| Source | Input | Filter param |
|--------|-------|--------------|
| Rare-variant evidence | `rare_variant_evidence_paths` (list, unioned) | `rare_variant_score_threshold` (0.75) |
| ChEMBL clinical precedence | `clinical_evidence_path` | `approved_clinical_phases` (PHASE_4/APPROVAL/PHASE_3/PREAPPROVAL) |
| Legacy OTG gold standard | `gold_standard_path` (JSON) | `gold_standard_confidence` (High/Medium) |

Output: one clean parquet with distinct `diseaseId`, `targetId`.

### `TrainingSetStep` (`gentropy.training_set`)
Labels the L2G feature matrix against the EGL (`positive`/`negative`) and applies a chain of **toggleable, parametrised** cleaning filters mirroring notebook 02:

- replication (`apply_replication_filter`, default **on**)
- max positives per locus (`max_gsp_per_locus=2`)
- protein-protein interaction / STRING (`apply_interaction_filter`, `interaction_source`, `interaction_score_threshold`)
- distance-leakage removal (`apply_distance_filter`)
- protein-coding restriction (`protein_coding_only`)
- deduplication of identical positive feature profiles (`apply_deduplication`)

Output: clean JSON matching the notebook example (`studyLocusId`, `geneId`, `diseaseIds`, `variantId`, `studyId`, `goldStandardSet`).

## Also

- Config dataclasses `EffectorGeneListConfig` / `TrainingSetConfig` registered as `effector_gene_list` and `training_set` steps.
- Docs pages under `docs/python_api/steps/`.
- Unit tests for both steps (EGL end-to-end; training-set filter logic + validation).

## Notes for review
- The replication filter reads the `REPLICATED` quality control flag raised by `StudyLocus.qc_replication` (#1287) rather than assessing replication itself, so the step takes the validated credible sets as input. The `min_replication_studies` parameter is gone with it — the threshold belongs to the check.
- `apply_interaction_filter` defaults to off, since it requires an `interaction_path`.
- Output format matches the notebook example rather than the `L2GGoldStandard` schema (kept `diseaseIds` array), per request. It does not yet plug directly into `LocusToGeneTrainTestSplitStep`; happy to add a schema-conformant variant if wanted.


src/gentropy/training_set.py
```diff
@@ -0,0 +1,427 @@
...
+        feature_matrix_path: str,
+        credible_set_path: str,
+        study_index_path: str,
+        effector_gene_list_path: str,
+        training_set_path: str,
+        interaction_path: str | None = None,
+        apply_replication_filter: bool = True,
+        min_replication_studies: int = 2,
+        max_gsp_per_locus: int = 2,
+        apply_interaction_filter: bool = True,
```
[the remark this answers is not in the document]
addramir: Defaults to False now.

src/gentropy/config.py
```diff
@@ -891,6 +891,46 @@ class FineMappingPlanGeneratorConfig(StepConfig):
...
+    feature_matrix_path: str = MISSING
+    credible_set_path: str = MISSING
+    study_index_path: str = MISSING
+    effector_gene_list_path: str = MISSING
+    training_set_path: str = MISSING
+    interaction_path: str | None = None
+    apply_replication_filter: bool = True
+    min_replication_studies: int = 2
+    max_gsp_per_locus: int = 2
+    apply_interaction_filter: bool = True
```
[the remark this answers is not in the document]
addramir: Done.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,427 @@
...
+        )
+        return (
+            interactions.filter(
+                (f.col("sourceDatabase") == interaction_source)
+                & (f.col("scoring") >= interaction_score_threshold)
+                & (f.col("targetA") != f.col("targetB"))
+            )
+            .select("targetA", "targetB")
+            .distinct()
+        )
```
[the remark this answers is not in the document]
addramir: The pairs are symmetrised in `_interaction_pairs` now, as `remove_false_negatives` does.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,427 @@
...
+                interaction_source,
+                interaction_score_threshold,
+            )
+            labelled = self._filter_interacting_negatives(labelled, interactions)
+
+        # 5. Optional distance-leakage filter.
+        if apply_distance_filter:
+            labelled = labelled.filter(
+                ~((f.col("GSP") == 1) & (f.col("distanceSentinelFootprint") == 0))
+            )
```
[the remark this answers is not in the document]
addramir: The cap is re-applied after the distance and protein-coding filters now, so a locus left without a positive is dropped.

tests/gentropy/step/test_training_set_step.py
```diff
@@ -0,0 +1,199 @@
...
+            [
+                ("sl1", "ENSG_POS", 1),
+                ("sl1", "ENSG_PARTNER", 0),  # interacts with ENSG_POS -> removed
+                ("sl1", "ENSG_OTHER", 0),  # no interaction -> kept
+            ],
+            ["studyLocusId", "geneId", "GSP"],
+        )
+        interactions = spark.createDataFrame(
+            [("ENSG_POS", "ENSG_PARTNER")], ["targetA", "targetB"]
+        )
```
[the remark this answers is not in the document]
addramir: Added, parameterised over both directions, and it now goes through `_interaction_pairs` so it covers the symmetrisation.

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,212 @@
...
+            else list(DEFAULT_GOLD_STANDARD_CONFIDENCE)
+        )
+
+        sources: list[DataFrame] = []
+
+        if rare_variant_evidence_paths:
+            rare = self._rare_variant_pairs(
+                session, rare_variant_evidence_paths, rare_variant_score_threshold
+            )
+            logger.info("Rare-variant source contributes %d pairs", rare.count())
```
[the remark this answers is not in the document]
addramir: Counts dropped; the log lines only say which sources are included.

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+                if evidence is None
+                else evidence.unionByName(df, allowMissingColumns=True)
+            )
+        assert evidence is not None  # noqa: S101 - guaranteed by non-empty evidence_paths
+        return (
+            evidence.filter(f.col("score") >= score_threshold)
+            .select("diseaseId", "targetId")
+            .distinct()
+        )
+
```
project-defiant: Could we make the input contract and aggregation shape explicit here? Define a  `evidence_schema` containing only required columns, then pass it to session.load_data. 

Filter and select each source before unioning them with reduce, and retain one final distinct(). This prevents from reading more then is needed (should give better performance)

Also you do not do the union on all columns, which may break if the evidence files have different schemas.

```suggestion
        evidence_schema = t.StructType(
            [
                t.StructField("diseaseId", t.StringType(), True),
                t.StructField("targetId", t.StringType(), True),
                t.StructField("score", t.DoubleType(), True),
            ]
        )
        def read_rv_evidence(path: str) -> DataFrame:
            return (
                session.load_data(path, "parquet", schema=evidence_schema)
                .filter(f.col("score") >= score_threshold)
                .select("diseaseId", "targetId")
            )
        
        return reduce(
            lambda left, path: left.unionByName(read_rv_evidence(path)),
            evidence_paths[1:],
            read_rv_evidence(evidence_paths[0]),
        ).distinct()
```

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+        Returns:
+            DataFrame: Distinct ``diseaseId``, ``targetId`` pairs.
+        """
+        clinical = session.load_data(clinical_evidence_path, "parquet")
+        return (
+            clinical.filter(f.col("clinicalStage").isin(approved_clinical_phases))
+            .select("diseaseId", "targetId")
+            .distinct()
+        )
```
project-defiant: The same, no need to query entire parquet if we use only 3 columns

```suggestion
        evidence_schema = t.StructType(
            [
                t.StructField("diseaseId", t.StringType(), True),
                t.StructField("targetId", t.StringType(), True),
                t.StructField("clinicalStage", t.StringType(), True),
            ]
        )
        
        clinical = session.load_data(
            clinical_evidence_path,
            "parquet",
            schema=evidence_schema,
        )
```

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+        Args:
+            session (Session): Active session.
+            gold_standard_path (str): Path to the OTG gold standard curation JSON.
+            gold_standard_confidence (list[str]): Confidence levels to keep.
+
+        Returns:
+            DataFrame: Distinct ``diseaseId``, ``targetId`` pairs. The ``diseaseId`` is one
+                EFO term per row (the ontology array is exploded).
+        """
+        gold = session.load_data(gold_standard_path, "json")
```
project-defiant: Same as above, you could limit to a subset of columns

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+# Clinical stages considered "approved enough" to seed a gene-disease pair.
+DEFAULT_APPROVED_CLINICAL_PHASES: list[str] = [
+    "PHASE_4",
+    "APPROVAL",
+    "PHASE_3",
+    "PREAPPROVAL",
+]
+
+# Confidence levels retained from the legacy OTG gold standard curation.
+DEFAULT_GOLD_STANDARD_CONFIDENCE: list[str] = ["High", "Medium"]
```
project-defiant: I would make them StrEnum instead of fields and then reuse the values by name, this will prevent reusing the same text multiple times.

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+from __future__ import annotations
+
+import logging
+
+import pyspark.sql.functions as f
+from pyspark.sql import DataFrame
+
+from gentropy.common.session import Session
+
+logger = logging.getLogger(__name__)
```
project-defiant: If you are going to use the logger, please use the one from the Session()

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+        if not sources:
+            raise ValueError(
+                "At least one EGL source must be provided "
+                "(rare_variant_evidence_paths, clinical_evidence_path or gold_standard_path)."
+            )
+
+        combined = sources[0]
+        for extra in sources[1:]:
+            combined = combined.unionByName(extra)
+        combined = combined.dropDuplicates(["targetId", "diseaseId"])
```
project-defiant: can be a one liner with a clearer msg
```
combined = reduce(...).dropDuplicates(...)
```

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+                    session, gold_standard_path, gold_standard_confidence
+                )
+            )
+
+        if not sources:
+            raise ValueError(
+                "At least one EGL source must be provided "
+                "(rare_variant_evidence_paths, clinical_evidence_path or gold_standard_path)."
+            )
+
```
project-defiant: Is it good to be optional in the evideces? 
Since we do not preserve the source of the evidence, I would make these not optional, as if someone generates the evidence from a single source != to the evidence from multiple sources.

src/gentropy/config.py
```diff
@@ -888,6 +888,45 @@ class FineMappingPlanGeneratorConfig(StepConfig):
...
+@dataclass
+class EffectorGeneListConfig(StepConfig):
+    """Effector Gene List (EGL) generation step configuration."""
+
+    effector_gene_list_path: str = MISSING
+    rare_variant_evidence_paths: list[str] | None = None
+    rare_variant_score_threshold: float = 0.75
+    clinical_evidence_path: str | None = None
+    approved_clinical_phases: list[str] = field(
+        default_factory=lambda: ["PHASE_4", "APPROVAL", "PHASE_3", "PREAPPROVAL"]
```
project-defiant: Is `PREAPPROVAL` a good phase to use?

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,215 @@
...
+        combined = sources[0]
+        for extra in sources[1:]:
+            combined = combined.unionByName(extra)
+        combined = combined.dropDuplicates(["targetId", "diseaseId"])
+
+        (
+            combined.coalesce(session.output_partitions)
+            .write.mode(session.write_mode)
+            .parquet(effector_gene_list_path)
+        )
```
project-defiant: This is a solid case where explicit dataset with a schema should be used.

src/gentropy/config.py
```diff
@@ -888,6 +888,45 @@ class FineMappingPlanGeneratorConfig(StepConfig):
...
+    )
+    gold_standard_path: str | None = None
+    gold_standard_confidence: list[str] = field(
+        default_factory=lambda: ["High", "Medium"]
+    )
+    _target_: str = "gentropy.effector_gene_list.EffectorGeneListStep"
+
+
+@dataclass
+class TrainingSetConfig(StepConfig):
```
project-defiant: The naming is very ambiguous. You already introduced a train_test_split step which has a different meaning. I would argue we first need to fix the terminology like:

1) what is gold standard
2) what is the dataset which you generate here
3) what are the train & test splits 

Without that it's hard to understand the overall flow of the steps.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
+"""Step to build the L2G training set (gold standard) from the Effector Gene List.
```
project-defiant: Is this a gold standard if you use the effector gene list that is build upon a gold standard already?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+
+The training set labels every ``(studyLocusId, geneId)`` row of the L2G feature matrix
+as ``positive`` or ``negative`` (``goldStandardSet``). Positives are the gene-disease
```
project-defiant: every?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+* protein-coding filter — restrict the set to protein-coding genes;
+* deduplication — collapse credible sets that share identical positive feature profiles.
+
+The output is written as JSON with the columns ``studyLocusId``, ``geneId``,
+``diseaseIds``, ``variantId``, ``studyId`` and ``goldStandardSet``.
+"""
+
+from __future__ import annotations
+
+import logging
```
project-defiant: Use session logger

src/gentropy/training_set.py
project-defiant: There is a lot of code in this step, probably it's a good excuse to figure out if we can reuse existing functionalities or enhance existing modules. Definitely move the core functionality into a dataset with a solid name for it (not ambiguous to gold standard)

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            apply_distance_filter (bool): Drop positives with a sentinel footprint distance of zero
+                (the closest gene) to avoid distance leakage. Defaults to True.
+            protein_coding_only (bool): Restrict the training set to protein-coding genes. Defaults to True.
+            apply_deduplication (bool): Collapse credible sets sharing identical positive feature
+                profiles. Defaults to True.
+
+        Raises:
+            ValueError: If the interaction filter is requested without an ``interaction_path``.
+        """
+        if apply_interaction_filter and not interaction_path:
```
project-defiant: Presence of the file can be a switch already, no need for the explicit flag.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+                "interaction_path is required when apply_interaction_filter is True."
+            )
+
+        feature_matrix = session.load_data(feature_matrix_path, "parquet")
+        credible_set = StudyLocus.from_parquet(
+            session, credible_set_path, recursiveFileLookup=True
+        )
+        study_index = StudyIndex.from_parquet(
+            session, study_index_path, recursiveFileLookup=True
+        )
```
project-defiant: why recursive lookup? This should work on the post validation `credible_set` and `study`, the dataset is flat then.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        effector_gene_list = session.load_data(effector_gene_list_path, "parquet")
```
project-defiant: This should be a dataset with a schema

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # 1. Attach study and disease context, then label positives against the EGL.
+        annotated_fm = self._annotate_feature_matrix(
+            feature_matrix, credible_set, study_index
+        )
```
project-defiant: i think this functionality already exists in the gentropy - annotating credible sets or gold standard. Please verify if it can be reused and tweak it to meet your expectations. Do not duplicate the code.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        labelled = self._label_gold_standard(annotated_fm, effector_gene_list)
```
project-defiant: As above.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # 2. Optional replication filter.
+        if apply_replication_filter:
+            labelled = labelled.join(
+                self._replicated_loci(credible_set), on="studyLocusId", how="inner"
+            )
```
project-defiant: semi join

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        Returns:
+            DataFrame: Rows of loci containing a positive, with a ``GSP`` column (1 positive, 0 negative).
+        """
+        egl = effector_gene_list.select("targetId", "diseaseId").withColumnRenamed(
+            "targetId", "geneId_egl"
+        )
+        positives = (
+            annotated_fm.join(
+                egl,
+                (f.array_contains(annotated_fm["diseaseIds"], egl["diseaseId"]))
```
project-defiant: will explode when more then one diseaseId will map to the list. Is that expected?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+                & (annotated_fm["geneId"] == egl["geneId_egl"]),
+                how="inner",
```
project-defiant: I am very curious what is the space that we miss in the EGL join

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        loci_with_positive = positives.select("studyLocusId").distinct()
+        return (
+            annotated_fm.join(loci_with_positive, on="studyLocusId", how="inner")
+            .join(
+                positives.withColumn("GSP", f.lit(1)),
+                on=["studyLocusId", "geneId"],
+                how="left",
+            )
+            .withColumn("GSP", f.when(f.col("GSP").isNotNull(), 1).otherwise(0))
+        )
```
project-defiant: So the GSP == 1 when the locus has evidence in the EGL list
and GSP == 0 when the GSP locus has some other (gene-disease) association, which is not found in the EGL. This is reflected by:

$$\mathrm{GSP}^{+}(l,g)=\mathbf{1}\left[\exists(d,g_{\mathrm{EGL}})\in\mathrm{EGL}:d\in D_l\;\land\;g_{\mathrm{EGL}}=g\right]$$


This also means that for a
* locus1, geneA, diseaseA (geneA, diseaseA) in EGL -> positive
* locus1 geneB, diseaseA (geneB, diseaseA) not in EGL -> negative due to a missing link to a disease in the EGL (but might be false negative if not tested)


* locus1, geneB, diseaseB (geneB, diseaseB) not in EGL -> negative due to no such evidence (unknown, might be false negative)

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            credible_set (StudyLocus): Validated credible set dataset.
+
+        Returns:
+            DataFrame: The ``studyLocusId`` values carrying the ``REPLICATED`` flag.
+        """
+        return credible_set.df.filter(
+            f.array_contains(
+                f.col("qualityControls"), StudyLocusQualityCheck.REPLICATED.value
+            )
+        ).select("studyLocusId")
```
project-defiant: This belongs to `StudyLocus` as a method.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        """
+        counts = (
+            labelled.filter(f.col("GSP") == 1)
+            .groupBy("studyLocusId")
+            .agg(f.count("*").alias("count"))
+        )
+        keep = counts.filter(
+            (f.col("count") > 0) & (f.col("count") <= max_gsp_per_locus)
+        ).select("studyLocusId")
+        return labelled.join(keep, on="studyLocusId", how="inner")
```
project-defiant: What is the rationale ? How would you calibrate this ? How big the outliers can be?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            session (Session): Active session.
+            interaction_path (str): Path to the interaction dataset.
+            interaction_source (str): ``sourceDatabase`` to keep.
+            interaction_score_threshold (float): Minimum interaction ``scoring``.
+
+        Returns:
+            DataFrame: Distinct ``targetA``, ``targetB`` interaction pairs, each pair present in
+                both directions (self-interactions removed).
+        """
+        interactions = session.load_data(
```
project-defiant: again read only required columns

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        # 3. Cap the number of positives per locus.
+        labelled = self._cap_positives_per_locus(labelled, max_gsp_per_locus)
+
+        # 4. Optional protein-protein interaction filter.
+        if apply_interaction_filter:
+            assert interaction_path is not None  # noqa: S101 - guaranteed by the guard above
+            interactions = self._interaction_pairs(
+                session,
+                interaction_path,
+                interaction_source,
```
project-defiant: Only one interaction source?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        )
+        return pairs.unionByName(
+            pairs.select(
+                f.col("targetB").alias("targetA"), f.col("targetA").alias("targetB")
+            )
+        ).distinct()
+
+    @staticmethod
+    def _filter_interacting_negatives(
+        labelled: DataFrame, interactions: DataFrame
```
project-defiant: The interactions could be a separate dataset with a `Interactions.to_lut()` method that transforms it to the lookup table.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            labelled (DataFrame): Labelled feature matrix with a ``GSP`` column.
+            interactions (DataFrame): Symmetrised ``targetA``, ``targetB`` interaction pairs.
+
+        Returns:
+            DataFrame: Labelled rows with the interacting negatives removed.
+        """
+        positive_partners = (
+            labelled.filter(f.col("GSP") == 1)
+            .select("geneId", "studyLocusId")
+            .join(interactions, f.col("geneId") == interactions["targetA"], how="inner")
```
project-defiant: semi join

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+        # A negative gene that also appears as a positive's interaction partner in the same
+        # locus shows up (at least) twice in the union: flag those for removal.
+        to_remove = (
+            negatives.union(positive_partners)
+            .groupBy("geneId", "studyLocusId")
+            .agg(f.count("*").alias("count"))
+            .filter(f.col("count") >= 2)
+            .select("geneId", "studyLocusId")
+        )
+        return labelled.join(to_remove, on=["geneId", "studyLocusId"], how="anti")
```
project-defiant: Would can be done more explicitly
1. interactions_intesect = interactions.groupBy('targetA').collect_set('targetB')
2. `negatives = labelled.where('GSP == 0').select('studyLocusId', geneId').groupBy('studyLocusId').agg(f.collect_set('geneId').alias("gsn_genes")`
3. `labelled.filter("GSP == 0").semi_join(negatives, on='studyLocusId').join(interactions_intersect, on=geneId = targetA & array_overlap(targetBs, gsn_genes)`

This is just pseudocode. 


The more important question - why do we think that these are not gold standard negatives?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            )
+
+        # 6. Optional protein-coding restriction.
+        if protein_coding_only:
+            labelled = labelled.filter(f.col("isProteinCoding") == 1)
+
+        # Both filters above can drop every positive of a locus and leave it all-negative:
+        # re-applying the cap drops the loci that no longer hold a positive.
+        if apply_distance_filter or protein_coding_only:
+            labelled = self._cap_positives_per_locus(labelled, max_gsp_per_locus)
```
project-defiant: why not doing that cap filtering at the end (just once) ??

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+logger = logging.getLogger(__name__)
+
+# Feature-matrix columns used, together with the rounded colocalisation columns, as the
+# key for deduplicating credible sets with identical positive feature profiles.
+_DEDUP_KEY_COLUMNS: list[str] = [
+    "geneId",
+    "diseaseIds",
+    "variantId",
+    "vepMaximum",
+    "vepMean",
```
project-defiant: why not rounded ?

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+
+        Returns:
+            DataFrame: Labelled rows restricted to the deduplicated credible sets.
+        """
+        positives = labelled.filter(f.col("GSP") == 1)
+        for col in _DEDUP_ROUNDED_COLUMNS:
+            positives = positives.withColumn(col, f.round(f.col(col), 2))
+        deduped = positives.dropDuplicates(_DEDUP_KEY_COLUMNS + _DEDUP_ROUNDED_COLUMNS)
+        loci_to_keep = deduped.select("studyLocusId").distinct()
+        return labelled.join(loci_to_keep, on="studyLocusId", how="inner")
```
project-defiant: It looks highly impossible to me to have the dupllicates via so many columns). Especially if you use the replicated loci, which should already be deduplicated signals.

src/gentropy/training_set.py
```diff
@@ -0,0 +1,413 @@
...
+            "studyId",
+            f.when(f.col("GSP") == 1, f.lit("positive"))
+            .otherwise(f.lit("negative"))
+            .alias("goldStandardSet"),
+        )
+        (
+            training_set.coalesce(session.output_partitions)
+            .write.mode(session.write_mode)
+            .json(training_set_path)
+        )
```
project-defiant: Please do explicit schema & parquet format.

project-defiant (changes requested):
Overall a good and solid proposal for the EGL and dynamic gold standard generation. Please have a look at the comments. The most important I think we need to discuss is the terminology of what is train, test sets, gold standard, egl and so on. We also need a consolidation of this functionality so it can be plugged directly to current L2G.

src/gentropy/training_set.py
addramir: It is actually quite common because there are many CSs for the saem disease but many different GWAS.

src/gentropy/config.py
addramir: Yes, it is still considered as phase 4.

src/gentropy/config.py
addramir: gold standard is a legacy term for old gold standard list. Here we generate the full training set that further is split by train and test,with the option to train the model on the full training set in the end. So i would say terminology here is ocrrect, but we probably need to change it in another step and call it as subtrainng and held-out or something. But is is a spearete question.

src/gentropy/effector_gene_list.py
addramir: I don't see the problem to be flexible here. Also, it allows us to play with different evidence sources.

src/gentropy/effector_gene_list.py
addramir: Each source is now filtered and selected right after the read, then combined with one `reduce` and `distinct()`. Parquet reads only the selected columns, so a read schema adds nothing.

src/gentropy/effector_gene_list.py
addramir: Same: filter and select straight after the read.

src/gentropy/effector_gene_list.py
addramir: Selected right after the read. It's JSON, so it's parsed in full either way.

src/gentropy/effector_gene_list.py
addramir: Made them module-level tuples used as the step defaults. An enum wouldn't remove the copy in `config.py`, which needs the literal values.

src/gentropy/effector_gene_list.py
addramir: Switched to `session.logger`.

src/gentropy/effector_gene_list.py
addramir: Done.

src/gentropy/effector_gene_list.py
addramir: The output is only `diseaseId` and `targetId`, selected explicitly. A dataset class for two columns seemed too much.

src/gentropy/training_set.py
addramir: No, "gold standard" is gone from this step. The output now uses the `L2GGoldStandard` schema, so the split step reads it as is.

src/gentropy/training_set.py
addramir: Only loci with at least one EGL gene. Docstring fixed.

src/gentropy/training_set.py
addramir: Removed, the step doesn't log.

src/gentropy/training_set.py
addramir: Cut down, and it now writes an `L2GGoldStandard` parquet that the split step reads directly. The filters need feature-matrix columns that `L2GGoldStandard` doesn't have, so they stay in the step.

src/gentropy/training_set.py
addramir: Done, `interaction_path` alone switches it on.

src/gentropy/training_set.py
addramir: Removed.

src/gentropy/training_set.py
addramir: Same as the EGL step: two explicit columns.

src/gentropy/training_set.py
addramir: `build_feature_matrix` starts from existing labels; here the labels are what's being built. The annotation is now one join.

src/gentropy/training_set.py
addramir: Same as above.

src/gentropy/training_set.py
addramir: Replication is now a filter on the credible set before the join.

src/gentropy/training_set.py
addramir: Yes, then collapsed back to one row per locus and gene.

src/gentropy/training_set.py
addramir: Only logging counts at each stage would show it, and each count is a full Spark action, so I left it out for now.

src/gentropy/training_set.py
addramir: Intended. The study is a GWAS of disease A too, and geneA is an effector for A. Labels are per locus and gene, not per disease.

src/gentropy/training_set.py
addramir: It's now one inline filter on `qualityControls`.

src/gentropy/training_set.py
addramir: More than two effector genes per credible set is biologically implausible and usually means the labels are off. 2 came out of benchmarking.

src/gentropy/training_set.py
addramir: Selected right after the filter, so only the needed columns are read.

src/gentropy/training_set.py
addramir: STRING only is the protocol's legacy choice. We may add other sources later.

src/gentropy/training_set.py
addramir: It's now a filter and a select, so a dataset seemed too much for now.

src/gentropy/training_set.py
addramir: Rewritten, see the next comment.

src/gentropy/training_set.py
addramir: Rewritten as an anti-join of the positives' partners among the negatives. A gene that interacts strongly with the effector may be causal too, so labelling it negative adds noise, same as `remove_false_negatives`.

src/gentropy/training_set.py
addramir: The first cap counts positives before the filters remove any. The second only drops loci left with no positive.

src/gentropy/training_set.py
addramir: VEP stays exact to match the notebook the current training set came from.

src/gentropy/training_set.py
addramir: Done.

src/gentropy/training_set.py
addramir: Also made dedup deterministic and sorted `diseaseIds` before comparing.

src/gentropy/training_set.py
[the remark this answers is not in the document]
addramir: Reverted to one direction (positive in `targetA`) to match the notebook.

tests/gentropy/step/test_training_set_step.py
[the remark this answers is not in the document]
addramir: The test now checks that only the `targetA` direction removes the negative.

src/gentropy/training_set.py
project-defiant: I will write in one place. I think that passing the arbitrary DataFrame between functions is a bad pattern to follow. This means that whenever function user want to know what can be used in such a function, they have to dig into the code (usually) while defining a custom type (class in python) is usually a good idea, as it introduces the type constraints. Especially if we make it a Dataset. The Dataframe is very loose concept.

src/gentropy/training_set.py
project-defiant: I disagree, we may have only explicit information about geneA - diseaseA link, so even if the GWAS links diseaseA with locus1, that does mean that geneA is linked to both diseases. May be just the locus is not the one that is causal for that particular gene. 

If on the other hand you would find evidence for both diseasesA - geneA - diseaseB links, then I would boost the confidence in that this locus may be a better GSP

src/gentropy/training_set.py
project-defiant: We already use interactions in more then one place. I think this is a valid idea. Besides that it's much cleaner then passing a DataFrame (dynamic concept).

src/gentropy/effector_gene_list.py
project-defiant: I think it introduces a lot of unnecesary code branches. If you are going to use this feature, it can stay.

src/gentropy/effector_gene_list.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Raises:
+            ValueError: If none of the three sources are provided.
+        """
+        sources: list[DataFrame] = []
+
+        if rare_variant_evidence_paths:
+            session.logger.info("Including the rare-variant source")
+            sources += [
+                session.load_data(path, "parquet")
```
project-defiant: Why not reading just required columns? This is a general comment on all evidence including rare variant

src/gentropy/config.py
project-defiant: To be fair, I see no reason for having two steps 
1) build EGL from evidence streams
2) generate training step from EGL and feature matrix
2) split to test & train batch

Why not merging train and split steps? Would reduce complexity.

src/gentropy/training_set.py
addramir: Added `EffectorGeneList` and `Interactions` datasets with schemas; the steps pass those instead of DataFrames.

src/gentropy/training_set.py
addramir: A study's `diseaseIds` map one phenotype, not separate diseases, so a locus for that phenotype is a locus for each of its terms. It's the same reasoning the `REPLICATED` flag uses, and it matches the protocol.

src/gentropy/training_set.py
addramir: Done, `Interactions` dataset with `high_confidence(source, min_score)`.

src/gentropy/effector_gene_list.py
addramir: We use it, so keeping it.

PR #1274, opened 2026-08-17
fix(qc): raise DUPLICATED_STUDYLOCUS_ID and make duplicate flagging deterministic

addramir:
## Problem

`StudyLocus.validate_unique_study_locus_id` exists and is unit-tested, but **it is never called anywhere in the codebase**:

```
$ grep -rn "validate_unique_study_locus_id" src/
src/gentropy/dataset/study_locus.py:388:    def validate_unique_study_locus_id(...)
```

So `DUPLICATED_STUDYLOCUS_ID` is never raised on any dataset. Release configs list it in the `credible_set_validation` step's `invalid_qc_reasons`, which currently has no effect — non-unique `studyLocusId`s pass straight through into `output/credible_set`.

Found while ingesting an external meta-analysis: a delivery whose credible sets contained 212 rows for 145 distinct signals (overlapping ±1 Mb fine-mapping windows re-fine-mapped the same region) produced 212 rows carrying only 144 distinct `studyLocusId`s, and the validation step raised nothing. Deduplication had been deliberately deferred to that step on the assumption the flag was live.

A second, smaller issue: `Dataset.flag_duplicates` ordered its window by an unseeded `rand()`, so which duplicate was retained varied between runs over identical input.

## Changes

**`StudyLocusValidationStep`** now calls `validate_unique_study_locus_id()`. This step is the right place for the check: its input is a union of independently generated credible set datasets (`gcca`, `gcss_pics`, `gcss_susie`, `finngen`, `eqtl_catalogue`, `ukb_ppp_eur`), so cross-source uniqueness can only be established once they are combined. `SUSIE_inf.credible_set_qc` already deduplicates, but only within a single fine-mapping output and by dropping rows rather than flagging them.

**`Dataset.flag_duplicates`** orders by `monotonically_increasing_id()` instead of `rand()`, and takes an optional `order_by` so callers can state which duplicate should win.

## Behaviour

Flag-only, no rows dropped — as with every other check in the step. Datasets that already have unique ids are unaffected. Datasets that do not will see `DUPLICATED_STUDYLOCUS_ID` on all but one row per id, and those rows move to `excluded/credible_set` under the existing release config. That is the intent of the config, but it is a live-output change worth a reviewer's attention.

## Tests

`tests/gentropy/dataset/test_dataset.py`:
- `test_flag_duplicates_is_deterministic` — repeated runs retain the same row
- `test_flag_duplicates_honours_order_by` — caller ordering decides the survivor

The existing `TestStudyLocusDuplicationFlagging` already covers the flagging logic itself.

## Notes for reviewers

- `StudyIndex.validate_unique_study_id` has the same problem — defined and tested, but not called by `StudyValidationStep`. Left out to keep this focused; happy to fold it in.
- `flag_duplicates` retains an arbitrary-but-reproducible row. For credible sets, keeping the highest `credibleSetlog10BF` would be more meaningful and would match `SUSIE_inf.credible_set_qc`; that is a one-line `order_by` at the call site if you want it here.


src/gentropy/dataset/dataset.py
```diff
@@ -422,19 +422,30 @@ def update_quality_flag(
         ).otherwise(qc)
 
     @staticmethod
-    def flag_duplicates(test_column: Column) -> Column:
```
project-defiant, line 425: This is more a comment on the design, rather then on the code.

There can be two duplicate examples:

1) two credible sets arose from the same locus - exactly the same, hence we can just keep one and remove others via flagging (which you refactored)
2) two credible sets are not the same, even though they have the same studyLocusId (we do not use entire locus object as a studyLocusId discriminator, hence we can have a mismatch). This case probably kills the idea of studyLocusId in the first place, as it could become non-unique. 

I am not convinced if we can just deal with 1) but not check 2) here and arbitrary choose one of loci.

src/gentropy/dataset/dataset.py
```diff
@@ -422,19 +422,30 @@ def update_quality_flag(
...
+        `monotonically_increasing_id()`, which preserves the order the rows were read in.
+
         Args:
             test_column (Column): Column to check for duplicates
+            order_by (list[Column] | None): Ordering deciding which duplicate is kept, the first
+                row of each group being the one that is not flagged. Defaults to
+                `monotonically_increasing_id()`.
 
         Returns:
             Column: Column with a boolean flag for duplicates
```
project-defiant, line 446: Hence it is a library function, please add an example usage.

src/gentropy/dataset/dataset.py
```diff
@@ -422,19 +422,30 @@ def update_quality_flag(
...
         """
+        ordering = order_by if order_by else [f.monotonically_increasing_id()]
```
project-defiant, line 466: I am not sure if using the `monotonically_increasing_id` do anything here. The order should be the same (read from top to bottom at random, depending on how the data is shuffled before the partition by test column)

tests/gentropy/dataset/test_dataset.py
```diff
@@ -99,3 +99,53 @@ def test_process_class_params(spark: SparkSession) -> None:
...
+            for row in df.withColumn(
+                "isDuplicate", Dataset.flag_duplicates(f.col("key"))
+            )
+            .filter(~f.col("isDuplicate"))
+            .collect()
+        }
+
+    first = retained()
+    assert len(first) == 2, "one row per key should be retained"
+    assert first == retained() == retained()
```
project-defiant: I am not sure I follow this, is it meant to test the resilience of monotonically_increasing_id with multiple `retained()` calls ? This would make sense for such a small scale dataset on test run at single machine, but will not be guaranteed on the cluster at all.

project-defiant (approved):
In general I agree with the decisions. Please have a look at the comments. No deal breaker.

src/gentropy/dataset/dataset.py
```diff
@@ -422,19 +422,30 @@ def update_quality_flag(
...
+    ) -> Column:
         """Return True for rows, where the value was already seen in column.
 
         This implementation allows keeping the first occurrence of the value.
 
+        The ordering that decides which occurrence is kept must be deterministic, otherwise
+        repeated runs over the same input flag different rows and the resulting dataset is not
+        reproducible. When no ordering is given, rows are ordered by
+        `monotonically_increasing_id()`, which preserves the order the rows were read in.
+
```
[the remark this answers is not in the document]
addramir: Agreed, that claim was wrong. `monotonically_increasing_id()` is a function of the physical row layout (partition index and row index within the partition), and nothing guarantees that layout is stable across runs. Rewritten in ef15d7e to describe the default as arbitrary but not random — it replaces the unseeded `rand()`, no more than that — and to point callers who need a defined survivor at `order_by`.

src/gentropy/dataset/dataset.py
addramir: Added in ef15d7e, showing the `order_by` form with `credibleSetlog10BF` since that is the ordering a credible set caller would actually want.

src/gentropy/dataset/dataset.py
addramir: It does two things, neither of them a determinism guarantee. It gives the window an ordering expression at all, which `row_number()` requires, and it removes the unseeded `rand()`, which drew a fresh random value per run and made the retained row vary over identical input. What it does not do is survive a reshuffle: the ids follow the physical layout, so if the layout changes the survivor changes with it. You are right that the ordering is then effectively "read from top to bottom, subject to how the data is partitioned". The docstring now says exactly that rather than promising reproducibility.

tests/gentropy/dataset/test_dataset.py
addramir: Correct, and the test was overclaiming. Three `retained()` calls on a four-row local DataFrame show that the plan is stable on one machine, which is not the property the name promised and not something that holds on a cluster. Narrowed in ef15d7e to the invariant that does hold — exactly one row per key is left unflagged — and renamed to `test_flag_duplicates_retains_one_row_per_key`. The defined-survivor case stays covered by `test_flag_duplicates_honours_order_by`, which is the path a caller should use when the choice matters.

src/gentropy/dataset/dataset.py
addramir: Fair point. Changed in 3c7f276.

Case 1, two identical credible sets: one is kept, as before.

Case 2, same id but different loci: the choice is no longer arbitrary. The credible set with the fewest variants is kept, as the better resolved one. I also moved the check to after `filter_credible_set`, otherwise `locus` still holds every tagging variant and the size would compare fine-mapping windows instead of credible sets. Credible sets of equal size are still resolved arbitrarily, but then neither is preferable. I did not use `credibleSetlog10BF` for this, since it is null for PICS.

Nothing is deleted either way: flagged rows go to `excluded/credible_set`, so a collision stays visible instead of disappearing.

I agree this does not fix the underlying problem, which is that `studyLocusId` is a hash of study and lead variant only, so it is not guaranteed unique. This PR makes such cases visible rather than silent. Happy to open a separate issue for the id definition if you want it tracked.

src/gentropy/dataset/dataset.py
addramir: Correction to the above: the example now orders by credible set size, not by `credibleSetlog10BF`. That column is null for PICS, so it is not usable here. See 3c7f276.

project-defiant (approved):

PR #1282, opened 2026-08-28
feat(l2g): pathway enrichment features

addramir:
Adds two opt-in L2G features built on a catalogue of pathways enriched among the genetically associated genes of each Open Targets phenotype.

## What the feature is

For a credible set, take every gene whose TSS is within 500 kb of the study locus. Resolve the study to its disease identifiers, collect the pathways enriched for those diseases, and score each gene by

```
pathwayEnrichment500kb = (pathways containing the gene that are enriched) / (pathways containing the gene that were tested)
```

The value is 0 for a gene no enriched pathway contains, 1 for a gene whose every pathway is enriched, and it is normalised so that promiscuous genes are not favoured just for belonging to many gene sets.

- `pathwayEnrichment500kb` — the score above
- `pathwayEnrichment500kbNeighbourhood` — the same score divided by the largest score at the locus

The second one exists because the raw score is dense rather than sparse: for a typical disease around half of the library's genes belong to at least one enriched pathway, so nearly every gene in a window is non-zero and what separates genes is how they compare with their neighbours.

Neither feature is in the default `features_list`. They have to be asked for explicitly, and the step raises if the inputs they need are missing.

## New datasets and step

**`PathwayIndex`** — the gene sets of a pathway library, one row per pathway: `pathway`, `pathwayId`, `source`, `geneSymbols` and `geneIds`. `from_gmt` parses a GMT file and `resolve_gene_ids` maps the symbols onto Ensembl identifiers through `TargetIndex.symbols_lut`, which covers approved and obsolete symbols. On the library used here that resolves 16,733 of 17,246 symbols into 16,775 genes; the remainder are tRNA and immunoglobulin segment names, and 23 of 8,859 pathways end up with no resolvable gene.

`pathwayId` comes from the second GMT field, which in this library is the identifier in the source ontology rather than a description — `GO:0044208` for the 5,954 GO BP sets, `R-HSA-164843` for the 2,855 Reactome ones, and empty for the 50 MSigDB Hallmark ones. `pathway` keeps the name verbatim, including its bracketed source tag, because that string is the only key linking the library to the enrichment results.

**`PathwayEnrichment`** — the enriched pathways of each disease: `diseaseId`, `pathway`, `source`, `normalisedEnrichmentScore`, `pValue` and `pValueAdjusted`. `from_gsea_catalogue` reads the producer's output and renames its columns, projecting away the ones the feature does not use.

**`PathwayIngestionStep`** — turns the GMT file and the raw enrichment results into those two Parquet datasets, once, so the feature matrix step only ever reads harmonised and schema-validated inputs.

## The missing adjusted p-values

`with_recomputed_adjusted_p_value` fills in a null `pValueAdjusted` with a Benjamini-Hochberg step-up value recomputed from the p-values of that disease. This is not cosmetic: in the catalogue this was developed against, 249 of 3,766 diseases have a null value on every single row, and those are the gene-rich ones — body mass index, systemic lupus erythematosus, psoriasis, ulcerative colitis, haematocrit among them. Roughly 15% of GWAS studies point at one. Their p-values are intact, so recomputing recovers them; without it a significance filter deletes them silently. A published value is kept as it is.

Two caveats, both documented on the method:

- The recomputation is adjusted over the rows the dataset holds, and the producer keeps only positively enriched pathways — a median of 4,107 rows per disease against 8,217 pathways actually tested. So the recomputed value is more conservative than the published one and not comparable with it: on diseases that have both, the count of pathways below 0.05 ranges from 1% to 100% of the published count. It exists to rescue the 249 diseases, and should give way to a fixed upstream column.
- Pathways with a non-finite enrichment score come with a p-value of exactly zero, which any recomputation would turn into the most significant result of the disease, so they are left out of the correction by default.

Psoriasis, one of the 249, goes from 0 to 400 enriched pathways once the correction runs.

## Implementation notes

Studies are grouped by their **set** of diseases rather than handled one at a time. A pathway enriched for more than one of a study's diseases has to count once, which cannot be done from per-disease gene counts, and there are far fewer distinct disease sets than studies (94% of GWAS studies carry exactly one disease). That keeps the pathway-level explosion bounded by the number of distinct sets instead of the number of studies.

The denominator is the set of pathways enrichment was tested for, which is the same 8,217 pathways for every disease in this catalogue, so the score is comparable across studies.

## Testing

- Unit tests for GMT parsing, symbol resolution, gene set explosion, the BH correction (values, agreement with R on ties, monotonicity, null p-values, non-finite enrichment scores, and that a published value is left alone) and the enriched-pathway filter.
- Feature tests asserting exact scores for a single-disease study, for a study carrying two diseases where the pathway union matters, and for the neighbourhood normalisation.
- Both feature classes added to the feature factory return-type test.
- Smoke tested end to end against the real catalogue and Open Targets 26.03: scores stay within [0, 1] and the recomputed adjusted p-values within [0, 1].

## For review

The catalogue ranks genes by the Open Targets `genetic_association` datatype score, which is itself built largely on L2G output. A feature derived from it therefore feeds L2G predictions back into L2G input: gold standard performance will overstate what the feature adds, and it couples each model to its predecessor. That is why the features are opt-in here, and it is worth settling before either reaches a released model.

### DIFF
--- a/docs/assets/schemas/pathway_enrichment.md
+++ b/docs/assets/schemas/pathway_enrichment.md
@@ -1,0 +1,11 @@
+```
+root
+ |-- diseaseId: string (nullable = false)
+ |-- pathway: string (nullable = false)
+ |-- source: string (nullable = true)
+ |-- nes: double (nullable = true)
+ |-- pval: double (nullable = true)
+ |-- fdr: double (nullable = true)
+
+
+```
--- a/src/gentropy/assets/schemas/pathway_index.json
+++ b/src/gentropy/assets/schemas/pathway_index.json
@@ -1,0 +1,27 @@
+{
+  "type": "struct",
+  "fields": [
+    {
+      "name": "pathway",
+      "type": "string",
+      "nullable": false,
+      "metadata": {}
+    },
+    {
+      "name": "source",
+      "type": "string",
+      "nullable": true,
+      "metadata": {}
+    },
+    {
+      "name": "geneSymbols",
+      "type": {
+        "type": "array",
+        "elementType": "string",
+        "containsNull": false
+      },
+      "nullable": false,
+      "metadata": {}
+    }
+  ]
+}
--- a/src/gentropy/dataset/l2g_features/pathway.py
+++ b/src/gentropy/dataset/l2g_features/pathway.py
@@ -1,0 +1,270 @@
+"""Methods to generate features based on the pathways enriched for a study's diseases."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import pyspark.sql.functions as f
+from pyspark.sql import Window
+
+from gentropy.common.spark import convert_from_wide_to_long
+from gentropy.dataset.l2g_features.l2g_feature import L2GFeature
+from gentropy.dataset.l2g_gold_standard import L2GGoldStandard
+from gentropy.dataset.pathway_enrichment import PathwayEnrichment
+from gentropy.dataset.pathway_index import PathwayIndex
+from gentropy.dataset.study_index import StudyIndex
+from gentropy.dataset.study_locus import StudyLocus
+from gentropy.dataset.target_index import TargetIndex
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame
+
+
+def common_pathway_enrichment_feature_logic(
+    study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+    feature_name: str,
+    *,
+    pathway_index: PathwayIndex,
+    pathway_enrichment: PathwayEnrichment,
+    study_index: StudyIndex,
+    study_locus: StudyLocus,
+    target_index: TargetIndex,
+    fdr_threshold: float,
+    genomic_window: int,
+) -> DataFrame:
+    """Score every gene at a locus by how much of its pathway membership is disease relevant.
+
+    For a gene the score is the fraction of the pathways it belongs to that are enriched
+    among the genes associated with the diseases of the study behind the credible set. A gene
+    that sits in ten pathways of which three are enriched scores 0.3; a gene in the window
+    that no enriched pathway contains scores 0.
+
+    Pathways are counted once per study even when several of the study's diseases flag the
+    same pathway, and the denominator only counts pathways that enrichment was tested for, so
+    the score always falls between 0 and 1.
+
+    Args:
+        study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci
+            that will be used for annotation
+        feature_name (str): The name of the feature
+        pathway_index (PathwayIndex): Gene set membership of the pathway library
+        pathway_enrichment (PathwayEnrichment): Pathways enriched for each disease
+        study_index (StudyIndex): Study index, used to resolve a study to its diseases
+        study_locus (StudyLocus): Credible sets, used for the position of the study locus
+        target_index (TargetIndex): Target index, used for gene positions and symbol mapping
+        fdr_threshold (float): Maximum FDR for a pathway to count as enriched
+        genomic_window (int): Distance up and downstream of the study locus to collect genes from
+
+    Returns:
+        DataFrame: Feature dataset with one row per study locus and gene in its window
+    """
+    # Gene set membership, restricted to the pathways enrichment was actually tested for and
+    # translated from gene symbols into the gene identifiers the feature matrix is keyed on.
+    membership = (
+        pathway_index.gene_membership()
+        .join(pathway_enrichment.df.select("pathway").distinct(), "pathway", "inner")
+        .join(
+            target_index.symbols_lut().select("geneSymbol", "geneId"),
+            "geneSymbol",
+            "inner",
+        )
+        .select("pathway", "geneId")
+        .distinct()
+    )
+    pathways_per_gene = membership.groupBy("geneId").agg(
+        f.count("pathway").alias("pathwaysPerGene")
+    )
+
+    # Studies are grouped by their set of diseases rather than handled one by one: there are
+    # far fewer distinct disease sets than studies, and the great majority hold one disease.
+    disease_sets = (
+        study_index.df.select(
+            "studyId", f.array_sort(f.col("diseaseIds")).alias("diseaseSet")
+        )
+        .filter(f.size("diseaseSet") > 0)
+        .distinct()
+    )
+    enriched_pathways_per_gene = (
+        disease_sets.select("diseaseSet")
+        .distinct()
+        .select("diseaseSet", f.explode("diseaseSet").alias("diseaseId"))
+        .join(
+            pathway_enrichment.enriched_pathways(fdr_threshold), "diseaseId", "inner"
+        )
+        .select("diseaseSet", "pathway")
+        .distinct()
+        .join(membership, "pathway", "inner")
+        .groupBy("diseaseSet", "geneId")
+        .agg(f.count("pathway").alias("enrichedPathwaysPerGene"))
+    )
+    scores = enriched_pathways_per_gene.join(
+        pathways_per_gene, "geneId", "inner"
+    ).select(
+        "diseaseSet",
+        "geneId",
+        (f.col("enrichedPathwaysPerGene") / f.col("pathwaysPerGene")).alias("score"),
+    )
+
+    genes_in_window = (
+        study_loci_to_annotate.df.select("studyLocusId")
+        .distinct()
+        .join(
+            study_locus.df.select(
+                "studyLocusId", "studyId", "chromosome", "position"
+            ),
+            "studyLocusId",
+            "inner",
+        )
+        .join(
+            target_index.df.select(
+                f.col("id").alias("geneId"),
+                f.col("genomicLocation.chromosome").alias("geneChromosome"),
+                "tss",
+            ),
+            on=(f.col("chromosome") == f.col("geneChromosome"))
+            & (f.abs(f.col("tss") - f.col("position")) <= genomic_window),
+            how="inner",
+        )
+        .select("studyLocusId", "studyId", "geneId")
+    )
+
+    return (
+        genes_in_window.join(disease_sets, "studyId", "left")
+        .join(scores, ["diseaseSet", "geneId"], "left")
+        .select(
+            "studyLocusId",
+            "geneId",
+            f.coalesce(f.col("score"), f.lit(0.0)).alias(feature_name),
+        )
+        .distinct()
+    )
+
+
+def common_neighbourhood_pathway_enrichment_feature_logic(
+    study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+    feature_name: str,
+    **kwargs: Any,
+) -> DataFrame:
+    """Rank the pathway enrichment score of a gene against the other genes at the same locus.
+
+    The score itself is dense - most genes in a window belong to at least one enriched
+    pathway - so what distinguishes genes is how they compare with their neighbours. This
+    divides each gene's score by the largest score at the locus.
+
+    Args:
+        study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci
+            that will be used for annotation
+        feature_name (str): The name of the neighbourhood feature, ending in "Neighbourhood"
+        **kwargs (Any): Arguments of `common_pathway_enrichment_feature_logic`
+
+    Returns:
+        DataFrame: Feature dataset with one row per study locus and gene in its window
+    """
+    local_feature_name = feature_name.replace("Neighbourhood", "")
+    local_scores = common_pathway_enrichment_feature_logic(
+        study_loci_to_annotate, local_feature_name, **kwargs
+    )
+    regional_max = f.max(local_feature_name).over(Window.partitionBy("studyLocusId"))
+    return (
+        local_scores.withColumn("regionalMax", regional_max)
+        .withColumn(
+            feature_name,
+            f.when(
+                f.col("regionalMax") > 0.0,
+                f.col(local_feature_name) / f.col("regionalMax"),
+            ).otherwise(f.lit(0.0)),
+        )
+        .drop("regionalMax", local_feature_name)
+    )
+
+
+class PathwayEnrichmentFeature(L2GFeature):
+    """Fraction of a gene's pathways that are enriched for the diseases of the study."""
+
+    feature_dependency_type = [
+        PathwayIndex,
+        PathwayEnrichment,
+        StudyIndex,
+        StudyLocus,
+        TargetIndex,
+    ]
+    feature_name = "pathwayEnrichment500kb"
+    fdr_threshold: float = 0.05
+    genomic_window: int = 500_000
+
+    @classmethod
+    def compute(
+        cls: type[PathwayEnrichmentFeature],
+        study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+        feature_dependency: dict[str, Any],
+    ) -> PathwayEnrichmentFeature:
+        """Computes the feature.
+
+        Args:
+            study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci that will be used for annotation
+            feature_dependency (dict[str, Any]): Datasets with the pathway library, the enrichment results, the studies, the credible sets and the genes
+
+        Returns:
+            PathwayEnrichmentFeature: Feature dataset
+        """
+        return cls(
+            _df=convert_from_wide_to_long(
+                common_pathway_enrichment_feature_logic(
+                    study_loci_to_annotate,
+                    cls.feature_name,
+                    fdr_threshold=cls.fdr_threshold,
+                    genomic_window=cls.genomic_window,
+                    **feature_dependency,
+                ),
+                id_vars=("studyLocusId", "geneId"),
+                var_name="featureName",
+                value_name="featureValue",
+            ),
+            _schema=cls.get_schema(),
+        )
+
+
+class PathwayEnrichmentNeighbourhoodFeature(L2GFeature):
+    """Pathway enrichment score of a gene relative to the maximum at the same locus."""
+
+    feature_dependency_type = [
+        PathwayIndex,
+        PathwayEnrichment,
+        StudyIndex,
+        StudyLocus,
+        TargetIndex,
+    ]
+    feature_name = "pathwayEnrichment500kbNeighbourhood"
+    fdr_threshold: float = 0.05
+    genomic_window: int = 500_000
+
+    @classmethod
+    def compute(
+        cls: type[PathwayEnrichmentNeighbourhoodFeature],
+        study_loci_to_annotate: StudyLocus | L2GGoldStandard,
+        feature_dependency: dict[str, Any],
+    ) -> PathwayEnrichmentNeighbourhoodFeature:
+        """Computes the feature.
+
+        Args:
+            study_loci_to_annotate (StudyLocus | L2GGoldStandard): The dataset containing study loci that will be used for annotation
+            feature_dependency (dict[str, Any]): Datasets with the pathway library, the enrichment results, the studies, the credible sets and the genes
+
+        Returns:
+            PathwayEnrichmentNeighbourhoodFeature: Feature dataset
+        """
+        return cls(
+            _df=convert_from_wide_to_long(
+                common_neighbourhood_pathway_enrichment_feature_logic(
+                    study_loci_to_annotate,
+                    cls.feature_name,
+                    fdr_threshold=cls.fdr_threshold,
+                    genomic_window=cls.genomic_window,
+                    **feature_dependency,
+                ),
+                id_vars=("studyLocusId", "geneId"),
+                var_name="featureName",
+                value_name="featureValue",
+            ),
+            _schema=cls.get_schema(),
+        )
--- a/src/gentropy/dataset/pathway_enrichment.py
+++ b/src/gentropy/dataset/pathway_enrichment.py
@@ -1,0 +1,80 @@
+"""Disease-pathway enrichment dataset."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from pyspark.sql import Window
+from pyspark.sql import functions as f
+
+from gentropy.common.schemas import parse_spark_schema
+from gentropy.dataset.dataset import Dataset
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame
+    from pyspark.sql.types import StructType
+
+
+@dataclass
+class PathwayEnrichment(Dataset):
+    """Pathways enriched among the genes associated with a disease.
+
+    One row per disease and pathway, as produced by running gene set enrichment analysis
+    over the genes ranked by their association with the disease. Gene set membership is not
+    part of this dataset; it lives in the matching
+    [`PathwayIndex`][gentropy.dataset.pathway_index.PathwayIndex].
+    """
+
+    @classmethod
+    def get_schema(cls: type[PathwayEnrichment]) -> StructType:
+        """Provide the schema for the PathwayEnrichment dataset.
+
+        Returns:
+            StructType: The schema of the PathwayEnrichment dataset.
+        """
+        return parse_spark_schema("pathway_enrichment.json")
+
+    def with_corrected_fdr(self: PathwayEnrichment) -> PathwayEnrichment:
+        """Fill in a missing FDR with one recomputed from the p-values.
+
+        Enrichment results can come with an FDR that the upstream tool failed to estimate,
+        sometimes for every pathway of a disease at once, which would silently remove that
+        disease from any FDR filter. Where `fdr` is null, this replaces it with the
+        Benjamini-Hochberg step-up value computed over the p-values of that disease:
+        `q(i) = min over j >= i of p(j) * n / j`, with the pathways ordered by ascending
+        p-value. Pathways with no p-value keep a null FDR and are left out of `n`.
+
+        Returns:
+            PathwayEnrichment: Dataset where `fdr` is null only if `pval` is too.
+        """
+        by_disease = Window.partitionBy("diseaseId")
+        by_pval = by_disease.orderBy(f.col("pval").asc_nulls_last())
+        n_tested = f.count("pval").over(by_disease)
+        # rank() rather than row_number() so that tied p-values get the same BH value.
+        raw_fdr = f.col("pval") * n_tested / f.rank().over(by_pval)
+        step_up = f.min(raw_fdr).over(
+            by_pval.rowsBetween(Window.currentRow, Window.unboundedFollowing)
+        )
+        return PathwayEnrichment(
+            _df=self.df.withColumn(
+                "fdr", f.coalesce(f.col("fdr"), f.least(step_up, f.lit(1.0)))
+            ),
+            _schema=self.get_schema(),
+        )
+
+    def enriched_pathways(self: PathwayEnrichment, fdr_threshold: float) -> DataFrame:
+        """Diseases and the pathways significantly enriched among their associated genes.
+
+        Args:
+            fdr_threshold (float): Maximum FDR for a pathway to count as enriched.
+
+        Returns:
+            DataFrame: Dataframe with `diseaseId` and `pathway` columns.
+        """
+        return (
+            self.with_corrected_fdr()
+            .df.filter(f.col("fdr") < fdr_threshold)
+            .select("diseaseId", "pathway")
+            .distinct()
+        )
--- a/src/gentropy/dataset/pathway_index.py
+++ b/src/gentropy/dataset/pathway_index.py
@@ -1,0 +1,83 @@
+"""Pathway index dataset."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from pyspark.sql import functions as f
+
+from gentropy.common.schemas import parse_spark_schema
+from gentropy.dataset.dataset import Dataset
+
+if TYPE_CHECKING:
+    from pyspark.sql import DataFrame
+    from pyspark.sql.types import StructType
+
+    from gentropy.common.session import Session
+
+
+@dataclass
+class PathwayIndex(Dataset):
+    """Gene set membership of a pathway library.
+
+    One row per pathway, holding the gene symbols that make up the gene set. This is the
+    reference that disease-pathway enrichment results are computed against, and it is what
+    turns an enriched pathway back into a set of genes.
+    """
+
+    SOURCE_PATTERN = r"\[([^\]]+)\]\s*$"
+    """Pattern extracting the library a pathway came from out of its name, e.g. `[Reactome]`."""
+
+    @classmethod
+    def get_schema(cls: type[PathwayIndex]) -> StructType:
+        """Provide the schema for the PathwayIndex dataset.
+
+        Returns:
+            StructType: The schema of the PathwayIndex dataset.
+        """
+        return parse_spark_schema("pathway_index.json")
+
+    @classmethod
+    def from_gmt(cls: type[PathwayIndex], session: Session, path: str) -> PathwayIndex:
+        """Parse a gene matrix transposed (GMT) file into a PathwayIndex.
+
+        A GMT file holds one tab separated gene set per line: the pathway name, a
+        description that is ignored here, and then the gene symbols.
+
+        Args:
+            session (Session): Spark session
+            path (str): Path to the GMT file
+
+        Returns:
+            PathwayIndex: Pathway index dataset
+        """
+        columns = f.split(f.col("value"), "\t")
+        pathway = columns.getItem(0)
+        return cls(
+            _df=(
+                session.spark.read.text(path)
+                .filter(f.trim(f.col("value")) != "")
+                .select(
+                    pathway.alias("pathway"),
+                    f.regexp_extract(pathway, cls.SOURCE_PATTERN, 1).alias("source"),
+                    f.array_distinct(
+                        f.filter(
+                            f.slice(columns, 3, f.size(columns)), lambda gene: gene != ""
+                        )
+                    ).alias("geneSymbols"),
+                )
+                .filter(f.size("geneSymbols") > 0)
+            ),
+            _schema=cls.get_schema(),
+        )
+
+    def gene_membership(self: PathwayIndex) -> DataFrame:
+        """Explode the gene sets into one row per pathway and gene symbol.
+
+        Returns:
+            DataFrame: Dataframe with `pathway` and `geneSymbol` columns.
+        """
+        return self.df.select(
+            "pathway", f.explode("geneSymbols").alias("geneSymbol")
+        ).distinct()
--- a/src/gentropy/l2g.py
+++ b/src/gentropy/l2g.py
@@ -20,6 +20,8 @@ from gentropy.dataset.intervals import Intervals
 from gentropy.dataset.l2g_feature_matrix import L2GFeatureMatrix
 from gentropy.dataset.l2g_gold_standard import L2GGoldStandard
 from gentropy.dataset.l2g_prediction import L2GPrediction
+from gentropy.dataset.pathway_enrichment import PathwayEnrichment
+from gentropy.dataset.pathway_index import PathwayIndex
 from gentropy.dataset.study_index import StudyIndex
 from gentropy.dataset.study_locus import StudyLocus
 from gentropy.dataset.target_index import TargetIndex
@@ -61,6 +63,8 @@ class LocusToGeneFeatureMatrixStep:
         target_index_path: str | None = None,
         intervals_path: str | None = None,
         gene_interactions_path: str | None = None,
+        pathway_index_path: str | None = None,
+        pathway_enrichment_path: str | None = None,
         feature_matrix_path: str,
         append_null_features: bool = False,
     ) -> None:
@@ -76,6 +80,8 @@ class LocusToGeneFeatureMatrixStep:
             target_index_path (str | None): Path to the target index dataset
             intervals_path (str | None): Path to the interval dataset
             gene_interactions_path (str | None): Path to the protein-protein interaction (PPI) dataset
+            pathway_index_path (str | None): Path to the pathway library, as a GMT file
+            pathway_enrichment_path (str | None): Path to the disease-pathway enrichment dataset
             feature_matrix_path (str): Path to the L2G feature matrix output dataset
             append_null_features (bool): Whether to append null features to the feature matrix. Defaults to False.
         """
@@ -122,6 +128,20 @@ class LocusToGeneFeatureMatrixStep:
             else None
         )
 
+        pathway_index = (
+            PathwayIndex.from_gmt(session, pathway_index_path)
+            if pathway_index_path
+            else None
+        )
+
+        # No recursive file lookup here: the enrichment results are partitioned by diseaseId,
+        # and recursing would stop Spark from reading that partition column back.
+        pathway_enrichment = (
+            PathwayEnrichment.from_parquet(session, pathway_enrichment_path)
+            if pathway_enrichment_path
+            else None
+        )
+
         trans_pqtl_features = {
             "transPQtlColocH4Maximum",
             "transPQtlColocH4MaximumNeighbourhood",
@@ -137,6 +157,23 @@ class LocusToGeneFeatureMatrixStep:
                 "Provide `target_index_path`."
             )
 
+        pathway_features = {
+            "pathwayEnrichment500kb",
+            "pathwayEnrichment500kbNeighbourhood",
+        }
+        if pathway_features.intersection(features_list) and (
+            pathway_index is None
+            or pathway_enrichment is None
+            or target_index is None
+            or studies is None
+        ):
+            raise ValueError(
+                "Pathway enrichment features need the pathway library, the enrichment "
+                "results, the target index and the study index. Provide "
+                "`pathway_index_path`, `pathway_enrichment_path`, `target_index_path` "
+                "and `study_index_path`."
+            )
+
         features_input_loader = L2GFeatureInputLoader(
             variant_index=variant_index,
             colocalisation=coloc,
@@ -145,6 +182,8 @@ class LocusToGeneFeatureMatrixStep:
             target_index=target_index,
             intervals=intervals,
             interactions=interactions,
+            pathway_index=pathway_index,
+            pathway_enrichment=pathway_enrichment,
         )
 
         fm = credible_set.filter(f.col("studyType") == "gwas").build_feature_matrix(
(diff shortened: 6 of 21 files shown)


src/gentropy/dataset/pathway_index.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@dataclass
+class PathwayIndex(Dataset):
+    """Gene set membership of a pathway library.
+
+    One row per pathway, holding the gene symbols that make up the gene set. This is the
+    reference that disease-pathway enrichment results are computed against, and it is what
+    turns an enriched pathway back into a set of genes.
+    """
```
project-defiant, line 27: Would be useful to see example dataset.

src/gentropy/dataset/pathway_index.py
```diff
@@ -0,0 +1,83 @@
...
+        A GMT file holds one tab separated gene set per line: the pathway name, a
+        description that is ignored here, and then the gene symbols.
+
+        Args:
+            session (Session): Spark session
+            path (str): Path to the GMT file
+
+        Returns:
+            PathwayIndex: Pathway index dataset
+        """
```
project-defiant, line 54: Would be useful to see example input gmt and output dataset

tests/gentropy/dataset/test_pathway.py
```diff
@@ -0,0 +1,141 @@
...
+        encoding="utf-8",
+    )
+
+    observed = {
+        row["pathway"]: (row["source"], row["geneSymbols"])
+        for row in PathwayIndex.from_gmt(session, str(gmt)).df.collect()
+    }
+    assert observed == {
+        "pathway1 [Reactome]": ("Reactome", ["GENE1", "GENE2"]),
+        "pathway2 [GO BP]": ("GO BP", ["GENE3"]),
```
project-defiant, line 44: why the Reactome, GO BP annotations are duplicated?

src/gentropy/dataset/pathway_index.py
```diff
@@ -0,0 +1,83 @@
...
+        columns = f.split(f.col("value"), "\t")
+        pathway = columns.getItem(0)
+        return cls(
+            _df=(
+                session.spark.read.text(path)
+                .filter(f.trim(f.col("value")) != "")
+                .select(
+                    pathway.alias("pathway"),
+                    f.regexp_extract(pathway, cls.SOURCE_PATTERN, 1).alias("source"),
```
project-defiant, line 63: I think, that even if you extracted the `source` from the `pathway`, the bracket source still remains in the `pathway` column. Is it useful to have duplicate info there?

src/gentropy/dataset/pathway_index.py
```diff
@@ -0,0 +1,83 @@
...
+                    f.array_distinct(
+                        f.filter(
+                            f.slice(columns, 3, f.size(columns)), lambda gene: gene != ""
+                        )
+                    ).alias("geneSymbols"),
+                )
+                .filter(f.size("geneSymbols") > 0)
+            ),
+            _schema=cls.get_schema(),
+        )
```
project-defiant, line 73: I think the schema is incomplete, the original gmt file reports the `pathway, description, gene1, gene2, ...`
which means that the `description` is currently lost. 

Second thing is that in order to have a proper validation of the Pathway names in general so they can be resolved to proper Ids instead of names. I would remove the pathway name, join to existing source that we have - ex. Reactome and ensure the ids are in place, but I lack of the understanding of the full process right now to place it correctly into the pipeline and the info on what kind of gene sets source we actually use (GO, reactome, ...) ?

The same thing is about the geneIds (ideally the Index shall put the geneIds resolved by the current release not only gene symbols).

src/gentropy/dataset/pathway_enrichment.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def with_corrected_fdr(self: PathwayEnrichment) -> PathwayEnrichment:
+        """Fill in a missing FDR with one recomputed from the p-values.
+
+        Enrichment results can come with an FDR that the upstream tool failed to estimate,
+        sometimes for every pathway of a disease at once, which would silently remove that
+        disease from any FDR filter. Where `fdr` is null, this replaces it with the
+        Benjamini-Hochberg step-up value computed over the p-values of that disease:
+        `q(i) = min over j >= i of p(j) * n / j`, with the pathways ordered by ascending
+        p-value. Pathways with no p-value keep a null FDR and are left out of `n`.
```
project-defiant, line 46: What is the reason the GSEA fdr can not be computed?

docs/assets/schemas/pathway_enrichment.md
```diff
@@ -0,0 +1,11 @@
+```
+root
+ |-- diseaseId: string (nullable = false)
+ |-- pathway: string (nullable = false)
+ |-- source: string (nullable = true)
+ |-- nes: double (nullable = true)
+ |-- pval: double (nullable = true)
```
project-defiant, line 7: It is worth to harmonise the field names with what other datasets use 
ex pval -> pValue

`nes` is hard to understand without prior knowledge.

src/gentropy/dataset/pathway_enrichment.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns:
+            PathwayEnrichment: Dataset where `fdr` is null only if `pval` is too.
+        """
+        by_disease = Window.partitionBy("diseaseId")
```
project-defiant: More problems with EFO duplicates probably...

project-defiant, line 51: Two general comments. I looked into the R implementation of `p.adjust(p, method = "BH")`, which appears to be the behaviour this code is intended to reproduce. https://support.bioconductor.org/p/49864/
1. (minor)The implementation computes adjusted p-values over all rows within a disease partition. This is correct for rows with a non-null p-value. The current use of f.count("pval") is also correct because count(column) excludes nulls when calculating the number of tested pathways. However, the window expressions are still evaluated for rows where pval is null. Although these rows ultimately remain null, we could make the intent clearer by explicitly skipping the calculation for null p-values with a when expression, or by filtering them out before applying the BH calculation.
2. (major) The handling of tied p-values differs between `Spark’s rank()` and the positional ranking used by the R BH implementation. **Spark’s rank() assigns the same rank to tied values, whereas BH-adjusted p-values require sequential positions in the sorted list.** As a result, using rank() can produce different adjusted p-values for ties. (based on [R-bloggers algorithm](https://www.r-bloggers.com/2023/07/the-benjamini-hochberg-procedure-fdr-and-p-value-adjusted-explained/)).
For example:
```{R}
pvals <- c(0.04, 0.04, 0.20)
m <- length(pvals)

# R's BH adjustment
p.adjust(pvals, method = "BH")
# [1] 0.06 0.06 0.20

# Spark-like rank(): ties receive the same rank
spark_rank <- rank(pvals, ties.method = "min")
spark_rank
# [1] 1 1 3

raw_bh <- pvals * m / spark_rank
raw_bh
# [1] 0.12 0.12 0.20

spark_like_adj <- rev(cummin(rev(raw_bh)))
spark_like_adj
# [1] 0.12 0.12 0.20
```
I think row_number() should be used for the BH position instead of rank(), with a deterministic secondary ordering for tied p-values. This would make the implementation consistent with the R BH adjustment.

src/gentropy/dataset/pathway_enrichment.py
```diff
@@ -0,0 +1,80 @@
...
+        by_pval = by_disease.orderBy(f.col("pval").asc_nulls_last())
+        n_tested = f.count("pval").over(by_disease)
+        # rank() rather than row_number() so that tied p-values get the same BH value.
+        raw_fdr = f.col("pval") * n_tested / f.rank().over(by_pval)
+        step_up = f.min(raw_fdr).over(
+            by_pval.rowsBetween(Window.currentRow, Window.unboundedFollowing)
+        )
+        return PathwayEnrichment(
+            _df=self.df.withColumn(
+                "fdr", f.coalesce(f.col("fdr"), f.least(step_up, f.lit(1.0)))
```
project-defiant, line 61: We end up with p.adjust_BH=1.0 in case p-value is undefined. That is again different then the R implemenation, which returns the NULL for NULL pvalues.

project-defiant, line 61: Is the `fdr` a correct name for p-value adjusted ? From my readings it looks like it is imprecise as the fdr is the property of the entire tested set and here we return a value for individual test.

tests/gentropy/dataset/test_pathway.py
```diff
@@ -0,0 +1,141 @@
...
+        .collect()
+    }
+    assert observed == {
+        "pathway1 [Reactome]": pytest.approx(0.003),
+        "pathway2 [GO BP]": pytest.approx(0.03),
+        "pathway3 [GO BP]": pytest.approx(0.5),
+    }
+
+
+def test_with_corrected_fdr_is_monotonic(spark: SparkSession) -> None:
```
project-defiant, line 101: Worth adding an edge case with NULL pvalue

tests/gentropy/dataset/test_pathway.py
project-defiant, line 1: NIT:
I was looking at the tests for the PathwayEnrichment dataset in a separate file (named after the dataset). Would be useful to move the out to `test_pathway_enrichment.py`

src/gentropy/l2g.py
```diff
@@ -76,6 +80,8 @@ def __init__(
             target_index_path (str | None): Path to the target index dataset
             intervals_path (str | None): Path to the interval dataset
             gene_interactions_path (str | None): Path to the protein-protein interaction (PPI) dataset
+            pathway_index_path (str | None): Path to the pathway library, as a GMT file
```
project-defiant, line 83: Is it expected to be a parquet of PathwayIndex or a GMT file? it's not obvious.

src/gentropy/l2g.py
```diff
@@ -122,6 +128,20 @@ def __init__(
             else None
         )
 
+        pathway_index = (
+            PathwayIndex.from_gmt(session, pathway_index_path)
```
project-defiant, line 132: I would probably not try to read a datasource (gmt file) here, rather just load the `pathway_index.parquet` that is already harmonised to a gentropy format. This would allow us to ensure the quality controls (geneId, pathwayId are valid). Especially that you already do this with `PathwayEnrichment` (which is read as a dataset instead of building it on the fly from source).

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+    # Gene set membership, restricted to the pathways enrichment was actually tested for and
+    # translated from gene symbols into the gene identifiers the feature matrix is keyed on.
+    membership = (
+        pathway_index.gene_membership()
+        .join(pathway_enrichment.df.select("pathway").distinct(), "pathway", "inner")
+        .join(
+            target_index.symbols_lut().select("geneSymbol", "geneId"),
+            "geneSymbol",
+            "inner",
+        )
```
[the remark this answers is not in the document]
project-defiant: agreed

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # Studies are grouped by their set of diseases rather than handled one by one: there are
+    # far fewer distinct disease sets than studies, and the great majority hold one disease.
+    disease_sets = (
+        study_index.df.select(
+            "studyId", f.array_sort(f.col("diseaseIds")).alias("diseaseSet")
+        )
+        .filter(f.size("diseaseSet") > 0)
+        .distinct()
+    )
```
project-defiant, line 86: I am not sure if the molQTLs have empty list or NULL in the diseaseIds field. In that context we should probably 
```{python}
study_index.df
  .select("studyId", "diseaseIds")
  .filter(f.col("studyType") = f.lit('gwas'))
  .filter(f.size(f.col("diseaseIds") > 0)
  .select("studyId", f.array_sort(f.col("diseaseIds")).alias("diseaseSet"))
```
This way you first filter, then do the sorting on smaller list

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+    enriched_pathways_per_gene = (
+        disease_sets.select("diseaseSet")
+        .distinct()
+        .select("diseaseSet", f.explode("diseaseSet").alias("diseaseId"))
```
project-defiant, line 90: You will end up with duplicates.
ex. 
```
A: [diseaseA, diseaseB, diseaseC]
B: [diseaseA, diseaseB]

# apply distinct
A: [diseaseA, diseaseB, diseaseC]
B: [diseaseA, diseaseB]
```

What you need is
```suggestion
    enriched_pathways_per_gene = (
        disease_sets
        .select("diseaseSet", f.explode("diseaseSet").alias("diseaseId")).distinct()
```

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+        .join(
+            pathway_enrichment.enriched_pathways(fdr_threshold), "diseaseId", "inner"
+        )
+        .select("diseaseSet", "pathway")
+        .distinct()
```
project-defiant, line 95: Is this meant to filter the enrichment to GWAS diseases?
In that case do the semi join.

src/gentropy/dataset/pathway_enrichment.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Args:
+            fdr_threshold (float): Maximum FDR for a pathway to count as enriched.
+
+        Returns:
+            DataFrame: Dataframe with `diseaseId` and `pathway` columns.
+        """
+        return (
+            self.with_corrected_fdr()
+            .df.filter(f.col("fdr") < fdr_threshold)
```
project-defiant, line 77: If you change the fdr for null pvalues to be NULl instead of 1.0. ensure to remove it here as well.

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+        .groupBy("diseaseSet", "geneId")
+        .agg(f.count("pathway").alias("enrichedPathwaysPerGene"))
+    )
+    scores = enriched_pathways_per_gene.join(
+        pathways_per_gene, "geneId", "inner"
+    ).select(
+        "diseaseSet",
+        "geneId",
+        (f.col("enrichedPathwaysPerGene") / f.col("pathwaysPerGene")).alias("score"),
+    )
```
project-defiant, line 106: I am not 100% sure, but it feels that we are loosing information here. 
The inner join will only give you the genes that have enrichment, so you can never get a value of 0 for this ratio.

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+    genes_in_window = (
+        study_loci_to_annotate.df.select("studyLocusId")
+        .distinct()
+        .join(
+            study_locus.df.select(
+                "studyLocusId", "studyId", "chromosome", "position"
+            ),
+            "studyLocusId",
+            "inner",
+        )
```
project-defiant, line 117: again a case of semi join

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+        .join(
+            target_index.df.select(
+                f.col("id").alias("geneId"),
+                f.col("genomicLocation.chromosome").alias("geneChromosome"),
+                "tss",
+            ),
+            on=(f.col("chromosome") == f.col("geneChromosome"))
+            & (f.abs(f.col("tss") - f.col("position")) <= genomic_window),
+            how="inner",
+        )
```
project-defiant, line 127: Do we need to ensure to use either end of the gene (whichever is closer) to the lead variant?

src/gentropy/dataset/l2g_features/pathway.py
```diff
@@ -0,0 +1,270 @@
...
+        local_scores.withColumn("regionalMax", regional_max)
+        .withColumn(
+            feature_name,
+            f.when(
+                f.col("regionalMax") > 0.0,
+                f.col(local_feature_name) / f.col("regionalMax"),
+            ).otherwise(f.lit(0.0)),
+        )
+        .drop("regionalMax", local_feature_name)
+    )
```
project-defiant, line 178: the logic is very similar to other neighbourhood features. I am curious if the fact that you can have multiple genes from the same pathway (that are in close proximity, perhaps very similar subunits of the same protein) can influence this feature. It seems to me that we should also account here to the strength of the enrichment of the entire pathway in 500kb (in other words, perhaps using GSEA to compute the enrichment of +/-500kb proxy genes vs pathways).

project-defiant (changes requested):

project-defiant:
@addramir when you are back, please have a look at the comments. I have some concerns about the whole design and few potential bugs that need to be resolved before we can merge this properly.

src/gentropy/dataset/l2g_features/pathway.py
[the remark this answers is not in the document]
addramir: I checked this against the catalogue the feature was developed on, and the pathway set used for
`membership` is the tested set:

- `n_pathways_tested` in the producer's stats file is 8,217 for every one of the 3,766 diseases,
  and the enrichment dataset holds exactly 8,217 distinct pathways.
- A single disease carries fewer rows than that (median 4,107) because the producer keeps only
  positive-NES rows. So the per-disease row set is a subset of the tested set, while the union
  over diseases — which is what the code takes — is the tested set.
- Rows with no p-value are 12,579 of 15,466,831 (0.08%), and there is no pathway whose p-value
  is null in every row, so nothing untestable reaches the denominator today.

The denominator is therefore the same 8,217 pathways for every disease, which is what makes the
score comparable across studies. I will still add the explicit filter on
`pval IS NOT NULL OR fdr IS NOT NULL` before the distinct, so the invariant is stated in the
code and not just true of this particular input, and make that join a left-semi one as you
suggest elsewhere.

src/gentropy/dataset/pathway_index.py
```diff
@@ -0,0 +1,83 @@
...
+                .filter(f.trim(f.col("value")) != "")
+                .select(
+                    pathway.alias("pathway"),
+                    f.regexp_extract(pathway, cls.SOURCE_PATTERN, 1).alias("source"),
+                    f.array_distinct(
+                        f.filter(
+                            f.slice(columns, 3, f.size(columns)), lambda gene: gene != ""
+                        )
+                    ).alias("geneSymbols"),
+                )
```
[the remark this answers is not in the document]
addramir: Both accepted. Neither case occurs in the library this was built against — 0 of the 17,246
symbols have leading or trailing whitespace, and all 8,859 lines carry a `[source]` tag — but
the parser should not depend on that. I will trim the symbols and map an empty extracted source
to null.

src/gentropy/dataset/pathway_index.py
addramir: Examples are in the thread on `from_gmt` below; the same ones are going into this class docstring.

src/gentropy/dataset/pathway_index.py
addramir: Adding this to the docstrings. Three real lines of the library, gene lists cut to four genes
(tabs shown as `\t`):

```
'De Novo' AMP Biosynthetic Process (GO:0044208) [GO BP]\tGO:0044208\tADSL\tADSS1\tADSS2\tATIC…
2-LTR circle formation [Reactome]\tR-HSA-164843\tBANF1\tHMGA1\tLIG4\tPSIP1…
HALLMARK_ADIPOGENESIS [MSigDB Hallmark]\t\tABCA1\tABCB8\tACAA2\tACADL…
```

parsed into `PathwayIndex` (with the `pathwayId` column from the second field, see the thread on
line 73):

| pathway | pathwayId | source | geneSymbols |
| --- | --- | --- | --- |
| `'De Novo' AMP Biosynthetic Process (GO:0044208) [GO BP]` | `GO:0044208` | `GO BP` | `[ADSL, ADSS1, ADSS2, ATIC, …]` |
| `2-LTR circle formation [Reactome]` | `R-HSA-164843` | `Reactome` | `[BANF1, HMGA1, LIG4, PSIP1, …]` |
| `HALLMARK_ADIPOGENESIS [MSigDB Hallmark]` | null | `MSigDB Hallmark` | `[ABCA1, ABCB8, ACAA2, ACADL, …]` |

and `gene_membership()` on the first row gives `('…(GO:0044208) [GO BP]', 'ADSL')`,
`('…(GO:0044208) [GO BP]', 'ADSS1')`, and so on.

For the matching `PathwayEnrichment` rows, two real ones from type 2 diabetes:

| diseaseId | pathway | source | nes | pval | fdr |
| --- | --- | --- | --- | --- | --- |
| `MONDO_0005148` | `Positive Regulation of DNA-templated Transcription (GO:0045893) [GO BP]` | `GO BP` | 6.355 | 2.08e-10 | 8.34e-08 |
| `MONDO_0005148` | `Regulation of Transcription by RNA Polymerase II (GO:0006357) [GO BP]` | `GO BP` | 6.274 | 3.52e-10 | 1.03e-07 |

Library composition: 5,954 GO BP, 2,855 Reactome, 50 MSigDB Hallmark. 642 of the 8,859 pathways
never appear in the enrichment dataset because they fall outside blitzGSEA's 5–4,000 gene size
window, so they were never tested.

src/gentropy/dataset/pathway_index.py
addramir: The bracketed tag has to stay in `pathway` because that column is the join key to the enrichment
dataset, which stores the name verbatim including the tag, e.g.
`T-helper 2 Cell Differentiation (GO:0045064) [GO BP]`. All 8,217 catalogue names match a GMT
name exactly on that string, which is the only link between the two inputs — the catalogue has
no gene column and the library has no enrichment.

So yes, `source` is redundant with the name, and the enrichment dataset carries its own `source`
column too. It is kept because it is the one column you can filter or group on without parsing
strings (the per-source QC in the notebooks uses it). The test fixture looks duplicated because
it mirrors the real shape. If you prefer, I can drop `source` from the index and derive it from
the `pathwayId` prefix instead, now that `pathwayId` is going in.

tests/gentropy/dataset/test_pathway.py
addramir: Answered in the thread on `pathway_index.py:63` — the tag stays in `pathway` because that string is the join key to the enrichment dataset, and the fixture mirrors the real shape.

src/gentropy/dataset/pathway_index.py
addramir: You are right that the second field is being thrown away, and it is worse than "description":
in this library the second field is the stable identifier.

- 5,954 GO BP lines: `GO:0044208` and the like
- 2,855 Reactome lines: `R-HSA-164843` and the like
- 50 MSigDB Hallmark lines: empty

So I will add `pathwayId` to the schema and stop discarding it. GO and Reactome ids can then be
resolved against the real ontologies; Hallmark has no id, hence nullable. The gene set sources
in this library are exactly those three.

On gene ids: agreed, and it fits with your comment on `l2g.py:132`. I will resolve the symbols
to `geneId` when the index is built rather than inside the feature. One caveat worth recording —
the mapping is release-dependent, so the parquet becomes release-specific: against 26.03,
16,711 of the 17,246 symbols match an `approvedSymbol` and 16,733 match once `obsoleteSymbols`
are included (97%), the rest being tRNA and immunoglobulin segment names; 142 symbols map to
more than one Ensembl id, which is why the feature de-duplicates on `(studyLocusId, geneId)`.

src/gentropy/dataset/pathway_enrichment.py
addramir: It is an upstream failure, not something the pipeline chooses not to do. The catalogue is
produced by blitzGSEA 1.3.54 and `fdr` is passed straight through from `blitz.gsea`. It is null
for every row of exactly 249 of the 3,766 diseases — 1,025,450 rows, 6.6% — and not scattered.
Those 249 are the gene-rich diseases (mean 900 genetically associated genes against 115 for the
rest; median `n_used_for_gsea` 410 against 107), and their p-values are intact: 236,469 of their
rows have p < 0.05. They are also precisely the diseases with zero pathways at FDR < 0.05, so
any FDR filter drops them silently. `sidak` is populated but family-wise and far too
conservative to substitute — 179,637 rows pass sidak < 0.05 across the whole catalogue against
2,705,538 at FDR < 0.05.

Why blitzGSEA returned NaN there cannot be told from the outputs; it needs to be reproduced on
the producer's side, and I will ask for the catalogue to be regenerated with a working FDR step.
The recomputation stays in the meantime, with the ranking fix and the caveat in the next thread
written into the docstring, because the alternative is 15% of GWAS studies getting a zero
feature value.

src/gentropy/dataset/pathway_enrichment.py
addramir: Both points accepted, and the second one is measurable. In the 249 diseases where the correction
actually fires, `rank()` and `row_number()` give a different value for 158,715 of 1,012,871 rows
(15.7%), and 163 rows change their FDR < 0.05 call (34,103 significant with `rank()` against
34,266 with `row_number()`). Small in effect, but there is no reason to be wrong: `row_number()`
ordered by `(pval, pathway)` for determinism, and the null p-values skipped explicitly.

While checking that, I found something bigger, which I think settles the design question. The
recomputed value is not on the same scale as the published one. blitzGSEA computed its FDR over
all 8,217 tested pathways of both NES signs, and the catalogue keeps only the positive-NES rows
(`pos = res[res["nes"] > 0]` in the producer's script), a median of 4,107 rows per disease.
Recomputing within the stored rows therefore uses a truncated *m* and shifted ranks. Comparing on
diseases that do have a published FDR, pathways at FDR < 0.05, published against recomputed:

| disease | published | recomputed |
| --- | --- | --- |
| DOID_0050890 | 1,470 | 1,474 |
| EFO_0000094 | 749 | 627 |
| EFO_0001073 | 423 | 5 |
| EFO_0004340 | 465 | 88 |

So the repair is not a like-for-like substitute, and the 249 repaired diseases would sit on a
different and disease-dependent scale from the other 3,517 — in one column, used by one
threshold. There is a second problem in the same direction: 18,577 rows have `nes = +inf` with
`pval = 0` (12,864 of them inside the 249), and a recomputed BH hands those an FDR of exactly 0,
so the most degenerate fits would become the most significant pathways.

What I am proposing to do with that: keep the recomputation, because dropping it means the 249
diseases — about 15% of GWAS studies, including body mass index, psoriasis, ulcerative colitis
and systemic lupus erythematosus — get a zero feature value, and fix what can be fixed:

- `row_number()` over `(pval, pathway)` and null p-values skipped, as above;
- the docstring saying plainly that the recomputed value is adjusted over the stored positive-NES
  rows only, so it is more conservative than and not comparable with the published column, and
  that it exists to rescue the 249 diseases rather than to reproduce blitzGSEA;
- and in parallel, asking for the catalogue to be regenerated with a working FDR step, ideally
  exporting *m* and each row's rank or keeping the negative-NES rows, after which this method
  becomes dead code and goes.

Two things I would like your view on. Whether the `nes = +inf`, `pval = 0` rows should be excluded
from the recomputation instead of being handed an adjusted p-value of 0 — I lean towards
excluding them, since an infinite NES means the permutation fit degenerated. And whether the
correction should be behind an explicit flag rather than applied inside `enriched_pathways`, so
the two scales can never be mixed without someone asking for it.

src/gentropy/dataset/pathway_enrichment.py
addramir: Confirmed, and it is `least` that does it: Spark's `least` skips nulls, so `least(NULL, 1.0)` is
`1.0` rather than null. If the recomputation stays in any form it gets a
`when(pval.isNotNull(), …)` guard so a null p-value keeps a null adjusted value, as in R.

addramir: Agreed — what upstream calls `fdr` is a Benjamini-Hochberg adjusted p-value per test, not the
false discovery rate of the set. Renaming it as part of the harmonisation in the schema thread.

docs/assets/schemas/pathway_enrichment.md
addramir: Agreed. Proposed: `pval` → `pValue`, `fdr` → `pValueAdjusted`, `nes` →
`normalisedEnrichmentScore`, and the enrichment score dropped since the feature never reads it.
On `pValue` as a plain double rather than the `pValueMantissa`/`pValueExponent` pair used
elsewhere: these p-values are permutation p-values and do not underflow — the smallest non-zero
one in the catalogue is 5.3e-51, and 5,998 rows are exactly 0 — so the split would buy nothing.
Tell me if you would rather have the pair for consistency anyway.

tests/gentropy/dataset/test_pathway.py
addramir: Will add, asserting a null adjusted value rather than 1.0, together with the fix in the thread
above.

tests/gentropy/dataset/test_pathway.py
addramir: Will split into `test_pathway_index.py` and `test_pathway_enrichment.py`.

src/gentropy/l2g.py
addramir: A GMT today, which is exactly the problem you point at on line 132 — it becomes a harmonised `PathwayIndex` parquet, and the docstring will say so.

src/gentropy/l2g.py
addramir: Agreed, and it answers your ids comment too. I will move the GMT parsing out of
`LocusToGeneFeatureMatrixStep` into its own step that writes a harmonised `PathwayIndex`
parquet — `pathway`, `pathwayId`, `source`, `geneSymbols` and `geneIds` resolved against the
target index — and the feature matrix step will read only that parquet, like every other input.
The docstring will say so.

src/gentropy/dataset/l2g_features/pathway.py
addramir: Taking the suggestion. For the record it is about doing less work rather than correctness: the
feature matrix is already built from `credible_set.filter(studyType == 'gwas')`, and in 26.03 all
136,245 GWAS studies carry at least one disease id, while `size(NULL)` is dropped by the
`> 0` filter either way. Filtering first and sorting the smaller frame is still the right order.

src/gentropy/dataset/l2g_features/pathway.py
addramir: The duplicates you describe are collapsed by the `.distinct()` five lines down, on
`(diseaseSet, pathway)` — a pathway flagged by two diseases of the same set survives once, which
is the reason the sets are handled as sets in the first place. Your version does the dedup
earlier on a smaller frame, so I will take it; it is not a behaviour change.

src/gentropy/dataset/l2g_features/pathway.py
addramir: No, this one is not a filter: the join brings the `pathway` column in from the enrichment side,
and the union of pathways over the diseases of a set is then de-duplicated. A semi join would
give back only the left-hand columns, so it cannot produce that. The `membership` join on line 65
*is* a filter and I am making that one a semi join.

src/gentropy/dataset/pathway_enrichment.py
addramir: Nothing to remove here — `fdr < fdr_threshold` already drops nulls, so a null adjusted value
behaves as "not enriched". The 1.0 was only ever produced by the `least` in `with_corrected_fdr`.

src/gentropy/dataset/l2g_features/pathway.py
addramir: The zeros come in at the end: `genes_in_window` is joined to `scores` with a left join and the
missing score is coalesced to 0.0 (lines 132–137), so every gene in the window with no enriched
pathway gets 0. The inner join here only restricts the score table to genes that have both
counts, and those genes all have at least one enriched pathway by construction. In the real data
the zeros are the majority for most diseases: for type 2 diabetes 9,039 of the 17,246 library
genes have a non-zero score and the rest are 0.

src/gentropy/dataset/l2g_features/pathway.py
addramir: Right, this one works as a semi join — every column needed comes from `study_locus` — so it
becomes `study_locus.df.select(…).join(study_loci_to_annotate.df.select("studyLocusId").distinct(), "studyLocusId", "semi")`.

src/gentropy/dataset/l2g_features/pathway.py
addramir: I used the TSS because that is what the existing window features use — the gene-count and
distance features in `l2g_features/other.py` select genes by `genes.tss` inside the window, and
the trans-pQTL feature does the same on `targetB_tss`. Changing it here alone would make the
pathway feature's gene set differ from `geneCount500kb` at the same locus, so I would rather
change them together if we change it. Worth flagging a related inconsistency I hit while writing
this: `other.py` treats `genomic_window` as the full width (±250 kb for 500,000) while
`distance.py` and this feature treat it as a radius, so "500kb" in these two feature names means
different spans.

src/gentropy/dataset/l2g_features/pathway.py
addramir: The per-gene half of the concern is handled by the denominator: a gene's score is divided by its
own pathway count, so hub genes are not favoured — for type 2 diabetes the Spearman correlation
between the number of pathways a gene belongs to and its score is 0.118, and `RPS27A`, which sits
in 462 pathways, does not score highly.

What is not handled is the case you describe: several genes in the same 500 kb window belonging
to the same enriched pathway, which is exactly what subunits of one complex look like. The
neighbourhood feature then gives all of them a high value. I would argue that is a genuine
ambiguity in the data rather than an artefact of the statistic — the evidence really does not
distinguish them — but I agree a locus-level statistic (GSEA of the window's genes against the
enriched pathways, or a permutation of the window) is a different and probably better feature,
and it would also fix the other soft spot I found: the denominator has a median of 11 and a
minimum of 1, so every gene that reaches exactly 1.0 does so from 2–4 pathways, outranking a gene
with 60 of its 100 pathways enriched. A shrunk ratio, `(enriched + r·k) / (total + k)` with `r`
the disease-wide hit rate and `k` around 5, fixes that ordering cheaply.

My preference is to keep this PR's statistic as the simple opt-in one and do the locus-level
version as a separate feature we can compare against it, but if you would rather not merge the
simple one at all, that is a reasonable call too, and the circularity point in the PR description
argues the same way.

addramir:
Thanks — going through these has changed two things beyond the line-level fixes.

1. `with_corrected_fdr` is not reproducing the published quantity, and the code has to say so.
   blitzGSEA adjusted over all 8,217 tested pathways of both NES signs, while the catalogue stores
   only the positive-NES rows (median 4,107 per disease), so recomputing over the stored rows uses
   a truncated *m* with shifted ranks and lands anywhere between 1% and 100% of the published count
   of enriched pathways depending on the disease. I am keeping it, because the alternative is a
   zero feature value for the 249 broken diseases and the 15% of GWAS studies pointing at them, but
   with `row_number()` ranking, null p-values left null, and the caveat stated in the docstring —
   and with a request upstream to regenerate the catalogue with a working FDR step, which would
   retire the method. Two open questions are in that thread: whether to exclude the
   `nes = +inf, pval = 0` rows from the correction, and whether it should sit behind an explicit
   flag.
2. The pathway library should be ingested as its own step into a harmonised parquet — with
   `pathwayId` (GO and Reactome ids are in the GMT's second field, which the current parser
   discards) and `geneId` resolved against the target index — and the feature matrix step should
   read only that parquet. Your `l2g.py:132` comment and your ids comment are the same fix.

The semi joins, the `row_number()` ranking, the field renaming, the test split and the earlier
dedup are all straightforward and I will push them together with the above.

addramir:
Pushed `b175e147`, which implements what I said in the threads above.

- `PathwayIndex` gains `pathwayId` from the GMT's second field and `geneIds` resolved through `TargetIndex.symbols_lut`, so the feature no longer maps symbols itself; `gene_membership` now returns `(pathway, geneId)` and refuses to run on an index whose identifiers were never resolved. Symbols are trimmed and an empty source comes back null.
- New `PathwayIngestionStep` (`pathway_ingestion`) writes both harmonised datasets from the GMT file and the raw catalogue, and `LocusToGeneFeatureMatrixStep` now reads Parquet only.
- `PathwayEnrichment` fields renamed to `pValue`, `pValueAdjusted` and `normalisedEnrichmentScore`, with `from_gsea_catalogue` doing the renaming on the way in, and `tested_pathways()` exposing the denominator explicitly.
- The correction is now `row_number()` over `(pValue, pathway)`, leaves a null p-value null rather than 1.0, and skips non-finite enrichment scores. It is also switchable off through `enriched_pathways(..., recompute_missing_adjusted_p_value=False)`, and its docstring states that the recomputed value is adjusted over the stored positively enriched rows only and is not comparable with the published column.
- Feature: semi joins on the two joins that were only filtering, `studyType` and non-empty disease filters before the sort, the earlier dedup, and the denominator restricted to tested pathways.
- Dataset tests split into `test_pathway_index.py` and `test_pathway_enrichment.py`, with cases for ties against R's `p.adjust`, null p-values, non-finite enrichment scores and unresolved identifiers.

Re-run against the real library and catalogue: 8,859 pathways with 8,809 identifiers (5,954 GO, 2,855 Reactome, 50 Hallmark with none), 17,246 symbols resolving to 16,775 genes over 333,052 membership rows, 8,217 tested pathways, and psoriasis — one of the 249 diseases with no published adjusted p-value — now recovering 400 enriched pathways with every value inside [0, 1]. The PR description is updated.

Still open, and both are yours to call: whether skipping the `nes = +inf` rows is right, and whether the correction should be off by default rather than on.

src/gentropy/dataset/pathway_index.py
project-defiant: In that case we shall make it the same as in other datasets 
`pathwayId` and `pathwayFromSourceName`, `pathwayFromSourceId`, the FromSource fields will represent the source information, while pathwayId would reflect validated entries that map to OT datasets.

src/gentropy/dataset/pathway_index.py
project-defiant: We should hold 2 versions of the dataset (1 - generated from enrichemnt - with source fields only and. 2- the release should validate the names and ids from source and map them to the `current` version while excluding ones that miss the mapping. This would mean a few columns (pathwayId -> resolved during release, pathwayFromSourceName, and pathwayFromSourceId -> populated from the source)

src/gentropy/dataset/pathway_index.py
addramir: Agreed: `pathwayFromSourceName`, `pathwayFromSourceId`, and `pathwayId` for the resolved entry. `PathwayEnrichment` gets the same rename. One constraint — `pathwayFromSourceName` is the join key between library and enrichment, so it stays verbatim, bracketed tag included.

src/gentropy/dataset/pathway_index.py
addramir: Agreed on the shape. But there is no pathway index in gentropy to validate against — only the target index's `pathways` (Reactome annotations of genes) and `go`, which are gene annotation sets, so "resolvable" means "annotated to at least one OT gene".

That resolves 6,911 of 8,217 tested pathways; the other 1,306 are 729 GO BP, 527 Reactome and the 50 Hallmark sets that have no identifier at all. Excluding them drops 15% of rows and, since the denominator is the gene's count of tested pathways, shifts every score — on annotation coverage, not on whether the pathway is real.

So: populate `pathwayId` where it resolves, null where it does not, join on `pathwayFromSourceName`. Exclusion can go behind a flag, off by default. Validating against Reactome and GO proper needs an ontology input we do not have — separate issue.

src/gentropy/dataset/l2g_features/pathway.py
project-defiant: Agreed.

src/gentropy/dataset/pathway_index.py
addramir: Pushed `3b2bf006`: `pathwayFromSourceName`, `pathwayFromSourceId`, `pathwayId` in both datasets, and `PathwayIndex.resolve_pathway_ids` filling `pathwayId` from the target index. Against 26.03 it resolves 7,498 of the 8,859 library pathways (5,225 GO, 2,273 Reactome). Unresolved ones keep a null `pathwayId`; `PathwayIngestionStep` takes `exclude_unmapped_pathways`, off by default.

src/gentropy/assets/schemas/pathway_index.json
```diff
@@ -0,0 +1,49 @@
+{
+  "type": "struct",
+  "fields": [
+    {
+      "name": "pathwayFromSourceName",
+      "type": "string",
+      "nullable": false,
```
project-defiant, line 7: could be null, you may have Name or Id from source probably.

src/gentropy/pathway.py
```diff
@@ -0,0 +1,64 @@
...
+                the pathway enrichment features divide by the number of pathways a gene belongs
+                to, so dropping pathways moves every score.
+        """
+        target_index = TargetIndex.from_parquet(session, target_index_path)
+
+        pathway_index = (
+            PathwayIndex.from_gmt(session, pathway_library_path)
+            .resolve_pathway_ids(target_index, exclude_unmapped_pathways)
+            .resolve_gene_ids(target_index)
+        )
```
project-defiant, line 54: Persist it, otherwise you would recompute the pathway_index in 2 streams of work
1) to save it 
2) to build & save the pathway enrichment

project-defiant (approved):
2 small comments, otherwise you can merge it.

In the future we may think if we want to produce the (output + excluded) datasets and do the QC steps, but these are not needed for the feature atm, as you can easily exclude failing pathways. 

 Great work!

project-defiant:
On the fallback adjusted-p-value correction: 

I think we should either recompute over the truncated pathway for each disease, when building the PathwayEnrichment dataset - this means that the meaning of adjustedPvalue reflects the multiple tests over the positive-NES subset of tested pathways and we should clearly state it. In case you want to make it optional, ensure that flag functionality is clear. 

For rows with NES = +inf and p-value = 0, I think we should first check with the experiments you did whether these are expected signals in the underlying experiments or degenerate results, before choosing the way to deal with them. 

You can zero-out the pvalues to infinitesimal positive value supported by type and then the recomputation will yield exactly that value for the padj (rank=1, padj=pval) for all

src/gentropy/assets/schemas/pathway_index.json
addramir: Kept non-null: it is the join key to `PathwayEnrichment`, so a row without it cannot reach the feature. In GMT the name is the mandatory first field; only the id (second field) can be empty, and `pathwayFromSourceId` is already nullable for that.

src/gentropy/pathway.py
addramir: `pathway_index` has one action here, the write. The enrichment is built from the catalogue alone and does not read the index, so there is no second pass to save.

addramir:
On the fallback correction, pushed `604f47b3`: the recomputation now runs in `PathwayIngestionStep` behind `recompute_missing_adjusted_p_value` (on by default), and the docstring states it is adjusted over the positive-NES subset only. For now, rows with non-finite NES get a null adjusted p-value, published or recomputed, so they cannot count as enriched but still count as tested. The underlying numbers will be fixed upstream in the GSEA catalogue.

PR #1287, opened 2026-09-02
feat(study locus): flag credible sets replicated in an independent study

### LINKED ISSUE #3999: Credible set replication flag
We want to add the flat to credible sets indicating that CS is considered as replicated.

addramir:
## Summary

Flags the credible sets whose lead variant is replicated in an independent study. The replication logic is the one behind the `replicated_gwas_cs` / `replicated_molqtl_cs` tables of the manuscript's data preparation notebooks.

- `StudyLocusQualityCheck.REPLICATED`, raised by the new `StudyLocus.qc_replication(study_index)`.
- Wired into `StudyLocusValidationStep`.

The flag is positive: it records that a credible set *is* replicated. Nothing is ever flagged as a failure to replicate, so `REPLICATED` is never a reason to consider a credible set invalid, and its absence carries no claim — a credible set that was never assessed simply does not have it.

## Logic

A GWAS credible set is replicated if its lead variant is associated with the same diseases in at least two independent studies. GWAS studies have to agree on their **full list** of diseases: two studies sharing a single disease out of several describe different phenotypes and are not a replication of one another. Studies agreeing on `cohorts`, `pubmedId` and `ldPopulationStructure` are the same evidence twice over and are collapsed before counting.

A molQTL credible set is replicated if its lead variant is associated with the same gene in at least two studies.

GWAS studies with no disease annotation and molQTL studies with no measured gene have nothing to replicate on, so their credible sets are left unflagged rather than reported as unreplicated.

## Placement in the validation step

`qc_replication` runs at the very end of `StudyLocusValidationStep`, after both valid/invalid splits, on the credible sets that passed every other check and survived the duplicate check. A credible set removed by an earlier flag is not evidence of replication, and two duplicated credible sets of one study must not count as two studies.

## Not in this PR

The confidence part is proposed separately. Replication does not feed `assign_confidence` here, there is no new `CredibleSetConfidenceClasses` member, and the `credibleSetConfidence` L2G feature is untouched.

### DIFF
--- a/src/gentropy/dataset/l2g_features/other.py
+++ b/src/gentropy/dataset/l2g_features/other.py
@@ -336,6 +336,10 @@ class CredibleSetConfidenceFeature(L2GFeature):
         """
         return (
             f.when(
+                f.col("confidence") == CredibleSetConfidenceClasses.REPLICATED.value,
+                f.lit(1.0),
+            )
+            .when(
                 f.col("confidence")
                 == CredibleSetConfidenceClasses.FINEMAPPED_IN_SAMPLE_LD.value,
                 f.lit(1.0),
--- a/src/gentropy/dataset/l2g_features/other.py
+++ b/src/gentropy/dataset/l2g_features/other.py
@@ -346,11 +346,6 @@ class CredibleSetConfidenceFeature(L2GFeature):
             )
             .when(
                 f.col("confidence")
-                == CredibleSetConfidenceClasses.REPLICATED_TOP_HIT.value,
-                f.lit(0.75),
-            )
-            .when(
-                f.col("confidence")
                 == CredibleSetConfidenceClasses.FINEMAPPED_OUT_OF_SAMPLE_LD.value,
                 f.lit(0.75),
             )
--- a/src/gentropy/dataset/l2g_features/other.py
+++ b/src/gentropy/dataset/l2g_features/other.py
@@ -346,6 +346,11 @@ class CredibleSetConfidenceFeature(L2GFeature):
             )
             .when(
                 f.col("confidence")
+                == CredibleSetConfidenceClasses.REPLICATED_TOP_HIT.value,
+                f.lit(0.75),
+            )
+            .when(
+                f.col("confidence")
                 == CredibleSetConfidenceClasses.FINEMAPPED_OUT_OF_SAMPLE_LD.value,
                 f.lit(0.75),
             )
--- a/src/gentropy/dataset/study_locus.py
+++ b/src/gentropy/dataset/study_locus.py
@@ -418,6 +418,20 @@ class StudyLocus(Dataset):
             _schema=StudyLocus.get_schema(),
         )
 
+    @staticmethod
+    def can_assess_replication(study_index: StudyIndex) -> bool:
+        """Whether credible set replication can be assessed against the given study index.
+
+        Args:
+            study_index (StudyIndex): Study index to test for the annotation `qc_replication` needs.
+
+        Returns:
+            bool: True if the study index carries both the disease and the gene annotation.
+        """
+        return all(
+            column in study_index.df.columns for column in ["diseaseIds", "geneId"]
+        )
+
     @qc_test
     def qc_replication(self: StudyLocus, study_index: StudyIndex) -> StudyLocus:
         """Flagging credible sets whose lead variant is not replicated in an independent study.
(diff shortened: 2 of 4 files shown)


src/gentropy/dataset/study_locus.py
```diff
@@ -405,13 +409,123 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+                "diseaseId", f.explode("diseaseIds")
+            )
+            replicated_gwas_pairs = (
+                gwas.select("variantId", "diseaseId", *gwas_independence_columns)
+                .distinct()
+                .groupBy("variantId", "diseaseId")
+                .agg(f.count("*").alias("studyCount"))
+                .filter(f.col("studyCount") >= 2)
+                .select("variantId", "diseaseId")
+            )
```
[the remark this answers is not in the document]
addramir: Fixed by dropping the optional-column handling: the check now requires the study index annotation instead of degrading without it.

src/gentropy/study_locus_validation.py
```diff
@@ -60,13 +60,18 @@ def __init__(
...
             # Flagging credible sets with PIP > 1 or PIP < 0.95
             .qc_abnormal_pips(
                 sum_pips_lower_threshold=0.95,
                 sum_pips_upper_threshold=1.0001,
             )
-            # Annotate credible set confidence:
-            .assign_confidence()
+            # Annotate credible set confidence, taking the replication flag raised above
+            # into account:
+            .assign_confidence(use_replication=True)
```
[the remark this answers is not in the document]
addramir: Correct as it stands now: `qc_replication` no longer returns `self`, and it runs immediately before this on the same rows.

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -336,6 +336,10 @@ def score_credible_set_confidence(
         """
         return (
             f.when(
+                f.col("confidence") == CredibleSetConfidenceClasses.REPLICATED.value,
+                f.lit(1.0),
+            )
```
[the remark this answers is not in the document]
addramir: Pre-existing and unrelated to this PR, leaving it alone here.

tests/gentropy/dataset/test_study_locus.py
```diff
@@ -1272,6 +1272,222 @@ def test_duplication_flag_retains_smallest_credible_set(
...
+            _schema=StudyIndex.get_schema(),
+        )
+
+        assert (
+            self.study_locus.qc_replication(unannotated)
+            .df.filter(f.size("qualityControls") > 0)
+            .count()
+            == 0
+        )
+
```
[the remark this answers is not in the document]
addramir: No longer applicable: the check does not degrade to partial annotation any more, so there is no partial-annotation path to cover.

src/gentropy/dataset/l2g_features/other.py
```diff
@@ -336,6 +336,10 @@ def score_credible_set_confidence(
         """
         return (
             f.when(
+                f.col("confidence") == CredibleSetConfidenceClasses.REPLICATED.value,
+                f.lit(1.0),
```
project-defiant: @addramir, I would not add the replication flag to this system, as person will not have an idea to understand the ambiguity of three cases
* When 2 cs are replicating each other (still somehow weak replication)
* When N > 2 cs are replicating each other (much better, especially when they come from ex. Exome and genotyping array)
* When the actual cs is from in sample ld mapping and has a replication

In my opinion the confidence shall be either (replication x info about LD usage) or leave confidence as is, but add the replication as a separate concern

src/gentropy/dataset/study_locus.py
```diff
@@ -412,6 +416,92 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+                "pubmedId",
+                "ldPopulationStructure",
+            ),
+            on="studyId",
+            how="left",
+        )
+
+        # Exploding the diseases drops GWAS credible sets without disease annotation:
+        gwas = loci.filter(f.col("studyType") == "gwas").withColumn(
+            "diseaseId", f.explode("diseaseIds")
```
project-defiant: I would be very careful about this, I would use only valid credible sets as input to replication.

That stated, we will not have cases where study does not have associated diseaseId, as these studies (and subsequent credible sets) will be left in invalid batch. 

and general question: Would 2 credible sets that have just one trait in common considered as replication?

Here just a few examples, we have many many more in the study dataset

```
study1 study2 trait1(uncommon) trait2(uncommon) commonTrait
GCST002309	GCST90093255	alcohol drinking	major depressive disorder	diastolic blood pressure
GCST006191	GCST90093255	smoking status measurement	major depressive disorder	diastolic blood pressure
GCST008411	GCST90093255	physical activity	major depressive disorder	diastolic blood pressure
GCST010254	GCST90093255	lifestyle measurement	major depressive disorder	diastolic blood pressure
```

src/gentropy/dataset/study_locus.py
```diff
@@ -412,6 +416,92 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+        )
+        replicated_gwas_loci = gwas.join(
+            gwas.select(
+                "variantId",
+                "diseaseId",
+                "cohorts",
```
project-defiant, line 477: `cohorts` is an unsorted list that sometimes contains non-normalised values like `UKB` or `UKBB`. I think using it in `distinct` is not super meaningful at the moment. We would need to somehow standardise the naming there and fuzzy match on duplicates

src/gentropy/dataset/study_locus.py
```diff
@@ -412,6 +416,92 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+                "pubmedId",
+                "ldPopulationStructure",
+            )
+            .distinct()
+            .groupBy("variantId", "diseaseId")
+            .count()
+            .filter(f.col("count") >= 2),
```
project-defiant, line 484: To be fair, I probably would like to see the `replication count` rather then just a boolean, since we can have a strong / weak replication

src/gentropy/dataset/study_locus.py
```diff
@@ -412,6 +416,92 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+            on=["variantId", "diseaseId"],
+            how="inner",
+        ).select("studyLocusId")
+
+        molqtl = loci.filter(f.col("studyType") != "gwas")
+        replicated_molqtl_loci = molqtl.join(
+            molqtl.groupBy("variantId", "geneId").count().filter(f.col("count") >= 2),
```
project-defiant, line 493: What about the biosample ? Does it not fit into the trait for molecular qtls?

I would also add the `studyType` or at least `projectId` as a discriminative, as we may just duplicate the eql that was also treated as tuqtl/sqtl for instance within one eqtl catalogue project (multiple quant methods). In the case below, even if we duplicate the signal, this comes from the fact that first study uses txrev, while second uses tx quant.

```
studyId	studyType	geneId	biosampleId	variantId	traitFromSource
gtex_txrev_fibroblast_ensg00000000003_grp_1_upstream_enst00000494424	tuqtl	ENSG00000000003	CL_0000057	X_100688247_G_C	ENSG00000000003.grp_1.upstream.ENST00000494424
gtex_tx_fibroblast_enst00000494424	eqtl	ENSG00000000003	CL_0000057	X_100688247_G_C	ENST00000494424
```

project-defiant (commented):
Having read the PR, I can see that we potentially can fit the binary replication information without any changes to the output schema, but I feel we are losing the resolution with this approach, when not reporting the magnitude of replication.

There are couple of comments on how the replication is computed (especially molqtl), please have a look.

addramir:
@project-defiant I have changed the PR according to our disscussion on Friday. The molQTL replication stays the same, but in GWAS we take into account all diseaseIDs. Please have a look.

src/gentropy/dataset/study_locus.py
```diff
@@ -62,6 +62,9 @@ class CredibleSetConfidenceClasses(Enum):
 class StudyLocusQualityCheck(Enum):
     """Study-Locus quality control options listing concerns on the quality of the association.
 
+    All flags but `REPLICATED` describe a concern; `REPLICATED` records a positive observation
```
project-defiant, line 65: Not 100% true, TOP_HIT is also an informative flag.

src/gentropy/dataset/study_locus.py
```diff
@@ -417,6 +422,101 @@ def validate_unique_study_locus_id(self: StudyLocus) -> StudyLocus:
...
+                        f.col("qualityControls"),
+                        f.col("isReplicated").isNotNull(),
+                        StudyLocusQualityCheck.REPLICATED,
+                    ),
+                )
+                .drop("isReplicated")
+            ),
+            _schema=self.get_schema(),
+        )
+
```
project-defiant, line 519: In general the whole validation step is quiet iefficient. Worth throwing the persist just after reading the data and broadcasting the study (as the dataset is really small).

From the performance reasons, filtering for molQTL and gwas and doing 2 separate joins may be faster (if caching is enabled), especially that you rejoin with credible set at the end.

src/gentropy/study_locus_validation.py
```diff
@@ -83,9 +83,14 @@ def __init__(
         # only when DUPLICATED_STUDYLOCUS_ID is one of the configured invalid reasons.
         unique = deduplicated.valid_rows(invalid_qc_reasons)
 
+        # Replication is assessed at the very end, once every invalid credible set has been
+        # removed: a credible set dropped by any earlier flag, a duplicate included, is not
+        # evidence of replication, and duplicated credible sets of one study would count as two.
+        replicated = unique.valid.qc_replication(study_index)
```
project-defiant, line 89: The study_index is used 4 times here. We should ensure it is persisted or cached just after reading.

project-defiant (approved):
Thank you for bringing this functionality to gentropy!

We need to improve the step performance in the future. We discussed all of the drawbacks of this approach, but agreed to proceed to align with the way how the replication is done in the manuscript. 

The concerns remain for future work:
1) cohorts field contains non-standardised values, so we can have a replicated cs which comes from independent study, but same cohort
2) no biosampleId in molqtl replication (due to diversity of molQTL biosamples, would cause the cs to be non-replicated most of the time)
3)  Resolving problem with molQTLs where the replication over different studyType not an independent signal (sQTL and eQTL from the same publication, biosample and gene tested) - eQTL Catalogue.

src/gentropy/dataset/study_locus.py
addramir: Right, fixed. The docstring now says that most flags describe a concern, and names `TOP_HIT` alongside `REPLICATED` as the descriptive ones that are not in themselves a reason to consider a credible set invalid.

src/gentropy/dataset/study_locus.py
addramir: Done. The study index is now split by study type before the join, so each branch joins only the annotation it counts on (`diseaseIdSet`, `cohorts`, `pubmedId`, `ldPopulationStructure` for GWAS, `geneId` for molQTL), and both small sides are broadcast. The joins are inner now, which also removes the separate filtering pass: a GWAS study with no disease annotation and a molQTL study with no gene simply do not make it into their branch.

The credible sets themselves are already cached when this runs -- `qc_replication` is called on `unique.valid`, which comes off the persisted `deduplicated` -- so the two branches and the final rejoin read from that cache rather than re-reading the input.

The wider inefficiency of the validation step is real but larger than this PR; I would rather address it separately than grow this one.

src/gentropy/study_locus_validation.py
addramir: Added: the study index is persisted right after reading and unpersisted at the end of the step, next to the two caches that were already released there.

